PR #12, opened 2021-12-06
Allow opt-in go cli scanning

tevoinea:
Using go cli to scan is more accurate than parsing go.sum files, but go projects aren't always restored in a workflow. This results in go restoring packages in a workflow where it's not required. By allowing an opt-in we give users the power to make the right choice for their specific situations.

We still fall back to manual parsing if go cli fails.


.vscode/launch.json
```diff
@@ -10,7 +10,7 @@
             "request": "launch",
             "preLaunchTask": "build",
             // If you have changed target frameworks, make sure to update the program path.
-            "program": "${workspaceFolder}/src/Microsoft.ComponentDetection/bin/Debug/net6.0/Microsoft.ComponentDetection.dll",
+            "program": "${workspaceFolder}/src/Microsoft.ComponentDetection/bin/Debug/netcoreapp3.1/Microsoft.ComponentDetection.dll",
```
tevoinea, line 13: Tiny thing we missed moving back to netcore 3.1

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -34,35 +38,62 @@ public class GoComponentDetector : FileComponentDetector
...
+                if (IsGoCliManuallyEnabled())
+                {
+                    Logger.LogInfo("Go cli scan was manually enabled");
+                    wasGoCliScanSuccessful = await UseGoCliToScan(file.Location, singleFileComponentRecorder);
+                }
+            }
+            catch
+            {
+                Logger.LogInfo("Failed to detect components using go cli.");
+            }
```
JamieMagee, line 64: Can you log the exception here

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -223,5 +254,10 @@ private bool TryCreateGoComponentFromRelationshipPart(string relationship, out G
             goComponent = new GoComponent(componentParts[0], componentParts[1]);
             return true;
         }
+
+        private bool IsGoCliManuallyEnabled()
+        {
+            return EnvVarService.DoesEnvironmentVariableExist("EnableGoCliScan");
```
JamieMagee, line 260: Can you create a constants class for environment variables? If we're going to create a lot of these, it'll be easier to keep track of.

JamieMagee (approved):
Couple of small things, but nothing blocking.

PR #27, opened 2021-12-21
Support target specific cargo dependencies

tevoinea:
Adds support for capturing platform specific dependencies[^1].

cc: @danielframpton who first raised this issue in another repository (thank you!)

[^1]: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies


tevoinea:
Snapshot verification failures are expected because there are platform specific dependencies in our verification repo. The number of components shouldn't change because they all get picked up in `cargo.lock` but the number of explicitly referenced components should change because we're now scanning them in `cargo.toml`.

Previous run total: [7047](https://github.com/microsoft/component-detection/runs/4607747234?check_suite_focus=true#step:16:1048)
Current run total: [7047](https://github.com/microsoft/component-detection/runs/4609348167?check_suite_focus=true#step:16:1048)

Previous run explicitly referenced for cargo v2: [45](https://github.com/microsoft/component-detection/runs/4607747234?check_suite_focus=true#step:16:1044)
Current run explicitly referenced for cargo v2: **[47](https://github.com/microsoft/component-detection/runs/4609348167?check_suite_focus=true#step:16:1044)**

JamieMagee (approved):

JamieMagee:
@tevoinea The latest run of snapshot verification has a maven component change in there_

tevoinea:
> @tevoinea The latest run of snapshot verification has a maven component change in there?

@JamieMagee Probably transient or a version drift issue, rerunning to check.

JamieMagee:
@tevoinea do you have capacity to deconflict this? If not, let me know and I'll get around to it.

tevoinea:
@JamieMagee I don't right now sorry, I skimmed it and it seems like just RustCrateUtilities that needs any 'actual' conflict resolution.

cobya:
@tevoinea I did some conflict resolution (sorry for the delay, Jamie got distracted). Can you do a quick check and verify everything looks correct?

tevoinea:
@cobya The logic looks right to me! Just don't forget to bump the detector version number

PR #46, opened 2022-01-20
Support renamed dependencies in Cargo.toml

tofay:
Cargo supports renamed dependencies in Cargo.toml: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#renaming-dependencies-in-cargotoml

component-detection currently misses such dependencies, as it uses the renamed dependency name when determining what the root dependencies on. This PR updates component detection to use the real dependency name in such cases.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/rust/RustCrateDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/rust/RustCrateDetector.cs
@@ -21,7 +21,7 @@ namespace Microsoft.ComponentDetection.Detectors.Rust
 
         public override IEnumerable<ComponentType> SupportedComponentTypes => new[] { ComponentType.Cargo };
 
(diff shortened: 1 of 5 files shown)


JamieMagee:
@tofay Sorry for the delay. Can you deconflict your PR?

tofay:
No problem, I've rebased.

JamieMagee:
Thanks, this all looks good to me.

We recently moved our integration tests into this repo as well. Could you plese add a test case [here](https://github.com/microsoft/component-detection/blob/main/test/Microsoft.ComponentDetection.VerificationTests/resources/rust/standard/Cargo.toml) as well?

I'll approve in the meantime, so we can merge this soon after.

tofay:
I've updated the standard verification test to have a renamed dependency.

JamieMagee (approved):

PR #52, opened 2022-01-24
Initial implementation of VcpkgDetector and VcpkgComponent

ras0219-msft:
This PR contains an initial implementation of detecting SPDX files produced by vcpkg. This depends upon experimental features that are not yet released in the vcpkg tool.

#### New Detector Checklist

- [x] I have gone through the docs for creating a new detector [here](https://github.com/microsoft/component-detection/blob/main/docs/creating-a-new-detector.md)
- [x] My new detector implements `IDefaultOffComponentDetector`
- [ ] I have created a PR to the [verification repo](https://github.com/microsoft/componentdetection-verification) with components that my new detector can find
- [ ] (If necessary) I have updated the [Feature Overview](https://github.com/microsoft/component-detection/blob/main/README.md#feature-overview) table in the README

### DIFF
--- a/src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
+++ b/src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
@@ -1,0 +1,63 @@
+﻿using PackageUrl;
+
+namespace Microsoft.ComponentDetection.Contracts.TypedComponent
+{
+    public class VcpkgComponent : TypedComponent
+    {
+        private VcpkgComponent()
+        {
+            /* Reserved for deserialization */
+        }
+
+        public VcpkgComponent(string spdxid, string name, string version, string triplet = null, string portVersion = null, string description = null, string downloadLocation = null)
+        {
+            SPDXID = ValidateRequiredInput(spdxid, nameof(SPDXID), nameof(ComponentType.Vcpkg));
+            Name = ValidateRequiredInput(name, nameof(Name), nameof(ComponentType.Vcpkg));
+            Version = version;
+            PortVersion = portVersion;
+            Triplet = triplet;
+            Description = description;
+            DownloadLocation = downloadLocation;
+        }
+
+        public string SPDXID { get; set; }
+
+        public string Name { get; set; }
+
+        public string DownloadLocation { get; set; }
+
+        public string Triplet { get; set; }
+
+        public string Version { get; set; }
+
+        public string Description { get; set; }
+
+        public string PortVersion { get; set; }
+
+        public override ComponentType Type => ComponentType.Vcpkg;
+
+        public override string Id
+        {
+            get
+            {
+                if (PortVersion != null)
+                    return $"{Name} {Version}#{PortVersion} - {Type}";
+                else
+                    return $"{Name} {Version} - {Type}";
+            }
+        }
+
+        public override PackageURL PackageUrl
+        {
+            get
+            {
+                if (PortVersion != null)
+                    return new PackageURL($"pkg:vcpkg/{Name}@{Version}?port_version={PortVersion}");
+                else if (Version != null)
+                    return new PackageURL($"pkg:vcpkg/{Name}@{Version}");
+                else
+                    return new PackageURL($"pkg:vcpkg/{Name}");
+            }
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


ras0219-msft:
+@grvillic for review

JamieMagee:
Moving this to draft for now. There's still an open question about a generic SPDX detector.

src/Microsoft.ComponentDetection.Detectors/vcpkg/VcpkgComponentDetector.cs
```diff
@@ -0,0 +1,105 @@
...
+            {
+                if (item.Name == null || item.Name.Length == 0)
+                {
+                    continue;
+                }
+
+                try
+                {
+                    Logger.LogWarning($"parsed package {item.Name}");
+                    if (item.SPDXID == "SPDXRef-port")
```
grvillic, line 72: nit: Should we trim things and make all SPDXIDs case insensitive?

ras0219-msft: They are by spec case-sensitive, I believe. Also as mentioned offline we'll eventually want to transition to using annotations to identify components instead of the ID.

grvillic (approved):
Thanks for the contribution!

src/Microsoft.ComponentDetection.Detectors/vcpkg/Contracts/Annotation.cs
```diff
@@ -0,0 +1,13 @@
+namespace Microsoft.ComponentDetection.Detectors.Vcpkg.Contracts
+{
+    public class Annotation
+    {
+        public string Date { get; set; }
```
grvillic, line 5: Why are we having a date as String?

ras0219-msft: Entirely because I'm not familiar enough with C# to know all the possible Date edge cases and don't have any plans to use the information. The format is specified to be ISO 8601 (e.g. `2010-01-29T18:30:22Z`) -- could you suggest code to correctly handle that?

grvillic: Sure I can push a change for that in a bit...

src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
```diff
@@ -0,0 +1,73 @@
+﻿using PackageUrl;
+
+namespace Microsoft.ComponentDetection.Contracts.TypedComponent
+{
+    public class VcpkgComponent : TypedComponent
```
grvillic, line 5: Can we add tests for this new detector? See other detector test samples in [Microsoft.ComponentDetection.Detectors.Tests](https://github.com/microsoft/component-detection/tree/main/test/Microsoft.ComponentDetection.Detectors.Tests).

grvillic (changes requested):
We need to add tests validating the new detector.

grvillic (approved):

src/Microsoft.ComponentDetection.Detectors/vcpkg/VcpkgComponentDetector.cs
```diff
@@ -0,0 +1,118 @@
...
+            try
+            {
+                sbom = JsonConvert.DeserializeObject<VcpkgSBOM>(await reader.ReadToEndAsync());
+            }
+            catch (Exception)
+            {
+                return;
+            }
+
+            if (sbom == null || sbom.Packages == null)
```
grvillic, line 70: nit
```suggestion
            if (sbom?.Packages == null)
```

test/Microsoft.ComponentDetection.Detectors.Tests/VcpkgComponentDetectorTests.cs
```diff
@@ -0,0 +1,174 @@
...
+                                                  .WithFile("vcpkg.spdx.json", spdxFile)
+                                                    .ExecuteDetector();
+
+            Assert.AreEqual(ProcessingResultCode.Success, scanResult.ResultCode);
+
+            var detectedComponents = componentRecorder.GetDetectedComponents();
+            var components = detectedComponents.ToList();
+
+            Assert.AreEqual(2, components.Count());
+            {
```
grvillic, line 125: We don't need these curly brackets here. you will can just set the variable below without initializing it (without `var`)

PR #57, opened 2022-01-31
Fix Pypi Dev version parsing

grvillic:
## Problem
Our PipComponentDetector is having issues resolving for dev dependency versions that do not have any number as suffix, e.g. `0.5.7.dev`. This causes `IPyPiClient` to parse that version as if it were a release candidate `0.5.7` and consequently throws errors when adding it to dictionary, see line below: https://github.com/microsoft/component-detection/blob/6176381b4dfefe649cc51be95a5ead764e3e5721/src/Microsoft.ComponentDetection.Detectors/pip/IPyPiClient.cs#L213

## Steps to reproduce
1. Create a `requirements.txt` file
2. Add a dependency on `nltk==3.6.2`
3. Run de detector look at the failure caused by [ntlk -> joblib](https://github.com/nltk/nltk/blob/6797ee3958fe3fa0192d347b26ea0fc433b8bfc5/pip-req.txt#L14) version parsing.

## Solution
Add a new property that stores the `dev` label, and set the `devNumber` version to 0 to ensure we don't flag this package version as a "released".

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/pip/IPyPiClient.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/pip/IPyPiClient.cs
@@ -210,7 +210,7 @@ namespace Microsoft.ComponentDetection.Detectors.Pip
                 }
                 catch (ArgumentException ae)
                 {
-                    Logger.LogError($"Component {release.Key} : {JsonConvert.SerializeObject(release.Value)} could not be added to the sorted list of pip components. Error details follow:");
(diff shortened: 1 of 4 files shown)


jcfiorenzano:
Shouldn't we add a test that validates that the components are parsed properly. The test that was modified in this PR validates the order, but in that test already existed versions with "dev" suffix so I think that we are missing a test in here where is evident the fix

jcfiorenzano (approved):

PR #62, opened 2022-02-02
refactor: use collection’s count property

JamieMagee:
`Count()` is a LINQ extension method, whereas `Count` is a property of the collection.

See: https://stackoverflow.com/questions/4098186/lists-count-vs-count

### DIFF
--- a/src/Microsoft.ComponentDetection.Common/TabularStringFormat.cs
+++ b/src/Microsoft.ComponentDetection.Common/TabularStringFormat.cs
@@ -48,7 +48,7 @@ namespace Microsoft.ComponentDetection.Common
             foreach (var row in rows)
             {
                 sb.Append(verticalLineChar);
-                if (row.Count() != columns.Count)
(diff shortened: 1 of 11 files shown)


jcfiorenzano:
what is the reason for this?

I understand that Count is more performant than Count().
Do we have data about this is affecting us?
Are we going to remove every linq use in this repo.
When new contributors create PRs, is this something that need to look at?

JamieMagee:
> what is the reason for this?

It's a small refactor that Visual Studio suggested to me as a build warning. Very low risk IMO.

> I understand that Count is more performant than Count().
> Do we have data about this is affecting us?

Not about us specifically, but it is a minor optimization.

> Are we going to remove every linq use in this repo

No, but this was a build warning.

> When new contributors create PRs, is this something that need to look at?

If we fix build warnings, and make warnings errors, this should fix the issue.

jcfiorenzano (approved):

PR #71, opened 2022-02-28
Move from PackageUrl package to packageurl-dotnet

dfederm:
This package was renamed as part of reviving the underlying repo.

See: https://github.com/package-url/packageurl-dotnet/pull/10 and https://github.com/package-url/packageurl-dotnet/pull/13


JamieMagee:
Test failures are due to snapshot differences. Looks like I didn't lock down the dependency graph enough. I'll fix that in a separate PR.

JamieMagee (approved):

PR #75, opened 2022-03-11
Support case-insensitive on non-Windows OSs.

boumenot:
The environment variable EnableGoCliScan existence gates usage of the Go
CLI tools for determing what modules are in-use.  The current check does
a get of the environment variable, and if it exists behavior is enabled.
On Windows this is case-insensitive, but on Linux (or MacOS) this is
case-sensitive so the user must exactly use the casing of
'EnableGoCliScan'.

Our CI system automatically capitalizes all environment variables when
they are defined, so EnableGoCliScan becomes ENABLEGOCLISCAN.  I am not
aware of a way to control this behavior, so there is no way to enable Go
CLI tooling.  My fix is to treat all environment variable exitence
checks as case-insensitive.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
@@ -33,7 +33,7 @@ namespace Microsoft.ComponentDetection.Detectors.Go
 
         public override IEnumerable<ComponentType> SupportedComponentTypes { get; } = new[] { ComponentType.Go };
 
-        public override int Version => 2;
+        public override int Version => 3;
 
         private HashSet<string> projectRoots = new HashSet<string>();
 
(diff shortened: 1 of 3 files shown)


grvillic (approved):
Thanks for the contribution! While not a perfect solution I understand why we need it to make it work cross-platform.

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -33,7 +33,7 @@ public class GoComponentDetector : FileComponentDetector
 
         public override IEnumerable<ComponentType> SupportedComponentTypes { get; } = new[] { ComponentType.Go };
 
-        public override int Version => 2;
```
grvillic, line 36: bumped this to 3, as we expect to see new/different components in the snapshot

PR #76, opened 2022-03-15
Bump to net6

sailro:
Bump to .NET 6 and remove morelinq dependency, as `DistinctBy` is now natively supported by `Sytem.Linq`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,10 +22,10 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Setup .NET
+      - name: Setup .NET 6
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: 3.1.x
+          dotnet-version: 6.0.x
 
       - name: Restore packages
         run: dotnet restore
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
 
     <PropertyGroup Label="Build">
-        <TargetFramework>netcoreapp3.1</TargetFramework>
+        <TargetFramework>net6</TargetFramework>
         <LangVersion>latest</LangVersion>
         <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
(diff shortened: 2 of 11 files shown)


global.json
```diff
@@ -1,6 +1,6 @@
 {
 	"sdk": {
-		"version": "3.1.300",
+		"version": "6.0.200",
```
JamieMagee, line 3: Any reason why you're targeting `6.0.200` specifically? `6.0.100` is the baseline for .NET 6, and `6.0.201` is the latest release.

sailro: No, we can use any 6.0.x, I just used this one by muscle memory.

JamieMagee:
Thanks for the contribution. This is definitely something that we want to do, but there are a few teams internally that consume Component Detection, and we'll need to check that they are ready and able to consume a .NET 6 dependency.

RushabhBhansali (approved):

Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup Label="Build">
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net6</TargetFramework>
```
JamieMagee, line 4: The target framework is `net6.0`

```suggestion
    <TargetFramework>net6.0</TargetFramework>
```

global.json
JamieMagee: `6.0.401` is now the latest release at time of writing https://dotnet.microsoft.com/en-us/download/dotnet/6.0

.github/workflows/build.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Setup .NET
         uses: actions/setup-dotnet@v3
         with:
-          dotnet-version: 3.1.x
+          dotnet-version: 6.0.x
```
JamieMagee, line 28: @RushabhBhansali please remove all these versions.

```yml
with:
  dotnet-version: 6.0.x
```

`actions/setup-dotnet` will read the `global.json` file to get the SDK version. Having the version defined here is just duplication.

JamieMagee:
Closing this in favour of #288

Issue #68, opened 2022-02-15
Yarn Detector issues


grvillic:
## Problem

Parsing Yarn lock files fails when adding same package to dictionary, we should not break yarn detection as a whole but rather log the conflicting entry.

### Error logs:
`{"An item with the same key has already been added. Key: lodash@version"}`

Error seems to be caused by this line:

https://github.com/microsoft/component-detection/blob/9fabc327880c4122f760a2554f7b1242f319f7b7/src/Microsoft.ComponentDetection.Detectors/yarn/YarnLockComponentDetector.cs#L72

## Solution
Try adding to dictionary, if entry already exists, log warning and proceed.

JamieMagee:
This is partially fixed in #77. Though further investigation or changes are required in our Yarn detector.

cobya:
I haven't heard any customer reports of this being an issue and we have already resolved the exception. Closing this unless something comes up.

PR #77, opened 2022-03-16
fix: log a warning when duplicate components are found in a yarn.lock

JamieMagee:
It's possible, but rare, for yarn dependencies in the same block to have different names. This is a partial fix for #68, but more investigation may be required.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/yarn/YarnLockComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/yarn/YarnLockComponentDetector.cs
@@ -69,7 +69,12 @@ namespace Microsoft.ComponentDetection.Detectors.Yarn
             {
                 foreach (var satisfiedVersion in entry.Satisfied)
                 {
-                    yarnPackages.Add($"{entry.Name}@{satisfiedVersion}", entry);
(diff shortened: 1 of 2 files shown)


grvillic (approved):

src/Microsoft.ComponentDetection.Detectors/yarn/YarnLockComponentDetector.cs
```diff
@@ -69,7 +69,12 @@ private void DetectComponents(YarnLockFile file, string location, ISingleFileCom
             {
                 foreach (var satisfiedVersion in entry.Satisfied)
                 {
-                    yarnPackages.Add($"{entry.Name}@{satisfiedVersion}", entry);
+                    var key = $"{entry.Name}@{satisfiedVersion}";
+                    var addSuccessful = yarnPackages.TryAdd(key, entry);
```
grvillic, line 73: changes look good, but perhaps we should bump detector version. I don't think many users are impacted by this, but still out of cautiousness

JamieMagee: Good idea. Done!

Issue #35, opened 2022-01-03
IPyPiClient response caching strategy should be improved


cobya:
Within [microsoft/component-detection/IPyPiClient.cs at main](https://github.com/microsoft/component-detection/blob/main/src/Microsoft.ComponentDetection.Detectors/pip/IPyPiClient.cs#L76), the current caching implementation is simply a ConcurrentDictionary that will cache all of a PyPi responses's data to be used in case the dependency is found again later. However, in repositories with large Python package lists, this has a potential to allocate more memory than can be allocated in resource constrained environments.

The caching strategy within the client should be improved so that large Python dependency graphs can be generated using less allocated memory. Potential improvements such as migrating to an Least Recently Used cache should be investigated.

JamieMagee:
Just gonna drop this link for future reference: https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.imemorycache?view=dotnet-plat-ext-6.0

cobya:
With #80, I am close to closing this issue. We will perform some validation across Microsoft internally to verify the new caching functionality is working as expected and then update this issue.

PR #80, opened 2022-03-23
Migrate IPyPiClient cache to LRU MemoryCache

cobya:
Updates the caching mechanism in IPyPiClient from using a ConcurrentDictionary to a Microsoft.Extensions.Caching.Memory MemoryCache which allows setting a maximum number of cache entries and a cache entry expiration time. By limiting the number of entries the memory usage will be lowered and because of expiration time if the tool has not finished by the time the cache is no longer being used memory will be freed before exiting the program.

For users that need additional configuration (potentially due to memory constraints), allow them to set a maximum number of cache entries using the PyPiMaxCacheEntries environment variable.

This resolves the issue #35

### DIFF
--- a/src/Microsoft.ComponentDetection.Common/EnvironmentVariableService.cs
+++ b/src/Microsoft.ComponentDetection.Common/EnvironmentVariableService.cs
@@ -10,12 +10,7 @@ namespace Microsoft.ComponentDetection.Common
     {
         public bool DoesEnvironmentVariableExist(string name)
         {
-            // Environment variables are case-insensitive on Windows, and case-sensitive on
-            // Linux and MacOS.
-            // https://docs.microsoft.com/en-us/dotnet/api/system.environment.getenvironmentvariable
-            return Environment.GetEnvironmentVariables().Keys
-                .OfType<string>()
-                .FirstOrDefault(x => string.Compare(x, name, true) == 0) != null;
+            return GetEnvironmentVariable(name) != null;
         }
 
         public string GetEnvironmentVariable(string name)
--- a/src/Microsoft.ComponentDetection.Common/EnvironmentVariableService.cs
+++ b/src/Microsoft.ComponentDetection.Common/EnvironmentVariableService.cs
@@ -17,5 +17,17 @@ namespace Microsoft.ComponentDetection.Common
                 .OfType<string>()
                 .FirstOrDefault(x => string.Compare(x, name, true) == 0) != null;
         }
+
+        public string GetEnvironmentVariable(string name)
+        {
+            // Environment variables are case-insensitive on Windows, and case-sensitive on
+            // Linux and MacOS.
+            // https://docs.microsoft.com/en-us/dotnet/api/system.environment.getenvironmentvariable
+            var caseInsensitiveName = Environment.GetEnvironmentVariables().Keys
+                .OfType<string>()
+                .FirstOrDefault(x => string.Compare(x, name, true) == 0);
+
+            return caseInsensitiveName != null ? Environment.GetEnvironmentVariable(caseInsensitiveName) : null;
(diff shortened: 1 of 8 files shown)


src/Microsoft.ComponentDetection.Common/EnvironmentVariableService.cs
```diff
@@ -17,5 +17,17 @@ public bool DoesEnvironmentVariableExist(string name)
                 .OfType<string>()
                 .FirstOrDefault(x => string.Compare(x, name, true) == 0) != null;
         }
+
+        public string GetEnvironmentVariable(string name)
```
JamieMagee, line 21: Can `DoesEnvironmentVariableExist` be simplied to be

```csharp
return GetEnvironmentVariable(name) != null
```

src/Microsoft.ComponentDetection.Detectors/pip/IPyPiClient.cs
```diff
@@ -31,10 +31,18 @@ public class PyPiClient : IPyPiClient
...
+        [Import]
+        public IEnvironmentVariableService EnvironmentVariableService { get; set; }
+
         private static HttpClientHandler httpClientHandler = new HttpClientHandler() { CheckCertificateRevocationList = true };
 
         internal static HttpClient HttpClient = new HttpClient(httpClientHandler);
 
+        // Values used for cache creation
+        private const long CACHEINTERVALSECONDS = 60;
+        private const long DEFAULTCACHEENTRIES = 128;
```
JamieMagee, line 43:
```suggestion
        private const int CACHEINTERVALSECONDS = 60;
        private const int DEFAULTCACHEENTRIES = 128;
```

src/Microsoft.ComponentDetection.Detectors/pip/IPyPiClient.cs
```diff
@@ -45,52 +53,68 @@ public class PyPiClient : IPyPiClient
...
+                {
+                    Logger.LogInfo($"Setting IPyPiClient max cache entries to {maxEntries}");
+                    cachedResponses = new MemoryCache(new MemoryCacheOptions { SizeLimit = maxEntries });
+                }
+
+                checkedMaxEntriesVariable = true;
             }
 
-            return message;
+            return await cachedResponses.GetOrCreateAsync(uri, cacheEntry =>
```
JamieMagee, line 102: The order of things is a little weird here.

In `GetPyPiResponse` we try to fetch from cache and fallback to making an HTTP call. Here we attempt to fetch from cache again, and only then set.

It might make sense to combine both `GetPyPiResponse` and `CachePyPiResponseMemory` into one?

JamieMagee:
@co

src/Microsoft.ComponentDetection.Common/EnvironmentVariableService.cs
cobya: Changed to above.

src/Microsoft.ComponentDetection.Detectors/pip/IPyPiClient.cs
cobya: I updated the logic so caching occurs in `GetAndCachePyPiResponse` rather than relying on the caller, since both occurrences were caching it anyway.

src/Microsoft.ComponentDetection.Detectors/pip/IPyPiClient.cs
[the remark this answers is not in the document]
cobya: Yes, they are currently an estimate, but introduced a new telemetry class to track cache usage statistics so we can adjust in the future.

JamieMagee (approved):

PR #81, opened 2022-03-24
Initial implementation of SPDX22 Detector

ByAgenT:
This PR introduces detector for SPDX 2.2 files and SpdxComponent type. Initial implementation parses JSON SPDX file and extracts minimal set of information about the SPDX document necessary to link SPDX file in other SPDX files. It implements `IDefaultOffComponentDetector` interface.

This detector could eventually be expanded to retrieve information about what SPDX documents describes and what are dependencies for a product that document describes.

Detector has been tested inside Microsoft teams in external referencing feature for SBOMs and has been live in SBOM generation process for about a month now.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/spdx/Spdx22ComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/spdx/Spdx22ComponentDetector.cs
@@ -115,6 +115,5 @@ namespace Microsoft.ComponentDetection.Detectors.Spdx
             return BitConverter.ToString(SHA1.Create().ComputeHash(stream)).Replace("-", string.Empty).ToLower();
 #pragma warning restore CA5350
         }
-
     }
 }
--- a/src/Microsoft.ComponentDetection.Detectors/spdx/Spdx22ComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/spdx/Spdx22ComponentDetector.cs
@@ -14,7 +14,7 @@ using Newtonsoft.Json.Linq;
 namespace Microsoft.ComponentDetection.Detectors.Spdx
 {
     /// <summary>
-    /// Spdx22ComponentDetector discover SPDX SBOM files and create components with the information about
+    /// Spdx22ComponentDetector discover SPDX SBOM files in JSON format and create components with the information about
     /// what SPDX document describes. 
     /// </summary>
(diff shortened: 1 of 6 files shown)


src/Microsoft.ComponentDetection.Detectors/spdx/Spdx22ComponentDetector.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            try
+            {
+                var hash = GetSHA1HashFromStream(file.Stream);
+
+                // Reset buffer to starting position after hash generation.
+                file.Stream.Seek(0, SeekOrigin.Begin);
+
+                using (StreamReader sr = new StreamReader(file.Stream))
+                using (JsonReader reader = new JsonTextReader(sr))
```
JamieMagee, line 49: Can you use using statements here to reduce nesting?

```suggestion
                using var sr = new StreamReader(file.Stream))
                using var reader = new JsonTextReader(sr))
```

src/Microsoft.ComponentDetection.Detectors/spdx/Spdx22ComponentDetector.cs
```diff
@@ -0,0 +1,119 @@
...
+                {
+                    JsonSerializer serializer = new JsonSerializer();
```
JamieMagee, line 51:
```suggestion
                    var serializer = new JsonSerializer();
```

src/Microsoft.ComponentDetection.Detectors/spdx/Spdx22ComponentDetector.cs
```diff
@@ -0,0 +1,119 @@
...
+
+                    try
+                    {
+                        JObject document = (JObject)serializer.Deserialize(reader);
```
JamieMagee, line 55: nit: there's a generic version of this method available
```suggestion
                        var document = serializer.Deserialize<JObject>(reader);
```

JamieMagee (approved):

ByAgenT:
I see that VerificationTests are failing on windows. Is there something that I missed and I need to address?

JamieMagee:
I don't think so. We've seen some flakiness regarding the verification tests on Windows recently. I know GitHub actions recently changed `windows-latest` from Windows Server 2019 to Windows Server 2022, and I suspect that might be part of the reason.

Let me try re-running and see if it passes.

PR #84, opened 2022-03-24
docs: document environment variables

JamieMagee:
Closes #80


cobya (approved):

cobya:
> 👋 Hi! It looks like you modified some files in the `Detectors` folder. You may need to bump the detector versions if any of the following scenarios apply:
> 
> * The detector detects more or fewer components than before
> * The detector generates different parent/child graph relationships than before
> * The detector generates different `devDependencies` values than before
> 
> If none of the above scenarios apply, feel free to ignore this comment 🙂

Looks like we need to fix the folder check...

PR #96, opened 2022-04-06
Bug Fix: While parsing maven scope, discard left-over string if present.

RushabhBhansali:
## Overview:
A bug was introduced in [this](https://github.com/microsoft/component-detection/commit/3a17feb7e1f5cf8acefc03a53adf932a2a5b7b5c) commit, 
which resulted in invalid scope if there was left over string in the scope substring.

an example [build break](https://github.com/microsoft/componentdetection-azdo/runs/5858218730?check_suite_focus=true) as a result of this was observed in azdo. 

## Testing
- new unit test for this special parsing use case.
- Tested locally to detect maven components. 
- All existing unit tests successfully passed.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/maven/MavenParsingUtilities.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/maven/MavenParsingUtilities.cs
@@ -59,7 +59,7 @@ namespace Microsoft.ComponentDetection.Detectors.Maven
             if (results.Length == 5)
             {
                 dependencyScope = MavenScopeToDependencyScopeMapping.TryGetValue(
-                    Regex.Match(results[4], @"^([\w\-]+)").Value,
+                    Regex.Match(results[4], @"^([\w]+)").Value,
                     out dependencyScope) 
                     ? dependencyScope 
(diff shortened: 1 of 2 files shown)


src/Microsoft.ComponentDetection.Detectors/maven/MavenParsingUtilities.cs
```diff
@@ -57,7 +58,9 @@ private static (string groupId, string artifactId, string version, bool? isDevel
             bool? isDevDependency = null;
             if (results.Length == 5)
             {
-                dependencyScope = MavenScopeToDependencyScopeMapping.TryGetValue(results[4], out dependencyScope) 
+                dependencyScope = MavenScopeToDependencyScopeMapping.TryGetValue(
+                    Regex.Match(results[4], @"^([\w\-]+)").Value,
```
cobya, line 62: Why is this RegEx matching the `-` character along with the word?

RushabhBhansali: left over from the pattern I referred. Removed the matching with "-"

cobya (approved):

PR #104, opened 2022-04-17
Adds support for pubspec.lock files

jhutchings1:
This PR adds support for pubspec.lock files, which are used in Flutter and Dart projects.


src/Microsoft.ComponentDetection.Detectors/pub/PubComponentDetector.cs
```diff
@@ -0,0 +1,68 @@
...
+using Microsoft.ComponentDetection.Contracts;
+using Microsoft.ComponentDetection.Contracts.Internal;
+using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using YamlDotNet.Serialization;
+using YamlDotNet.Serialization.NamingConventions;
+
+namespace Microsoft.ComponentDetection.Detectors.Pub
+{
+    [Export(typeof(IComponentDetector))]
+    public class PubComponentDetector : FileComponentDetector
```
cobya, line 16: New detectors should initially inherit the `IDefaultOffComponentDetector` class until validation can be performed.
```suggestion
    public class PubComponentDetector : FileComponentDetector, IDefaultOffComponentDetector 
```

src/Microsoft.ComponentDetection.Detectors/pub/PubComponentDetector.cs
```diff
@@ -0,0 +1,68 @@
...
+            NeedsAutomaticRootDependencyCalculation = true;
+        }
+
+        protected override Task OnFileFound(ProcessRequest processRequest, IDictionary<string, string> detectorArgs)
+        {
+            var singleFileComponentRecorder = processRequest.SingleFileComponentRecorder;
+            var file = processRequest.ComponentStream;
+
+            Logger.LogVerbose("Found Pubspec.lock: " + file.Location);
+            ParsePubspecYamlFile(singleFileComponentRecorder, file);
```
cobya, line 39: Can this be wrapped in a try-catch block with a LogFailedReadingFile call?
```
catch (Exception e)
{
   Logger.LogFailedReadingFile(file.Location, e);
}
```

src/Microsoft.ComponentDetection.Detectors/pub/PubComponentDetector.cs
```diff
@@ -0,0 +1,68 @@
...
+            using (var reader = new StreamReader(file.Stream))
+            {
+                text = reader.ReadToEnd();
+            }
+
+            var deserializer = new YamlDotNet.Serialization.DeserializerBuilder()
+                .WithNamingConvention(UnderscoredNamingConvention.Instance)
+                .Build();
+
+            dynamic pubspec = deserializer.Deserialize<System.Dynamic.ExpandoObject>(text);
```
cobya, line 58: Can the deserialization contract be captured in a class like https://github.com/microsoft/component-detection/blob/main/src/Microsoft.ComponentDetection.Detectors/pnpm/PnpmYaml.cs so we aren't using dynamic objects?

src/Microsoft.ComponentDetection.Detectors/pub/PubComponentDetector.cs
```diff
@@ -0,0 +1,68 @@
...
+            Console.WriteLine(pubspec.packages);
```
cobya, line 59: If we are keeping this console output, please modify to be a `Logger.LogVerbose` call

test/Microsoft.ComponentDetection.Detectors.Tests/PubDetectorTests.cs
```diff
@@ -0,0 +1,118 @@
...
+            Assert.AreEqual(6, detectedComponents.Count());
+
+            AssertPubComponentNameAndVersion(detectedComponents, "aligned_dialog", "0.0.6");
+            AssertPubComponentNameAndVersion(detectedComponents, "analyzer", "3.4.1");
+            AssertPubComponentNameAndVersion(detectedComponents, "archive", "3.3.0");
+            AssertPubComponentNameAndVersion(detectedComponents, "args", "2.3.0");
+            AssertPubComponentNameAndVersion(detectedComponents, "async", "2.8.2");
+            AssertPubComponentNameAndVersion(detectedComponents, "boolean_selector", "2.1.0");
+        }
+
```
cobya, line 108: Can you add a few tests here to check invalid input to verify the detector handles them correctly?

cobya (changes requested):
Apologies for the late review here, our team has been managing other priorities.

src/Microsoft.ComponentDetection.Detectors/pub/PubComponentDetector.cs
jhutchings1: I tried that in an earlier commit, and my testing was running into parsing failures because the lock files weren't consistent enough. The dynamic object seems to be more reliable, but I may have missed something. https://github.com/jhutchings1/component-detection/commit/829c548ffd19fedd6a3c294ad4cf59c410b15128

test/Microsoft.ComponentDetection.Detectors.Tests/PubDetectorTests.cs
jhutchings1: I added one test for invalid input. If you've got ideas on other scenarios that'd be useful, let me know!

jhutchings1:
@cobya I think I've addressed most of your outstanding feedback on this one. Do you want to have another look?

.github/workflows/build.yml
```diff
@@ -3,6 +3,7 @@ name: Build
 on:
   push:
   pull_request:
+  workflow_dispatch:
```
cobya, line 6: Can you please exclude the changes from .github and .vscode from this PR? Happy to add them in a follow up but would prefer to not add in work outside of the PR scope.

cobya (approved):

JamieMagee:
We should also add some test cases to [verification tests](https://github.com/microsoft/component-detection/tree/main/test/Microsoft.ComponentDetection.VerificationTests/resources)

.github/workflows/build.yml
jhutchings1: Fixed.

JamieMagee (approved):

FernandoRojo:
Going through some older PRs now, do we still want this to get merged in @jhutchings1

FernandoRojo:
Closing this PR just because it's duplicated in https://github.com/microsoft/component-detection/pull/104, 

But I can merge the code from this PR into there if we can confirm that there haven't been breaking changes/ it is still needed.

PR #105, opened 2022-04-18
Go CLI detector enhancement (`go list -m all`)

grvillic:
## Problem
Currently for Golang projects component detection works by scanning the go.sum file or using `go mod graph` command as an alternative, but results should be the same as using go.sum file because go mod graph prints out all the versions in the graph. This means we are over detecting modules of versions that are no longer included in the build binaries.

## Solution
Use [go list -m all](https://go.dev/ref/mod#go-list-m) command to find the actual versions of the components [used during a build](https://go.dev/ref/mod#glos-build-list) as it uses the [minimal version selection](https://go.dev/ref/mod#minimal-version-selection) algorithm. That command does not generate a graph for us, hence by running both `go mod graph` and `go list -m all`, we are able to prune all the dependencies that were not included in the build list and keep the dependency graph in of Go's main module.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
@@ -261,14 +261,17 @@ namespace Microsoft.ComponentDetection.Detectors.Go
 
         private void RecordBuildDependencies(string goListOutput, ISingleFileComponentRecorder singleFileComponentRecorder)
         {
-            // Yes, go always returns \n even on Windows
-            var goListObjects = goListOutput.Split("}\n{").Where(x => !string.IsNullOrWhiteSpace(x));
+            var goBuildModules = new List<GoBuildModule>();
+            var reader = new JsonTextReader(new StringReader(goListOutput));
+            reader.SupportMultipleContent = true;
 
-            // The json output has multiple roots. We need to update it to make it a valid string
-            // Namely do the following conversion { "propA": 1}{"propA": 2} => [{ "propA": 1},{"propA": 2}]
-            // Yes, go always returns \n even on Windows
-            var validJsonList = $"[{string.Join("},{", goListObjects)}]";
-            var goBuildModules = JsonConvert.DeserializeObject<List<GoBuildModule>>(validJsonList);
+            while (reader.Read())
+            {
+                var serializer = new JsonSerializer();
+                var buildModule = serializer.Deserialize<GoBuildModule>(reader);
+
+                goBuildModules.Add(buildModule);
+            }
 
             foreach (var dependency in goBuildModules)
             {
--- a/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
@@ -9,6 +9,7 @@ using Microsoft.ComponentDetection.Common.Telemetry.Records;
 using Microsoft.ComponentDetection.Contracts;
 using Microsoft.ComponentDetection.Contracts.Internal;
 using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using Newtonsoft.Json;
 
 namespace Microsoft.ComponentDetection.Detectors.Go
 {
@@ -33,7 +34,7 @@ namespace Microsoft.ComponentDetection.Detectors.Go
 
         public override IEnumerable<ComponentType> SupportedComponentTypes { get; } = new[] { ComponentType.Go };
 
-        public override int Version => 4;
+        public override int Version => 5;
 
         private HashSet<string> projectRoots = new HashSet<string>();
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -264,5 +308,16 @@ private bool IsGoCliManuallyEnabled()
         {
             return EnvVarService.DoesEnvironmentVariableExist("EnableGoCliScan");
         }
+
+        private class GoBuildModule
+        {
+            public string Path { get; set; }
+
+            public bool Main { get; set; }
```
cobya, line 316: Are the non-nullable types here guaranteed to exist?

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -104,22 +105,31 @@ private async Task<bool> UseGoCliToScan(string location, ISingleFileComponentRec
...
-            var isGoAvailable = await CommandLineInvocationService.CanCommandBeLocated("go", null, workingDirectory: projectRootDirectory, new List<string> { "version" }.ToArray());
+            var isGoAvailable = await CommandLineInvocationService.CanCommandBeLocated("go", null, workingDirectory: projectRootDirectory, new[] { "version" });
             record.IsGoAvailable = isGoAvailable;
 
             if (!isGoAvailable)
             {
                 return false;
             }
 
+            var goDependenciesProcess = await CommandLineInvocationService.ExecuteCommand("go", null, workingDirectory: projectRootDirectory, new[] { "list", "-m", "-json", "all" });
```
cobya, line 116: Do you have an expected execution time for this? I know `go mod graph` has caused issues for us in the past.

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -247,6 +254,43 @@ private void PopulateDependencyGraph(string goGraphOutput, ISingleFileComponentR
...
         }
 
+        private bool IsModuleInBuildList(ISingleFileComponentRecorder singleFileComponentRecorder, GoComponent component)
+        {
+            return singleFileComponentRecorder.GetComponent(component.Id) != null;
+        }
+
+        private void RecordBuildDependencies(string goListOutput, ISingleFileComponentRecorder singleFileComponentRecorder)
+        {
+            // Yes, go always returns \n even on Windows
```
cobya, line 264: Is this documented anywhere in the Go docs?

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
grvillic: Yes to most of them, `Main` property could be missing, because Go doesn't include in their serialization the default value (false).

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
grvillic: The problem is not `go mod graph` itself IIRC, the problem is when users have `go.sum` and `go.mod` files in their repo, but they actually aren't building it. Go CLI will try to fetch the packages behind the scenes if they aren't in the system already, which will add time to any automated process. Execution time should be the about the same, but now `go list` takes the hit of fetching the packages, rather than `go mod graph`. 

At the end of the day, we are using this command because we do not have any other native way to accurately detect components in Go. I tested with a big repo reporting ~730 dependencies using only `go mod graph` vs the latest changes that only pick up 137 dependencies that are actually included in their Go binaries.

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
grvillic: Nope (they lack quite a bit of doc tbh, and what's out there can be confusing), me running the commands in windows saw the line endings.

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -247,6 +254,43 @@ private void PopulateDependencyGraph(string goGraphOutput, ISingleFileComponentR
...
+            var goListObjects = goListOutput.Split("}\n{").Where(x => !string.IsNullOrWhiteSpace(x));
+
+            // The json output has multiple roots. We need to update it to make it a valid string
+            // Namely do the following conversion { "propA": 1}{"propA": 2} => [{ "propA": 1},{"propA": 2}]
```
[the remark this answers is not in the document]
grvillic: I wasn't aware of that multi-fragment support in library, that's a nifty utility.

cobya (approved):

AzureMarker:
Thanks for implementing this! I look forward to testing it.

PR #106, opened 2022-04-18
Add Paket support

jhutchings1:
This PR adds support for the [paket.lock format](https://fsprojects.github.io/Paket/lock-file.html) to the existing Nuget detector.


amitla1 (approved):

JamieMagee:
@amitla1 Can you update the version of this detector please?

https://github.com/microsoft/component-detection/blob/3254e5da79fd4273b8eb9ec0c1b84df427172129/src/Microsoft.ComponentDetection.Detectors/nuget/NuGetComponentDetector.cs#L31

It needs to be updated any time the detector is expected to return a different number of dependencies.

PR #113, opened 2022-04-29
Enable GoCliDetector by default. Using env variable 'DisableGoCliDetector=true" to manually disable GoCliDetector.

RushabhBhansali:
## Summary:
Enabling Go cli detector by default.

## Details
Previously, the Go-Detector by default scanned the manifest and generated components. We were using `EnableGoCliScan` env. variable to activate the Go Cli Detector. With this change, the use of `EnableGoCliScan` is removed. The Go detector by default uses Cli scan. 

To manually override this behavior, new env. variable `DisableGoCliScan` is introduced. 

## How was this tested. 
- Unit tests
- Locally, ran the component-detection on [componentdetection-verification/tree/main/projects/go/helm](https://github.com/microsoft/componentdetection-verification/tree/main/projects/go/helm)
  - without setting the env. variable, validated that go cli scan was ran. approx. 400 components generated. 
  - when `DisableGoCliScan=true` env. variable, validated that GoDetector with manifest scan was ran. Approx. 700 component generated. 

## Why is verification test failing in this PR ? 
This PR changes default go detection to use the Go cli. This results in overall noise reduction in the detected go components. 
The verification test uses  [windows-2022](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md) which includes Go cli. Hence its expected that verification test will fail for go detection.

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -2,10 +2,10 @@
 
 Environment variables are sometimes used to control experimental features or advanced options
 
-## `EnableGoCliScan`
+## `DisableGoCliScan`
 
-If the environment variable `EnableGoCliScan` is set, to any value, the Go detector uses [`go mod graph`][1] to discover Go dependencies.
-If the environment variable is not set, we fall back to parsing `go.mod` and `go.sum` ourselves.
+If the environment variable `DisableGoCliScan` is set to "true", we fall back to parsing `go.mod` and `go.sum` ourselves. 
+Otherwise, the Go detector uses go-cli command: `go list -m all` to discover Go dependencies.
 
 ## `PyPiMaxCacheEntries`
 
--- a/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
@@ -59,7 +59,7 @@ namespace Microsoft.ComponentDetection.Detectors.Go
                 } else
                 {
                     Logger.LogInfo("Go cli scan was manually disabled, fallback strategy performed." +
-                        " More info: https://docs.opensource.microsoft.com/tools/cg/reference/feature-coverage/#go");
+                        " More info: https://github.com/microsoft/component-detection/blob/main/docs/detectors/go.md#fallback-detection-strategy");
                 }
             }
             catch (Exception ex)
@@ -118,7 +118,7 @@ namespace Microsoft.ComponentDetection.Detectors.Go
             }
 
             Logger.LogInfo("Go CLI was found in system and will be used to generate dependency graph. " +
-                "Detection time may be improved by activating fallback strategy (https://docs.opensource.microsoft.com/tools/cg/reference/feature-coverage/#go). " +
+                "Detection time may be improved by activating fallback strategy (https://github.com/microsoft/component-detection/blob/main/docs/detectors/go.md#fallback-detection-strategy). " +
                 "But, it will introduce noise into the detected components.");
             var goDependenciesProcess = await CommandLineInvocationService.ExecuteCommand("go", null, workingDirectory: projectRootDirectory, new[] { "list", "-m", "-json", "all" });
             if (goDependenciesProcess.ExitCode != 0)
--- a/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
@@ -56,7 +56,8 @@ namespace Microsoft.ComponentDetection.Detectors.Go
                 if (!IsGoCliManuallyDisabled())
                 {
                     wasGoCliScanSuccessful = await UseGoCliToScan(file.Location, singleFileComponentRecorder);
-                } else
+                }
+                else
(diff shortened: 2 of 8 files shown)


docs/environment-variables.md
```diff
@@ -2,10 +2,10 @@
 
 Environment variables are sometimes used to control experimental features or advanced options
 
-## `EnableGoCliScan`
+## `DisableGoCliScan`
 
-If the environment variable `EnableGoCliScan` is set, to any value, the Go detector uses [`go mod graph`][1] to discover Go dependencies.
-If the environment variable is not set, we fall back to parsing `go.mod` and `go.sum` ourselves.
+If the environment variable `DisableGoCliScan` is set to "true", we fall back to parsing `go.mod` and `go.sum` ourselves. 
+Otherwise, the Go detector uses [`go mod graph`][1] to discover Go dependencies.
```
grvillic, line 8: Note: the detector now uses `go list -m all` to discover dependencies by `go mod graph` to find relationships. We should also update the [Go doc](https://github.com/microsoft/component-detection/blob/main/docs/detectors/go.md)

RushabhBhansali: updated the doc,

src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs
```diff
@@ -247,6 +247,9 @@ private async Task<ScanResult> Dispatch(IScanArguments arguments, CancellationTo
                     catch (TimeoutException timeoutException)
                     {
                         Logger.LogError(timeoutException.Message);
+
+                        // TODO: remove the below message after go cli detector improvement has been stabilized.  
+                        Logger.LogWarning("If your repository contains Go packages, your build time may be improved by activating fallback strategy: https://aka.ms/go-detection-fallback-strategy");
```
grvillic, line 252: A similar message should also be added to the Go Detector. They might not have timed out, but if their pipeline is taking a lot longer than before, they should still be able to see a log about it. We should print `your build time may be improved by activating fallback strategy: https://aka.ms/go-detection-fallback-strategy` whenever Go CLI was found in system within the detector itself.

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -52,10 +53,12 @@ protected override async Task OnFileFound(ProcessRequest processRequest, IDictio
             var wasGoCliScanSuccessful = false;
             try
             {
-                if (IsGoCliManuallyEnabled())
+                if (!IsGoCliManuallyDisabled())
                 {
-                    Logger.LogInfo("Go cli scan was manually enabled");
```
grvillic, line 57: We should keep this log, Go CLI was found in system and will be used to generate dependency graph. Detection time can be improved by activating fallback strategy (https://aka.ms/go-detection-fallback-strategy), but it will introduce noise into the detected component list.

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -52,10 +53,12 @@ protected override async Task OnFileFound(ProcessRequest processRequest, IDictio
...
                     wasGoCliScanSuccessful = await UseGoCliToScan(file.Location, singleFileComponentRecorder);
+                } else
+                {
+                    Logger.LogInfo("Go cli scan was manually disabled.");
```
grvillic, line 61: "Go cli scan was manually disabled, fallback strategy performed. For more info see https://aka.ms/go-detection-fallback-strategy"

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
RushabhBhansali: added the suggested logging statement in `USeGoCliToScan` only after validating that go-cli exist in the environment.

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
RushabhBhansali: updated.

src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs
RushabhBhansali: added the suggested log.

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -52,10 +53,13 @@ protected override async Task OnFileFound(ProcessRequest processRequest, IDictio
             var wasGoCliScanSuccessful = false;
             try
             {
-                if (IsGoCliManuallyEnabled())
+                if (!IsGoCliManuallyDisabled())
                 {
-                    Logger.LogInfo("Go cli scan was manually enabled");
                     wasGoCliScanSuccessful = await UseGoCliToScan(file.Location, singleFileComponentRecorder);
+                } else
```
grvillic, line 59: nit
```suggestion
                }
                else
```

grvillic (approved):

PR #114, opened 2022-04-30
Transcribe docker reference parsing and introduce dockerFile detector

FernandoRojo:


src/Microsoft.ComponentDetection.Common/DockerReference/DockerRegex.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        public static Regex NameRegexp = Expression(
+            Optional(
+                DomainRegexp,
+                new Regex(@"\/"),
+                NameComponentRegexp,
+                Optional(
+                    Repeated(
+                        new Regex(@"\/"),
+                        NameComponentRegexp))));
```
JamieMagee, line 90: This first `Optional` encapsulates too much.

JamieMagee:
Needs tests. I wrote up some, feel free to copy from [here](https://github.com/JamieMagee/DockerReference/tree/main/test/JamieMagee.DockerReference.Test)

src/Microsoft.ComponentDetection.Detectors/dockerfile/DockerfileComponentDetector.cs
```diff
@@ -0,0 +1,154 @@
...
+        public ICommandLineInvocationService CommandLineInvocationService { get; set; }
+
+        [Import]
+        public IEnvironmentVariableService EnvVarService { get; set; }
+
+        public override string Id { get; } = "dockerfile";
+
+        public override IEnumerable<string> Categories => new[] { Enum.GetName(typeof(DetectorClass), DetectorClass.GoMod) };
+
+        public override IList<string> SearchPatterns { get; } = new List<string> { "dockerfile", "dockerfile.*" };
```
JamieMagee, line 29: `*.dockerfile` is also sometimes used.

src/Microsoft.ComponentDetection.Detectors/dockerfile/DockerfileComponentDetector.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        public override IEnumerable<ComponentType> SupportedComponentTypes { get; } = new[] { ComponentType.Dockerfile };
+
+        public override int Version => 5;
```
JamieMagee, line 33: Version 5?

src/Microsoft.ComponentDetection.Detectors/dockerfile/DockerfileComponentDetector.cs
```diff
@@ -0,0 +1,154 @@
...
+        private IDockerReference ProcessDockerfileConstruct(DockerfileConstruct construct, char escapeChar, Dictionary<string, string> stageNameMap)
+        {
+            var instructionKeyword = construct.Type;
+            IDockerReference baseImage = null;
+            if (instructionKeyword == ConstructType.Instruction)
+            {
+                var constructType = construct.GetType().ToString();
+                switch (constructType)
+                {
+                    case "Valleysoft.DockerfileModel.FromInstruction":
```
JamieMagee, line 79: Magic strings? Is there an exported type you can use instead?

src/Microsoft.ComponentDetection.Common/DockerReference/DigestUtility.cs
```diff
@@ -0,0 +1,54 @@
...
+    public class DigestUtility
+    {
+        private static readonly Dictionary<string, int> AlgorithmsSizes = new Dictionary<string, int>()
+        {
+            { "sha256", 32 },
+            { "sha384", 48 },
+            { "sha512", 64 },
+        };
+
+        public static bool CheckDigest(string digest, bool throwError = true)
```
JamieMagee, line 14: I think the C# equivalent would be to have `TryCheckDigest(string digest)` which returns `bool`, and `CheckDigest(string digest)` which can throw.

src/Microsoft.ComponentDetection.Contracts/DockerReference.cs
```diff
@@ -0,0 +1,132 @@
...
+    {
+        public override string type { get; } = "dual";
+        public string domain;
+        public string repository;
+        public string tag;
+        public string digest;
+
+        public override string ToString()
+        {
+            return $"{this.repository}:${this.tag}@${this.digest}";
```
JamieMagee, line 118: Missing `domain` here.

src/Microsoft.ComponentDetection.Contracts/DockerReference.cs
```diff
@@ -0,0 +1,132 @@
+namespace Microsoft.ComponentDetection.Contracts
+{
+    #pragma warning disable SA1402
+    public abstract class IDockerReference
+    {
+        public abstract string type { get; }
```
JamieMagee, line 6: Should this be an enum?

src/Microsoft.ComponentDetection.Contracts/DockerReference.cs
```diff
@@ -0,0 +1,132 @@
...
+
+    // docker.io/library/ubuntu
+    public class RepositoryReference : IDockerReference
+    {
+        public override string type { get; } = "repository";
+        public string domain;
+        public string repository;
+        public override string ToString()
+        {
+            return $"{this.repository}";
```
JamieMagee, line 71: Missing `domain`

src/Microsoft.ComponentDetection.Detectors/dockerfile/DockerfileComponentDetector.cs
```diff
@@ -0,0 +1,154 @@
...
+                }
+            }
+            
+            return Task.CompletedTask;
+        }
+
+        private IDockerReference ProcessDockerfileConstruct(DockerfileConstruct construct, char escapeChar, Dictionary<string, string> stageNameMap)
+        {
+            var instructionKeyword = construct.Type;
+            IDockerReference baseImage = null;
```
JamieMagee, line 73:
```suggestion
            var baseImage = null;
```

src/Microsoft.ComponentDetection.Detectors/dockerfile/DockerfileComponentDetector.cs
```diff
@@ -0,0 +1,154 @@
...
+                return null;
+            }
+
+            return DockerReferenceUtility.ParseFamiliarName(reference);
+        }
+
+        private bool HasUnresolvedVariables(string reference)
+        {
+            return new Regex("[${}]").IsMatch(reference);
+        }
```
JamieMagee, line 152: I'm not sure you need this in the C# port. It was to work around limitations in node's regex.

JamieMagee:
Note on naming: Docker -> Container
Also look for `Containerfile`

src/Microsoft.ComponentDetection.Detectors/dockerfile/DockerfileComponentDetector.cs
FernandoRojo: Can't be done since it can't infer the type from null

src/Microsoft.ComponentDetection.Detectors/dockerfile/DockerfileComponentDetector.cs
FernandoRojo: There is, but Switch cases based on Type are in preview, and apparently not widely accepted: https://docs.microsoft.com/en-us/archive/blogs/jaredpar/switching-on-types

PR #117, opened 2022-05-08
Change Rust/Cargo detector to be lock file based

### LINKED ISSUE #116: Rust component detection fails to detect all Rust crate usage
The current implementation of the Rust detectors use a combination of `Cargo.lock` and `Cargo.toml` information. The `Cargo.lock` file is used to understand what edges exist within the dependency graph, while the `Cargo.toml` files are found and used to walk the graph to detect component usage.

The manifest format within `Cargo.toml` is complex, supporting many features including workspaces, conditional dependencies (including per-target dependencies), package renaming, etc. Today, there are at least two significant issues where components are not correctly detected:
1) Workspace members added indirectly via path dependencies from other workspace members are not found or processed
2) Target conditional dependencies in crate manifests are not found or processed.

While these issues could be resolved, the current approach requires us to actively track any changes or enhancements to `Cargo.toml`.

Instead, we should consider taking a more conservative approach that relies solely on the `Cargo.lock` file, which is a simpler and significantly more stable format. This will remove our current ability to identify dev-dependencies (as this information is not in the lock file) but will simplify the implementation of the detector and address the issues mentioned above. In the future, we could then (optionally) call out to the package manager directly (e.g., calling `cargo metadata`) to provide more detailed information about dependencies without relying on the presence of Cargo for component detection.

danielframpton:
This is a significant change to Rust component detection to solely use the `Cargo.lock` file to detect component usage (e.g., no longer processing `Cargo.toml`).

As discussed in #116 there are several issues with the existing `Cargo.toml` processing that this change sidesteps by reporting all crates from `Cargo.lock`.

This change also explicitly includes the package registry in the identifier for components. I have not investigated what implications there may be for this to downstream tools, but because Cargo has support for multiple registries the package name and version number is insufficient to identify the package (because of multiple registries or a local package with the same name as one published in the registry).

In general, the sole use of the `Cargo.lock` file should strictly increase the number of crates reported, although I did observe some reductions that were due to this change no longer reporting some local/path crates as components (because they had a version and path specified) but I believe this was unintentional/unexpected.

There was also significant duplication in the code between v1 and v2 detectors, but in rewriting the logic to process the lock file I was able to use the same logic for both formats.

I expect this to need a careful review and some changes before landing, but I wanted to send out the PR to make some progress on getting these issues resolved.

Fixes #116


src/Microsoft.ComponentDetection.Contracts/TypedComponent/CargoComponent.cs
```diff
@@ -9,20 +9,38 @@ private CargoComponent()
...
+
         public override ComponentType Type => ComponentType.Cargo;
 
-        public override string Id => $"{Name} {Version} - {Type}";
+        public override string Id
+        {
+            get {
+                if (Source != null)
+                {
+                    return $"{Name} {Version} {Source} - {Type}";
```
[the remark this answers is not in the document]
arlosi: If it does make sense to keep `Source` in the `Id` string, we may want to put it in parentheses `({Source})`, which better matches what's in the lock file.

src/Microsoft.ComponentDetection.Detectors/rust/RustCrateDetector.cs
```diff
@@ -34,61 +36,207 @@ protected override Task OnFileFound(ProcessRequest processRequest, IDictionary<s
...
+            }
+
+            return match.Success;
+        }
+
+        private static bool IsLocalPackage(CargoPackage package) => package.source == null;
+
+        private const string CargoLockSearchPattern = "Cargo.lock";
+
+        private static string MakeDependencyStringFromPackage(CargoPackage package)
```
arlosi, line 237: Would this make more sense as an instance method on `CargoPackage` ? Maybe `IsLocalPackage` too?

src/Microsoft.ComponentDetection.Detectors/rust/RustCrateDetector.cs
```diff
@@ -34,61 +36,207 @@ protected override Task OnFileFound(ProcessRequest processRequest, IDictionary<s
...
+        {
+            return $"{package.name} {package.version} ({package.source})";
```
arlosi, line 239: Can `package.source` be `null`?

arlosi (commented):
I'm not an expert in CG, but I think this makes sense to get Cargo support to stop missing dependencies as mentioned in #116 - even if it means we give up dev-dependency detection for now.

src/Microsoft.ComponentDetection.Detectors/rust/RustCrateDetector.cs
```diff
@@ -34,61 +36,207 @@ protected override Task OnFileFound(ProcessRequest processRequest, IDictionary<s
...
+
+                record.PackageInfo = $"{parentPackage.name}, {parentPackage.version}, {parentPackage.source}";
+                record.Dependencies = dependency;
+
                 Logger.LogFailedReadingFile(cargoLockFile.Location, e);
             }
+        }
 
-            return Task.CompletedTask;
+        private static readonly Regex DependencyFormatRegex = new Regex(
```
arlosi, line 209: Should this regex include start/end markers `^$` to prevent partially matching a string?

src/Microsoft.ComponentDetection.Contracts/TypedComponent/CargoComponent.cs
```diff
@@ -9,20 +9,38 @@ private CargoComponent()
             // reserved for deserialization
         }
 
-        public CargoComponent(string name, string version)
+        public CargoComponent(string name, string version, string checksum = null, string source = null)
```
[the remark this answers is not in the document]
danielframpton: Yes.

src/Microsoft.ComponentDetection.Detectors/rust/RustCrateDetector.cs
danielframpton: Added

src/Microsoft.ComponentDetection.Detectors/rust/RustCrateDetector.cs
danielframpton: Removed this method.

src/Microsoft.ComponentDetection.Detectors/rust/RustCrateDetector.cs
danielframpton: No (note: this code is now gone).

src/Microsoft.ComponentDetection.Contracts/TypedComponent/CargoComponent.cs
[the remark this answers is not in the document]
danielframpton: `Id` is used for identity, so if it is not included in this string, then matching nodes will be merged as part of the detection.

I have added the parentheses as suggested. Still waiting to validate if/how this might cause issues for other tools downstream.

src/Microsoft.ComponentDetection.Detectors/rust/Contracts/CargoPackage.cs
```diff
@@ -30,8 +30,10 @@ public override bool Equals(object obj)
...
             return package != null &&
-                   name.Equals(package.name) &&
-                   version.Equals(package.version, StringComparison.OrdinalIgnoreCase);
+                   string.Equals(name, package.name) &&
+                   string.Equals(version, package.version, StringComparison.OrdinalIgnoreCase) &&
+                   string.Equals(source, package.source) &&
+                   string.Equals(checksum, package.checksum);
         }
 
         public override int GetHashCode()
```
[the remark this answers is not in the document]
danielframpton: Added

juvazq:
Hey @arlosi and @danielframpton, is there anything else blocking this pr? It would be nice to have component detection in good shape for rust. Thanks!

arlosi:
It needs review from the maintainers. Neither @danielframpton nor I can approve it.

juvazq:
Thanks @arlosi, 

@jcfiorenzano @cobya do you think this can be merged? Do you miss something before it can be approved?

danielframpton:
I have simplified this PR to focus on the source of dependencies (lock vs toml) deferring the question of multiple registries to see if we can land the highest priority part this change.

src/Microsoft.ComponentDetection.Detectors/rust/RustCrateDetector.cs
```diff
@@ -34,61 +38,200 @@ protected override Task OnFileFound(ProcessRequest processRequest, IDictionary<s
...
+                    {
+                        // Get or create the list of packages with this name
+                        if (packagesByName.TryGetValue(cargoPackage.name, out var packageList))
+                        {
+                            if (packageList.Any(p => p.package.Equals(cargoPackage)))
+                            {
+                                // Ignore duplicate packages
+                                continue;
+                            }
+                        }
```
JamieMagee, line 57: Should there be an `else` here? Or can these two conditionals be combined?

src/Microsoft.ComponentDetection.Detectors/rust/RustCrateDetector.cs
```diff
@@ -34,61 +38,200 @@ protected override Task OnFileFound(ProcessRequest processRequest, IDictionary<s
...
+            ISingleFileComponentRecorder singleFileComponentRecorder,
+            HashSet<CargoPackage> seenAsDependency,
+            Dictionary<string, List<(CargoPackage package, CargoComponent component)>> packagesByName,
+            CargoPackage parentPackage,
+            CargoComponent parentComponent,
+            string dependency)
+        {
+            try
+            {
+                // Extract the informationfrom the dependency (name with optional version and source)
```
JamieMagee, line 131:
```suggestion
                // Extract the information from the dependency (name with optional version and source)
```

src/Microsoft.ComponentDetection.Detectors/rust/RustCrateDetector.cs
```diff
@@ -34,61 +38,200 @@ protected override Task OnFileFound(ProcessRequest processRequest, IDictionary<s
...
+        private const int PackageNameGroup = 1;
+        private const int VersionGroup = 2;
+        private const int SourceGroup = 3;
+
+        private static bool ParseDependency(string dependency, out string packageName, out string version, out string source)
+        {
+            var match = DependencyFormatRegex.Match(dependency);
+            packageName = match.Groups[PackageNameGroup].Success ? match.Groups[PackageNameGroup].Value : null;
+            version = match.Groups[VersionGroup].Success ? match.Groups[VersionGroup].Value : null;
+            source = match.Groups[SourceGroup].Success ? match.Groups[SourceGroup].Value : null;
```
JamieMagee, line 225: nit: named groups[^1]

```suggestion
           @"^(?<packageName>[^ ]+)(?: (?<version>[^ ]+))?(?: \((?<source>[^()]*)\))?$",
           RegexOptions.Compiled);

        private static bool ParseDependency(string dependency, out string packageName, out string version, out string source)
        {
            var match = DependencyFormatRegex.Match(dependency);
            packageName = match.Groups["packageName"].Success ? match.Groups["packageName"].Value : null;
            version = match.Groups["version"].Success ? match.Groups["version"].Value : null;
            source = match.Groups["source"].Success ? match.Groups["source"].Value : null;
```

[^1]: https://docs.microsoft.com/en-us/dotnet/standard/base-types/grouping-constructs-in-regular-expressions#named_matched_subexpression

JamieMagee:
@danielframpton are the [verification tests for rust](https://github.com/microsoft/component-detection/tree/main/test/Microsoft.ComponentDetection.VerificationTests/resources/rust) still sufficient, or do they need updating as well?

JamieMagee (approved):
Approved pending comments

JamieMagee:
Snapshot tests are expected to fail due to changes in the detector.

arlosi (approved):

JamieMagee:
@danielframpton Do you have time to address the comments and resolve conflicts?

src/Microsoft.ComponentDetection.Detectors/rust/RustCrateDetector.cs
```diff
@@ -34,61 +38,200 @@ protected override Task OnFileFound(ProcessRequest processRequest, IDictionary<s
...
+                    // Pass 3: Conservatively mark packages we found no dependency to as roots
+                    foreach (var packageList in packagesByName.Values)
+                    {
+                        // Get the package and component.
+                        foreach (var (package, component) in packageList)
+                        {
+                            if (!IsLocalPackage(package) && !seenAsDependency.Contains(package))
+                            {
+                                var detectedComponent = new DetectedComponent(component);
+                                singleFileComponentRecorder.RegisterUsage(detectedComponent, isExplicitReferencedDependency: true);
```
grvillic, line 105: What happened to dev dependency detection? Is that being considered in these updates, if so, how will those be treated now? Users won't like to start picking up dependencies as regular build time dependencies when they used to be dev dependencies.

danielframpton: This does not report dev dependencies separately (e.g., they are included alongside other dependencies as well) as that information is not currently included in the lock file. Unfortunately, there isn't an obvious middle ground here between having something that is simple and robust (lock file processing) or replicating the full functionality of cargo.

This change puts us in the position of not missing dependencies, with the option to (in the future) introduce a tool that uses cargo to get a more accurate picture (either directly or by consuming SBOMs generated from cargo).

katshup:
@danielframpton Do we believe this PR is on track for September?

JamieMagee:
@danielframpton thanks for deconflicting this PR ❤️ 

I just wanted to double check with you that you're good with this being merged?

danielframpton:
Thanks @JamieMagee!

I am, I have responded to all the feedback and updated across the code analysis fixes. Let me know if there is anything I have missed.

PR #123, opened 2022-06-10
Add support for newer RPM based distros (Mariner 2.0) 

tofay:
Update syft to 0.47.0 which supports newer RPM db formats as used in Mariner 2.0, fedora 33+, etc

I updated the syft contracts, moving them to a single file to better isolate autogen'd code and make updating easier. Key change was that the distro contract has changed the Version/Name fields, I switched component-detection to use the VersionID/Id fields to preserve current behaviour. See e.g  
https://github.com/anchore/syft/blob/706f2916791acd8b7355126c9b2b166d0dfcf543/syft/linux/identify_release_test.go#L25 for the new fields.

Todo:
- [x] Mirror new syft image


src/Microsoft.ComponentDetection.Detectors/linux/LinuxScanner.cs
```diff
@@ -17,7 +17,7 @@ namespace Microsoft.ComponentDetection.Detectors.Linux
     [Export(typeof(ILinuxScanner))]
     public class LinuxScanner : ILinuxScanner
     {
-        private const string ScannerImage = "governancecontainerregistry.azurecr.io/syft:0.24.1@sha256:3cc99325855732073ae403a3fa010c3d2fadf27bf7b0a58d4eeaf8454a034ce0";
+        private const string ScannerImage = "docker.io/anchore/syft:v0.47.0";
```
tofay, line 20: Please can you mirror this into the governance ACR?
Then I can update this.

JamieMagee (approved):

PR #126, opened 2022-06-13
Preserve Linux source package name

tofay:
I've updated the LinuxComponent to include the source package name where possible.

Why?
Many Linux distributions (debian/alpine/mariner) publish CVE data against source package names only, so this is required for users to CVE check against the output of component-detection.

Note: Syft doesn't always provide the source package name (it doesn't appear to provide this for ubuntu packages when the source and binary package names are the same).

This builds on https://github.com/microsoft/component-detection/pull/88, but adds a new field to LinuxComponent instead of of making the Name of LinuxComponent be the source package name iff the distro publishes CVEs against source package name.


cobya:
@JamieMagee you probably have the most context to review this

tofay:
@JamieMagee before I rebase this, is this a change that you'd accept?

My team is currently working around this by discovering source packages where binary name != source name and registering them ourselves, but ideally we wouldn't have to.

FernandoRojo:
Closing due to staleness/large number of conflicts, have opened an issue for this here: https://github.com/microsoft/component-detection/issues/1286 and will revisit this code when priorities allow.

PR #128, opened 2022-06-15
log info when the go cli is not present

jcfiorenzano:


src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -115,6 +115,7 @@ private async Task<bool> UseGoCliToScan(string location, ISingleFileComponentRec
 
             if (!isGoAvailable)
             {
+                Logger.LogInfo("Go CLI was not found in the system");
```
grvillic, line 118: Could we also do another small change? To stop printing `Unexpected relationship output from go mod graph:` when relationship is empty. Go mod graph always has an empty line at the end, and causes confusion for users.

https://github.com/microsoft/component-detection/pull/128/files#diff-2eb0ceacc6881719053c971a532e41eef7da6857453203a5ae2c394cec89d151L235

grvillic (approved):

PR #165, opened 2022-07-01
DockerFile Detector Improvements - swallow DockerReference Detector error, turn off DockerReferece Detector by default, Update DockerReference Component Contract

RushabhBhansali:


src/Microsoft.ComponentDetection.Detectors/dockerfile/DockerfileComponentDetector.cs
```diff
@@ -69,25 +69,33 @@ private Task ParseDockerFile(string fileContents, string fileLocation, ISingleFi
...
+                            break;
+                        default:
+                            break;
+                    }
                 }
-            }
 
-            return baseImage;
+                return baseImage;
+            } catch (Exception e)
```
jcfiorenzano, line 93: Add a test for your changes

jcfiorenzano (approved):

src/Microsoft.ComponentDetection.Detectors/dockerfile/DockerfileComponentDetector.cs
RushabhBhansali: @FernandoRojo  is going to fast follow with unit tests

PR #179, opened 2022-07-26
fix: make poetry an experimental detector

JamieMagee:
#167 enabled the poetry detector by default. We don't want to globally enable it yet, but instead move from an opt-in to an experimental detector.


grvillic (approved):

annaowens (approved):

JamieMagee:
Verification failures are expected as we disabled the poetry detector.

PR #180, opened 2022-07-27
Detect python components from Linux containers

tofay:
Azure for Operators have many small teams producing Linux container images and sharing them with other teams. We want the consuming teams to be registering the components from those container images with Component Governance, in order for the consuming team to meet various OSS requirements.

To that end, we need component-detection to detect more component types in Linux containers than just OS packages.

This PR adds support for python package detection in containers as that is a) used by our teams b) supported by syft and component-detection. I expect we'll want to add more non-Linux types in future (e.g java, ruby gems).

Are you ok with this approach?
Should I put this behind a feature flag initially that we can enable in Azure for Operators? I'll update the docs/readme after we've decided these.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -10,7 +10,7 @@
     <ItemGroup>
         <PackageVersion Include="CommandLineParser" Version="2.9.1"/>
         <PackageVersion Include="coverlet.msbuild" Version="3.1.2"/>
-        <PackageVersion Include="Docker.DotNet" Version="3.125.5"/>
+        <PackageVersion Include="Docker.DotNet" Version="3.125.10"/>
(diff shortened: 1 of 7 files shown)


tofay:
@RushabhBhansali is this change something component-detection maintainers would accept? i.e is there a desire to expand container scanning function of component-detection.

tevoinea:
Another thing to consider with respect to detecting non 'linux' components in containers is the (likely) difference in detection coverage between syft and component-detection. It could be confusing for users if there's a discrepancy between what can be detected inside a container vs outside. Ideally, component-detection would run _inside_ the container to detect all but the 'linux' components.

JamieMagee:
Closing this as we don't intend to merge this.

PR #185, opened 2022-08-04
refactor: use `var` in place of explicit types

JamieMagee:
Automatic refactor applied by IDE to use `var` in place of explicit types where the type is otherwise defined. For example:

```C#
var sb = new StringBuilder();
```

instead of

```C#
StringBuilder sb = new StringBuilder
```

This matches the `.editorconfig` enforced style guidelines.

### DIFF
--- a/src/Microsoft.ComponentDetection.Common/AsyncExecution.cs
+++ b/src/Microsoft.ComponentDetection.Common/AsyncExecution.cs
@@ -13,9 +13,9 @@ namespace Microsoft.ComponentDetection.Common
                 throw new ArgumentNullException(nameof(toExecute));
             }
 
-            Task<T> work = Task.Run(toExecute);
+            var work = Task.Run(toExecute);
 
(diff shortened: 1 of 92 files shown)


RushabhBhansali:
Have we considered effect on readability ?  
especially since its open source and our audience is large.... while browsing code on github its hard to figure out type of variable with `var`  since we don't have luxury of hovering over it on visual studio code. 

I think at the end of the day, it comes to the preference. I would appreciate if we can expand on any feedback / datapoint / rationale that triggered this refactor ?

RushabhBhansali:
> **Have we considered effect on readability ?** especially since its open source and our audience is large.... while browsing code on github its hard to figure out type of variable with `var` since we don't have luxury of hovering over it on visual studio code.
> 
> I think at the end of the day, it comes to the preference. I would appreciate if we can expand on any feedback / datapoint / rationale that triggered this refactor ?

Discussed offline, to determine that this is the path we want to go forward. Approving the PR

RushabhBhansali (approved):

PR #194, opened 2022-08-10
docs: how to update syft

JamieMagee:
This was an action item out of our last community meeting: https://github.com/microsoft/component-detection/discussions/134

### DIFF
--- a/docs/update-syft.md
+++ b/docs/update-syft.md
@@ -1,0 +1,51 @@
+﻿# Updating Syft
+
+For container detection we use [Syft][1].
+Occasionally, as new versions of Syft are released, we need to update the version we use.
+To do this:
+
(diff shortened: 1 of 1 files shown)


docs/update-syft.md
```diff
@@ -0,0 +1,51 @@
+﻿# Updating Syft
+
+For container detection we use [Syft][1].
```
cobya, line 3: Probably worth leaving a note here that this can only be done by maintainers, not by all users.

cobya (approved):

docs/update-syft.md
JamieMagee: IMO it should be obvious that not everyone can push container images to our container registry.

Issue #202, opened 2022-08-16
Fix analyzer warnings


JamieMagee:
This is an overarching issue to track analyzer warnings that are currently silenced. To fix a single warning:

1. Remove the warning silence from [`.editorconfig`](https://github.com/microsoft/component-detection/blob/main/.editorconfig)
2. Run a build locally
3. Fix any warnings that are emitted
4. Create a pull request
5. Link it to this issue
6. Once merged check the relevant box below

<details>
<summary>Table of Contents</summary>

- `StyleCop.Analyzers`
- `Microsoft.VisualStudio.Threading.Analyzers`
- `Roslyn`
- Introduced with .NET 6 migration
  - Language rules
  - Code quality rules
  - `StyleCop.Analyzers`

</details>

## [`StyleCop.Analyzers`](https://github.com/DotNetAnalyzers/StyleCopAnalyzers)

- [x] [SA1001](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1001.md)
  - Commas should not be preceded by whitespace
  - #204 
- [x] [SA1005](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1005.md)
  - Single line comment should begin with a space
  - #205
- [x] [SA1101](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1101.md)
  - Prefix local calls with this
  - #236 
- [x] [SA1105](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1105.md)
  - Single line comment should begin with a space
  - #206 
- [x] [SA1008](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1008.md)
  - Opening parenthesis should not be followed by a space.
  - #212 
- [x] [SA1009](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1009.md)
  - Closing parenthesis should not be followed by a space
  - #210 
- [x] [SA1011](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1011.md)
  - Closing square bracket should be followed by a space
  - #211 
- [x] [SA1012](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1012.md)
  - Opening brace should be preceded by a space
  - #213 
[...]

PR #206, opened 2022-08-16
refactor: fix SA1105

JamieMagee:
Related to #202

Single line comment should begin with a space
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1105.md


cobya (commented):
Why was this disabled previously if there are no occurrences of it in the codebase?

JamieMagee:
> Why was this disabled previously if there are no occurrences of it in the codebase?

I believe it was fixed alongside one of the other fixes.

amitla1 (approved):

PR #216, opened 2022-08-23
Make Vcpkg Experimental detector

grvillic:
## Context
As we do Vcpkg rollout, we want to make sure we have good metrics about its performance and unexpected failures across the board by having it running as an experimental detector


JamieMagee (approved):

iskiselev:
Is there any information how it can be used and what is supported?

Issue #196, opened 2022-08-10
Scanning of files that aren't really Dockerfiles is causing errors


mthalman:
After the upgrade to version 0.197.26 of `AzureArtifacts.manifest-generator-task.manifest-generator-task.ManifestGeneratorTask`, component detection is causing it to fail when attempting to scan files that are named like a Dockerfile but aren't really Dockerfiles.

An example of such a file is https://github.com/dotnet/dotnet-docker/blob/main/eng/dockerfile-templates/Dockerfile.linux.copy-files. It's a template file, not an actual Dockerfile that gets built. But component detection is finding this file and attempting to parse it, leading to this error:

```
[INFO] Finding components... 
[INFO] Starting enumeration of /mnt/vss/_work/1 
[INFO] Enumerated 2129 files and 1025 directories in 00:00:00.0722318 
[INFO] No instructions received to scan docker images. 
[INFO] Discovered dockerfile: /mnt/vss/_work/1/s/dotnet-dotnet-docker/eng/dockerfile-templates/Dockerfile.linux.copy-files 
[INFO] No compatible Conda installation found on system. Conda detection will not run. 
[ERROR] There was an unexpected error:  
[ERROR] SafelyExecute logged ParseException: Parsing failure: unexpected '{'; expected A or C or E or F or H or L or M or O or R or S or U or V or W (Line 1, Column 1); recently consumed: 
##[error]Encountered an error while generating the manifest.
##[error]Error details: Component detector failed. Result: Error.
##[error]Manifest generation failed, please check output file for more info.
```

aasim:
This is using CD version 1.2.0

JamieMagee:
Closed in #217

PR #217, opened 2022-08-23
Fix for Bug 196

amitla1:
Related: #196 

Added a try-catch statement in the OnFileFound method for the Docker component because component detection was attempting to scan files that are named like a Dockerfile but aren't really a Dockerfile which was causing an error.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/dockerfile/DockerfileComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/dockerfile/DockerfileComponentDetector.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Composition;
 using System.IO;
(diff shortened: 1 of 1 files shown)


src/Microsoft.ComponentDetection.Detectors/dockerfile/DockerfileComponentDetector.cs
```diff
@@ -36,19 +36,26 @@ public class DockerfileComponentDetector : FileComponentDetector, IDefaultOffCom
...
+                }
 
-            string contents;
-            using (var reader = new StreamReader(file.Stream))
+                var stageNameMap = new Dictionary<string, string>();
+                var dockerFileComponent = this.ParseDockerFile(contents, file.Location, singleFileComponentRecorder, stageNameMap);
+            } catch (Exception e)
             {
-                contents = await reader.ReadToEndAsync();
+                this.Logger.LogError($"The file is not a Dockerfile it just has 'Docker' in the name. \n Error Message: <{e.Message}>");
```
JamieMagee, line 56: I would downgrade this to a warning. Error is usually used for unrecoverable issues. We are able to recover from this by skipping the file.

src/Microsoft.ComponentDetection.Detectors/dockerfile/DockerfileComponentDetector.cs
```diff
@@ -36,19 +36,26 @@ public class DockerfileComponentDetector : FileComponentDetector, IDefaultOffCom
...
+                this.Logger.LogException(e, isError: true, printException: true);
```
JamieMagee, line 57: I'm not sure we need the full exception printed here in quiet mode

```suggestion
                this.Logger.LogException(e);
```

JamieMagee (changes requested):
Could you add a little more context on what the changes you made were and why they were made to the PR description. I know this is a small change, but it's a good habit to get into and can help us track down issues later.

src/Microsoft.ComponentDetection.Detectors/dockerfile/DockerfileComponentDetector.cs
jcfiorenzano: What do you think of logging "The file doesn't appear to be a Dockerfile." instead?

JamieMagee: Yeah. Like this?

```suggestion
                this.Logger.LogError($"The file doesn't appear to be a Dockerfile: '{file.Location}'");
```

JamieMagee (approved):
Approved pending fix for indentation warning

> Elements should have the same indentation

PR #220, opened 2022-08-24
feat: add container image

### LINKED ISSUE #38: Publish container image
To make Component Detection easier to consume and use, we should start publishing container images.

Two open questions are:
1. What should we use as a base image?
2. Where should we publish our image?

## 1. What should we use as a base image?

For our base image, we can use a one of dotnet's runtime-deps images[^1] once #37 is merged. These images don't have any dotnet runtime installed, which makes them a lot smaller than the standard runtime images.

These are currently the .NET Core 3.1 tags available:

<details>
<summary>Tags</summary>

3.1
3.1-alpine
3.1-alpine-arm64v8
3.1-alpine3.12
3.1-alpine3.12-arm64v8
3.1-alpine3.13
3.1-alpine3.13-arm64v8
3.1-alpine3.14
3.1-alpine3.14-arm64v8
3.1-bionic
3.1-bionic-arm32v7
3.1-bionic-arm64v8
3.1-bullseye-slim
3.1-bullseye-slim-arm32v7
3.1-bullseye-slim-arm64v8
3.1-buster-slim
3.1-buster-slim-arm32v7
3.1-buster-slim-arm64v8
3.1-cbl-mariner1.0
3.1-focal
3.1-focal-arm32v7
3.1-focal-arm64v8

</details>

For simplicity, we may want to use `3.1-bullseye-slim`, but `3.1-cbl-mariner1.0` and `3.1-alpine3.14` may also be considered

## 2. Where should we publish our image?

We should push to the Microsoft Container Registry[^2], but we might also want to push to the GitHub Container Registry as well.

[^1]: https://github.com/dotnet/dotnet-docker/tree/main/src/runtime-deps/3.1
[^2]: https://github.com/microsoft/containerregistry

JamieMagee:
Closes #38


Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    ./src/Microsoft.ComponentDetection
+
+FROM mcr.microsoft.com/dotnet/runtime-deps:3.1-cbl-mariner1.0 AS runtime
+WORKDIR /app
+COPY --from=build /app/out ./
+
+RUN tdnf install -y \
+    golang \
+    moby-engine \
+    python
```
sbs2001, line 22: In addition to these, for all detectors to work we'll need to have `gradle`, `maven`, `pnpm` and `poetry` to be installed.

amitla1 (approved):

PR #223, opened 2022-08-24
refactor: fix SA1500

JamieMagee:
Related to #202

Braces for multi-line statements should not share line https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1500.md

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/linux/LinuxContainerDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/linux/LinuxContainerDetector.cs
@@ -215,7 +215,10 @@ namespace Microsoft.ComponentDetection.Detectors.Linux
                 record.BaseImageLayerMessage = $"Base image annotations not found on image {image}, Results will not be mapped to base image layers";
                 this.Logger.LogInfo(record.BaseImageLayerMessage);
                 return 0;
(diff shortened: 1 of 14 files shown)


src/Microsoft.ComponentDetection.Detectors/linux/LinuxContainerDetector.cs
```diff
@@ -215,7 +215,10 @@ private async Task<int> GetBaseImageLayerCount(ContainerDetails scannedImageDeta
                 record.BaseImageLayerMessage = $"Base image annotations not found on image {image}, Results will not be mapped to base image layers";
                 this.Logger.LogInfo(record.BaseImageLayerMessage);
                 return 0;
-            } else if (scannedImageDetails.BaseImageRef == "scratch") {
+            }
+
+            if (scannedImageDetails.BaseImageRef == "scratch")
```
amitla1, line 220: was it intentional for the "else if" to turn into just an "if" statement

JamieMagee: Yes, it was a small optimization recommended by my IDE[^1]. I can revert if you want.

[^1]: https://www.jetbrains.com/help/resharper/RedundantIfElseBlock.html

amitla1 (approved):

Issue #221, opened 2022-08-24
Make Component Detection entrypoints asynchronous


JamieMagee:
Currently our standalone executable[^1] and our library[^2] entrypoints are implemented synchronously. However, further down the stack we call an asynchronous method with `GetAwaiter().GetResult()`

https://github.com/microsoft/component-detection/blob/8e8c3a4557380c22bd337b7c7f0b650ea27216e8/src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs#L154

This pattern is called 'sync over async' and can easily cause deadlocks. The problem is described well in Stephen Toub's article "Should I expose synchronous wrappers for asynchronous methods?" from 2012[^3]

We should convert both entrypoints to true asynchronous calls all the way down the stack. Asynchronous `Main` methods have been a feature of C# since 7.1[^4]

[^1]: https://github.com/microsoft/component-detection/blob/bf900cef6e7e5d81be3e4506f9b575435d2dd414/src/Microsoft.ComponentDetection/Program.cs#L10
[^2]: https://github.com/microsoft/component-detection/blob/8e8c3a4557380c22bd337b7c7f0b650ea27216e8/src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs#L30
[^3]: https://devblogs.microsoft.com/pfxteam/should-i-expose-synchronous-wrappers-for-asynchronous-methods/
[^4]: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-7.1/async-main

PR #228, opened 2022-08-26
refactor: use asynchronous main method

JamieMagee:
Partial fix for #221

### DIFF
--- a/src/Microsoft.ComponentDetection/Program.cs
+++ b/src/Microsoft.ComponentDetection/Program.cs
@@ -1,13 +1,14 @@
-﻿using System;
-using System.Diagnostics;
-using System.Linq;
-using Microsoft.ComponentDetection.Contracts;
-
-namespace Microsoft.ComponentDetection.Loader
+﻿namespace Microsoft.ComponentDetection
 {
+    using System;
+    using System.Diagnostics;
+    using System.Linq;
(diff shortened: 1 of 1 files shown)


src/Microsoft.ComponentDetection/Program.cs
```diff
@@ -1,13 +1,14 @@
...
+    using System;
+    using System.Diagnostics;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using Microsoft.ComponentDetection.Contracts;
+
     public class Program
     {
-        public static void Main(string[] args)
+        public static async Task Main(string[] args)
```
cobya, line 11: Are you aware of how the SBOM tool and other partners consume the package? Do we need this to be a major version breaking change?

JamieMagee: Yes, all other consumers use the Orchestrator. This entrypoint is for the standalone `component-detection` CLI. Modification of the Orchestrator entrypoint will require a major version bump:

https://github.com/microsoft/component-detection/blob/7c5d7fed6ef1c372331adf612b6bbb674e6aef9d/src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs#L30

cobya (approved):

PR #263, opened 2022-09-08
Update detector execution telemetry to set id in timeout cases

FernandoRojo:


JamieMagee:
@FernandoRojo what's the context for this change? And is it still relevant?

PR #288, opened 2022-10-06
Migration component detection to .net6

RushabhBhansali:
Created based on #76 

Changes from #76 
- upgraded CD build to .net6

Additional work (only in this PR):
- Fixed compilation errors and suppressed the analyzer errors 

### Testing
- successful dotnet build.
- successful unit tests.

PR #296, opened 2022-10-11
Disable warnings as errors setting

grvillic:
## Problem
When working directly in this repository the Warnings are not breaking the builds as the setting would expect. When other repositories consume this repo as a submodule, this setting breaks them big time for submodule code.

## Solution
Short term: Disable treat warnings as errors setting
Long term: Understand why it isn't breaking builds in this repo


grvillic:
Closing this PR, as this was fixed indirectly by this change that converted warnings as suggestions in: https://github.com/microsoft/component-detection/pull/288/files#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5b

PR #300, opened 2022-10-17
fix: Add MacOS support

### LINKED ISSUE #267: Component Detection timing out on a repo with symlinks on a Mac environment
**Some initial context**
We have a javascript monorepo with around 1000 packages. For doing the usual dependency installation (equivalent to the normal usage of `yarn install` or `npm install`), we use a modified version of `yarn install`. It gets rid of the flat structure of the node_modules directory and instead creates symlinks inside each package. The closest open source contemporary would be [pnpm](https://pnpm.io/). You can refer to https://pnpm.io/symlinked-node-modules-structure for more details on the directory structure post installation. 

When we moved to this method of installation, component detection started timing out on Mac machines. Going back to the usual installation (flat node_modules) fixed the issue, so this means that the issue shows up with symlinks. Requesting for a quick triage as the current state that we are in is not a viable long-term solution. 

Please @ me if you need further details or if there is anything I can do to increase the clarity of the issue.

asmundg:
On MacOS file systems with heavily nested symlinks, component detection would stall out. This happened because symlinks to the same location were never detected as such, since the realpath logic only ran on Linux.

We don't have any dependencies on glibc, so calling standard unix libc functions works on both Linux and BSD. This change adds symlink resolution on MacOS.

Fixes #267.

### DIFF
--- a/src/Microsoft.ComponentDetection.Common/PathUtilityService.cs
+++ b/src/Microsoft.ComponentDetection.Common/PathUtilityService.cs
@@ -71,6 +71,7 @@ namespace Microsoft.ComponentDetection.Common
 
         private static readonly bool IsWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
         private static readonly bool IsLinux = RuntimeInformation.IsOSPlatform(OSPlatform.Linux);
+        private static readonly bool IsMacOS = RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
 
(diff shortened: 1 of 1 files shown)


JamieMagee:
Thanks for the PR. Were you able to test this locally?

asmundg:
Yes, tested and appears to work fine.

sweggersen:
Hi. Can we get this merged? Thanks, Sam

JamieMagee (approved):

PR #301, opened 2022-10-17
Update tomlyn to 0.15.1 fixing PoetryDetector

dsteeley:
We have hit issues with the PoetryDetector. 
```
[ERROR] SafelyExecuteAsync logged TomlException: (991,2) : error : Unable to set the property extras on object type Microsoft.ComponentDetection.Detectors.Poetry.Contracts.PoetryLock.
```
I believe this is due to an issue with Tomlyn where defaulting missing tables didn't work for `object` types. This has now been fixed in version [0.15.1](https://github.com/xoofx/Tomlyn/releases/tag/0.15.1).
[Corresponding PR](https://github.com/xoofx/Tomlyn/pull/44)

Could I please request a release of component-detection once this change is in, to fixup my pipelines? 

Question for reviewer:
- Why did the testing in `component-detection` not identify this issue and could you suggest some additional testing to add to verify that PRs haven't broken detectors.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -39,7 +39,7 @@
         <PackageVersion Include="System.Reactive" Version="5.0.0"/>
         <PackageVersion Include="System.Runtime.Loader" Version="4.3.0"/>
         <PackageVersion Include="System.Threading.Tasks.Dataflow" Version="4.11.1"/>
-        <PackageVersion Include="Tomlyn" Version="0.15.0"/>
+        <PackageVersion Include="Tomlyn" Version="0.15.1"/>
         <PackageVersion Include="yamldotnet" Version="11.2.1"/>
         <PackageVersion Include="Faker.net" Version="2.0.154"/>
         <PackageVersion Include="Valleysoft.DockerfileModel" Version="1.0.0"/>


JamieMagee:
Thanks for the PR.

Do you have a sample `poetry.lock` we can use in a unit test to make sure we don't regress this behaviour in future?

dsteeley:
The following test has a few example `poetry.lock` files that I would have expected to catch this. 
https://github.com/microsoft/component-detection/blob/main/test/Microsoft.ComponentDetection.Detectors.Tests/PoetryComponentDetectorTests.cs#L88

I can provide a full example file if there is another location that tests are run?

JamieMagee:
In the example you shared we're missing [`metadata` object](https://github.com/microsoft/component-detection/blob/main/src/Microsoft.ComponentDetection.Detectors/poetry/Contracts/PoetryLock.cs)? Have I got that right?

dsteeley:
Yes the metadata object isn't tested and example would be the toml table below
```
[metadata]
lock-version = "1.1"
python-versions = "^3.8"
content-hash = "af93ed08cae8fee3abe9fefd73c2471b3c195dadc4ce1f316ad68bf68ed170dc"
```
I think that the tests weren't being run somehow as the [minimal example](https://github.com/microsoft/component-detection/blob/main/test/Microsoft.ComponentDetection.Detectors.Tests/PoetryComponentDetectorTests.cs#L29) would have hit the issue of being unable to generate the `extras` Dictionary from default values.

JamieMagee:
Okay, I'll merge this and cut a release to unblock you. I created #304 to followup on why this was not caught by our unit tests.

PR #303, opened 2022-10-17
refactor: fix IDE0009, IDE0033, IDE0150, IDE0051, SA1028, IDE0005,  IDE0041

RushabhBhansali:
Warnings fixed
- IDE0051: Private member is unused
-  SA1028: Code should not contain trailing whitespace
- IDE0009: Add 'this' or 'Me' qualification
- IDE0033: Prefer explicitly provided tuple element name
- IDE0150: Null check can be clarified
- IDE0051: Private member is unused
- IDE0005: Using directive is unnecessary
- IDE0041: Null check can be simplified

### DIFF
--- a/test/Microsoft.ComponentDetection.Orchestrator.Tests/Services/DetectorProcessingServiceTests.cs
+++ b/test/Microsoft.ComponentDetection.Orchestrator.Tests/Services/DetectorProcessingServiceTests.cs
@@ -450,7 +450,7 @@ namespace Microsoft.ComponentDetection.Orchestrator.Tests.Services
             });
 
             records.Should().Contain(x => x is DetectorExecutionTelemetryRecord);
-            records.Count(x => x is DetectorExecutionTelemetryRecord)
+            records.Count(x => x is not null)
                 .Should().Be(2);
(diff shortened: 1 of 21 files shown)


test/Microsoft.ComponentDetection.Orchestrator.Tests/Services/DetectorProcessingServiceTests.cs
```diff
@@ -450,7 +450,7 @@ public void ProcessDetectorsAsync_CapturesTelemetry()
             });
 
             records.Should().Contain(x => x is DetectorExecutionTelemetryRecord);
-            records.Count(x => x is DetectorExecutionTelemetryRecord)
```
JamieMagee, line 453: I just want to double check these are equivalent? I am not super familiar with pattern matching syntax.

JamieMagee (approved):

test/Microsoft.ComponentDetection.Orchestrator.Tests/Services/DetectorProcessingServiceTests.cs
RushabhBhansali: the idea is that `records` is a collection of Type `DetectorExecutionTelemetryRecord` so, it can't be anything other than  `null` or that type.

PR #305, opened 2022-10-18
fix IDE0031, IDE0042, IDE0018, IDE0017, IDE0004, IDE0007, IDE0016

RushabhBhansali:
fixed following analyzer warning
- fix IDE0031: Null check can be simplified
- IDE0042: Variable declaration can be deconstructed
- IDE0018: Variable declaration can be inlined
- IDE0017: Object initialization can be simplified
- IDE0004: Cast is redundant.
- IDE0007: use 'var' instead of explicit type
- IDE0016: Null check can be simplified

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/npm/NpmComponentDetectorWithRoots.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/npm/NpmComponentDetectorWithRoots.cs
@@ -296,7 +296,7 @@ namespace Microsoft.ComponentDetection.Detectors.Npm
         {
             if (dependencies != null)
             {
-                foreach (JProperty dependency in dependencies.Cast<JProperty>())
+                foreach (var dependency in dependencies.Cast<JProperty>())
                 {
                     if (dependency != null)
                     {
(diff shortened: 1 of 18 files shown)


src/Microsoft.ComponentDetection.Detectors/npm/NpmComponentDetectorWithRoots.cs
```diff
@@ -296,7 +296,7 @@ private void EnqueueDependencies(Queue<(JProperty, TypedComponent)> queue, JToke
         {
             if (dependencies != null)
             {
-                foreach (JProperty dependency in dependencies)
+                foreach (var dependency in dependencies.Cast<JProperty>())
```
JamieMagee, line 299: This seems less readable to me. Do you know which warning specifically this is addressing?

test/Microsoft.ComponentDetection.Orchestrator.Tests/Services/BcdeScanExecutionServiceTests.cs
```diff
@@ -492,18 +492,18 @@ public void VerifyTranslation_DevDependenciesAreMergedWhenSameComponentInDiffere
             var componentDWithDevDepTrueCopy = new DetectedComponent(new NpmComponent("testD", "1.0.0"), detector: npmDetector);
 
             // The hint for reading this test is to know that each "column" you see visually is what's being merged, so componentAWithNoDevDep is being merged "down" into componentAWithDevDepTrue.
-#pragma warning disable format
+#pragma warning disable IDE0007 // Use implicit type
```
JamieMagee, line 495: Why is this required?

JamieMagee (approved):

PR #322, opened 2022-10-21
added verification for dockerfile and spdx detector, minor bug fixes

RushabhBhansali:
- Enabled verification tests for `DockerReference` and `SPDX22SBOM` detectors
- minor bug fix in `DockerRefernce` detection that detected Repository kind as canonical.
- bug fix in de-serialization of "spdx" component during verification tests. 

Note: verification tests are expected to fail because new detectors are enabled in this PR.

### DIFF
--- a/src/Microsoft.ComponentDetection.Contracts/BcdeModels/TypedComponentConverter.cs
+++ b/src/Microsoft.ComponentDetection.Contracts/BcdeModels/TypedComponentConverter.cs
@@ -25,6 +25,7 @@ namespace Microsoft.ComponentDetection.Contracts.BcdeModels
             { ComponentType.Conda, typeof(CondaComponent) },
             { ComponentType.DockerReference, typeof(DockerReferenceComponent) },
             { ComponentType.Vcpkg, typeof(VcpkgComponent) },
+            { ComponentType.Spdx, typeof(SpdxComponent) },
         };
 
         public override bool CanWrite
--- a/src/Microsoft.ComponentDetection.Contracts/DockerReference.cs
+++ b/src/Microsoft.ComponentDetection.Contracts/DockerReference.cs
@@ -32,7 +32,7 @@ namespace Microsoft.ComponentDetection.Contracts
             }
             else if (string.IsNullOrEmpty(tag))
             {
-                if (string.IsNullOrEmpty(digest))
+                if (!string.IsNullOrEmpty(digest))
                 {
                     return new CanonicalReference
                     {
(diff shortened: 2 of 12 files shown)


src/Microsoft.ComponentDetection.Contracts/DockerReference.cs
```diff
@@ -32,7 +32,7 @@ public static DockerReference CreateDockerReference(string repository, string do
             }
             else if (string.IsNullOrEmpty(tag))
             {
-                if (string.IsNullOrEmpty(digest))
+                if (!string.IsNullOrEmpty(digest))
```
RushabhBhansali, line 35: if digest is not null, we should indentify as canonical.

src/Microsoft.ComponentDetection.Contracts/BcdeModels/TypedComponentConverter.cs
```diff
@@ -25,6 +25,7 @@ public class TypedComponentConverter : JsonConverter
             { ComponentType.Conda, typeof(CondaComponent) },
             { ComponentType.DockerReference, typeof(DockerReferenceComponent) },
             { ComponentType.Vcpkg, typeof(VcpkgComponent) },
+            { ComponentType.Spdx, typeof(SpdxComponent) },
```
RushabhBhansali, line 28: this was causing deserialization error while running the verification test

JamieMagee:
Two points:

It looks like the serialization of SPDX components is wrong. They're all serialized as

> Expected foundComponent to be true because The component for SPDX22SBOM--- was not present in the old manifest file. Verify this is expected behavior before proceeding, but found False.

I'd expect a name and version in the output.

Secondly, I'd rather some more targeted `Dockerfile`s to test with i.e. testing each different reference type, testing `AS` statements, testing variable resolutions.

```Dockerfile
FROM ubuntu AS base

FROM base
```

or

```Dockerfile
ARG tag
FROM ubuntu:${tag} as base

FROM base
```

RushabhBhansali:
> Two points:

It looks like the serialization of SPDX components is wrong. They're all serialized as

> Expected foundComponent to be true because The component for SPDX22SBOM--- was not present in the old manifest file. Verify this is expected behavior before proceeding, but found False.

I'd expect a name and version in the output.


Fixed it by addiing setter in `SpdxComponent` class.

RushabhBhansali:
> Secondly, I'd rather some more targeted Dockerfiles to test with i.e. testing each different reference type, testing AS statements, testing variable resolutions.

I have covered Canonical = 0, Tagged = 2, Dual = 3, Digest = 4, in the existing example. Only `Respository = 1` is not reproducible because if I don't provide domain in the dockerfile, code deafults it to `docker.io/` and its identified as canonical.

JamieMagee (approved):

PR #333, opened 2022-10-27
Add param to skip searching /Plugins for extra detectors

cobya:
On most runs of component detection, the /Plugins folder is searched for additional detector assemblies, but the folder does not exist, showing a [WARN] message like
```
[WARN] Provided search path D:\a\_tasks\SomeFolder\1.1.48\detection\Plugins does not exist.
```

This adds a default parameter to allow skipping the check for additional detectors in /Plugins without modifying logic of additional directory search parameters.


cobya:
This parameter could most likely be set as a default of true, but keeping backwards compatibility in mind it will default to false.

PR #335, opened 2022-10-27
fix: CA1308

daniel-akili:
Related to #202

### DIFF
--- a/src/Microsoft.ComponentDetection.Contracts/TypedComponent/PipComponent.cs
+++ b/src/Microsoft.ComponentDetection.Contracts/TypedComponent/PipComponent.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics.CodeAnalysis;
 using PackageUrl;
 
 namespace Microsoft.ComponentDetection.Contracts.TypedComponent
@@ -21,9 +22,8 @@ namespace Microsoft.ComponentDetection.Contracts.TypedComponent
 
         public override ComponentType Type => ComponentType.Pip;
 
-#pragma warning disable CA1308
+        [SuppressMessage("Usage", "CA1308:Normalize String to Uppercase", Justification = "Casing cannot be overwritten.")]
         public override string Id => $"{this.Name} {this.Version} - {this.Type}".ToLowerInvariant();
-#pragma warning restore CA1308
 
         public override PackageURL PackageUrl => new PackageURL("pypi", null, this.Name, this.Version, null, null);
     }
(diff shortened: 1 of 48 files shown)


RushabhBhansali (commented):
I only see one file changed in the PR. Did your changes got reverted in later commits ?

src/Microsoft.ComponentDetection.Detectors/rust/Contracts/CargoPackage.cs
```diff
@@ -39,7 +39,9 @@ public override int GetHashCode()
         {
             return HashCode.Combine(
                 EqualityComparer<string>.Default.GetHashCode(this.name),
+#pragma warning disable CA1308
```
RushabhBhansali, line 42: this suppression can be moved to method level.

src/Microsoft.ComponentDetection.Detectors/pip/PythonVersion.cs
```diff
@@ -174,8 +174,10 @@ private static int ComparePreRelease(PythonVersion a, PythonVersion b)
                 return -1;
             }
 
+#pragma warning disable CA1308
```
RushabhBhansali, line 177: this suppression can be moved to method level.

src/Microsoft.ComponentDetection.Contracts/TypedComponent/PipComponent.cs
```diff
@@ -21,7 +21,9 @@ public PipComponent(string name, string version)
 
         public override ComponentType Type => ComponentType.Pip;
 
+#pragma warning disable CA1308
```
RushabhBhansali, line 24: use `SuppressMessage` attribute instead.

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -75,7 +75,9 @@ protected override async Task OnFileFound(ProcessRequest processRequest, IDictio
                 }
                 else
                 {
+#pragma warning disable CA1308
```
RushabhBhansali, line 78: no need to suppress here,  instead  we can use .toUpperInvariant` and change `.mod` -> `.MOD` and `.sum` -> `.SUM` in the switch statement below.

RushabhBhansali (commented):
Provided

src/Microsoft.ComponentDetection.Detectors/cocoapods/PodComponentDetector.cs
```diff
@@ -404,7 +404,9 @@ public string GetSpecRepositoryOfSpec(string specName)
                     {
                         // CocoaPods specs are stored in a git repo but depending on settings/CocoaPods version
                         // the repo is shown differently in the Podfile.lock
+#pragma warning disable CA1308
                         return repository.Key.ToLowerInvariant() switch
```
JamieMagee, line 408: I think you could use `ToUpperInvariant()` here as well, and change the switch case from `trunk` to `TRUNK`.

RushabhBhansali (approved):

JamieMagee (approved):

PR #336, opened 2022-10-28
Address CA1831 after path substring changes

cobya:
After #330 CA1831 was being flagged in PathUtilityService (on some build machines) due to the new use of [..1] indexers. This PR adds the .AsSpan call to avoid making data copies.

`Microsoft.ComponentDetection.Common\PathUtilityService.cs(85,77): error CA1831: Use 'AsSpan' instead of the 'System.Range'-based indexer on 'string' to avoid creating unnecessary data copies`


RushabhBhansali (approved):

RushabhBhansali:
its interesting that build did not break in #330

JamieMagee (approved):

PR #337, opened 2022-10-28
Fix Analyzer warnings: CA1707, CA1300, CA1040

RushabhBhansali:
Related to #202  
fixed 
- CA1707: Identifiers should not contain underscores
- SA1300: Element Must Begin With Upper CaseLetter
- CA1040: Avoid empty interfaces

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/poetry/Contracts/PoetrySource.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/poetry/Contracts/PoetrySource.cs
@@ -14,6 +14,7 @@ namespace Microsoft.ComponentDetection.Detectors.Poetry.Contracts
         public string reference { get; set; }
 
         [SuppressMessage("StyleCop.CSharp.NamingRules", "SA1300:ElementMustBeginWithUpperCaseLetter", Justification = "Deserialization contract. Casing cannot be overwritten.")]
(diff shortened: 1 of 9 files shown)


src/Microsoft.ComponentDetection.Detectors/poetry/Contracts/PoetrySource.cs
```diff
@@ -14,6 +14,7 @@ public class PoetrySource
         public string reference { get; set; }
 
         [SuppressMessage("StyleCop.CSharp.NamingRules", "SA1300:ElementMustBeginWithUpperCaseLetter", Justification = "Deserialization contract. Casing cannot be overwritten.")]
+        [SuppressMessage("Naming", "CA1707:IdentifiersShouldNotContainUnderscores", Justification = "Deserialization contract. Casing cannot be overwritten.")]
         public string resolved_reference { get; set; }
```
JamieMagee, line 18: Use `DataMember` attributes[^1] denote the serialized name, and fix the name in our codebase.

[^1]: https://learn.microsoft.com/en-us/dotnet/api/system.runtime.serialization.datamemberattribute?view=net-6.0

RushabhBhansali: fixed. Also fixed suppression for SA1300 at several other places using the same.

JamieMagee (approved):

PR #338, opened 2022-10-28
fix: CA1829, CA1834, CA1846, CA1847

daniel-akili:
Related to #202

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/spdx/Spdx22ComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/spdx/Spdx22ComponentDetector.cs
@@ -91,7 +91,7 @@ namespace Microsoft.ComponentDetection.Detectors.Spdx
             var name = document["name"]?.ToString();
             var spdxVersion = document["spdxVersion"]?.ToString();
 
-            if (rootElements != null && rootElements?.Length > 1)
+            if (rootElements != null && rootElements.Length > 1)
             {
                 this.Logger.LogWarning($"SPDX file at {processRequest.ComponentStream.Location} has more than one element in documentDescribes, first will be selected as root element.");
             }
--- a/src/Microsoft.ComponentDetection.Detectors/spdx/Spdx22ComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/spdx/Spdx22ComponentDetector.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Composition;
 using System.IO;
@@ -91,7 +91,7 @@ namespace Microsoft.ComponentDetection.Detectors.Spdx
             var name = document["name"]?.ToString();
             var spdxVersion = document["spdxVersion"]?.ToString();
 
-            if (rootElements.Length > 1)
+            if (rootElements != null && rootElements.Length > 1)
             {
                 this.Logger.LogWarning($"SPDX file at {processRequest.ComponentStream.Location} has more than one element in documentDescribes, first will be selected as root element.");
             }
--- a/src/Microsoft.ComponentDetection.Detectors/spdx/Spdx22ComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/spdx/Spdx22ComponentDetector.cs
@@ -91,7 +91,7 @@ namespace Microsoft.ComponentDetection.Detectors.Spdx
             var name = document["name"]?.ToString();
             var spdxVersion = document["spdxVersion"]?.ToString();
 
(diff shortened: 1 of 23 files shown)


src/Microsoft.ComponentDetection.Detectors/spdx/Spdx22ComponentDetector.cs
```diff
@@ -91,7 +91,10 @@ private SpdxComponent ConvertJObjectToSbomComponent(ProcessRequest processReques
             var name = document["name"]?.ToString();
             var spdxVersion = document["spdxVersion"]?.ToString();
 
-            if (rootElements != null && rootElements.Count() > 1)
+            if (rootElements == null || rootElements.Length <= 1)
+            {
+            }
+            else
```
JamieMagee, line 97: The empty `if` looks a little odd. What about `if (rootElements?.Length > 1)`?

test/Microsoft.ComponentDetection.Detectors.Tests/VcpkgComponentDetectorTests.cs
```diff
@@ -29,7 +29,7 @@ public void TestInitialize()
         [TestMethod]
         public async Task TestNlohmann()
         {
-            var spdxFile = @"{
+            var spdxFile = /*lang=json,strict*/ @"{
```
JamieMagee, line 32: What's this?

test/Microsoft.ComponentDetection.Detectors.Tests/VcpkgComponentDetectorTests.cs
```diff
@@ -76,7 +76,7 @@ public async Task TestNlohmann()
         [TestMethod]
         public async Task TestTinyxmlAndResource()
         {
-            var spdxFile = @"{
+            var spdxFile = /*lang=json,strict*/ @"{
```
JamieMagee, line 79: Same here

JamieMagee (changes requested):
Couple of small comments for CA1829. Other changes look good 👍

test/Microsoft.ComponentDetection.Detectors.Tests/VcpkgComponentDetectorTests.cs
daniel-akili: Will look for more documentation about the `JSON002` error that raised the issue

JamieMagee: We spoke about this offline and decided to mark `JSON002` as a suggestion until we can find some documentation about it.

JamieMagee, line 32: re-added?

JamieMagee (changes requested):
Couple of comments on `JSON002` and also snapshot diffs.

test/Microsoft.ComponentDetection.Detectors.Tests/VcpkgComponentDetectorTests.cs
daniel-akili: was re-added from previous commit, but should be good now

src/Microsoft.ComponentDetection.Detectors/spdx/Spdx22ComponentDetector.cs
```diff
@@ -91,7 +91,7 @@ private SpdxComponent ConvertJObjectToSbomComponent(ProcessRequest processReques
             var name = document["name"]?.ToString();
             var spdxVersion = document["spdxVersion"]?.ToString();
 
-            if (rootElements.Length > 1)
+            if (rootElements != null && rootElements.Length > 1)
```
JamieMagee, line 94: I'm not sure if this is valid syntax, but you might be able to combine this into a single expression.

```suggestion
            if (rootElements?.Length > 1)
```

src/Microsoft.ComponentDetection.Detectors/spdx/Spdx22ComponentDetector.cs
```diff
@@ -91,7 +91,7 @@ private SpdxComponent ConvertJObjectToSbomComponent(ProcessRequest processReques
             var name = document["name"]?.ToString();
             var spdxVersion = document["spdxVersion"]?.ToString();
 
-            if (rootElements != null && rootElements.Count() > 1)
+            if (rootElements?.Length > 1)
```
daniel-akili, line 94: changing to this syntax builds successfully but does not pass the snapshot checks. Revert it?

JamieMagee (approved):

PR #341, opened 2022-10-29
fix: analyzer warnings SA1210, SA1414, VSTHRD002, CA1849, VSTHRD103, VSTHRD104

RushabhBhansali:
Related to #202 
- fix SA1210: Using directives should be ordered alphabetically

- fix SA1414: Tuple types in signatures should have element names

- fix VSTHRD002: synchronously waiting on Task

- fix CA1849: Call async methods when in an async method

- fix VSTHRD103: Call async methods when in an async method


src/Microsoft.ComponentDetection.Common/Telemetry/TelemetryRelay.cs
```diff
@@ -60,10 +62,10 @@ public void Shutdown()
                 try
                 {
                     // Set a timeout for services that flush synchronously.
-                    AsyncExecution.ExecuteVoidWithTimeoutAsync(
+                    await AsyncExecution.ExecuteVoidWithTimeoutAsync(
                         () => service.Flush(),
```
JamieMagee, line 66: Is there a `flushAsync()` available? And maybe the [`Task.Wait(TimeSpan)`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.wait?view=net-6.0#system-threading-tasks-task-wait(system-timespan)) that was added in .NET 6 is better?

JamieMagee: We might need to create a followup issue to address `AsyncExecution.ExecuteVoidWithTimeoutAsync` calls.

src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs
```diff
@@ -202,8 +202,7 @@ private async Task GenerateEnvironmentSpecificTelemetryAsync(BcdeExecutionTeleme
 
                         return string.Join(Environment.NewLine, aptListResult.Split(Environment.NewLine).Where(x => x.Contains("libssl")));
                     });
-
-                    if (!getLibSslPackages.Wait(taskTimeout))
+                    if (await Task.WhenAny(getLibSslPackages, Task.Delay(taskTimeout)) != getLibSslPackages)
```
JamieMagee, line 205: [`Task.WaitAsync(TimeSpan)`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.waitasync?view=net-7.0)?

JamieMagee (changes requested):
Couple of comments on the async fixes

src/Microsoft.ComponentDetection.Common/Telemetry/TelemetryRelay.cs
RushabhBhansali: no `flushAsync()` is available here. 

the `AsyncExecution.ExecuteVoidWithTimeoutAsync()` has custom exceptions, may be not a good idea to change that logic here by using `Task.Wait(timespan)`.  We can probably keep the api of `ExecuteVoidWithtimeoutAsync()` same and use `Task.Wait(timespan)` inside of it (as part of the follow up issue) ?

src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs
RushabhBhansali: done.

src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs
```diff
@@ -202,10 +202,10 @@ private async Task GenerateEnvironmentSpecificTelemetryAsync(BcdeExecutionTeleme
 
                         return string.Join(Environment.NewLine, aptListResult.Split(Environment.NewLine).Where(x => x.Contains("libssl")));
                     });
-
-                    if (!getLibSslPackages.Wait(taskTimeout))
+                    await getLibSslPackages.WaitAsync(taskTimeout);
+                    if (getLibSslPackages.IsFaulted)
                     {
-                        throw new TimeoutException($"The execution did not complete in the alotted time ({taskTimeout} seconds) and has been terminated prior to completion");
+                        throw new TimeoutException($"The execution did not complete in the alloted time ({taskTimeout} seconds) and has been terminated prior to completion");
```
JamieMagee, line 208:
```suggestion
                        throw new TimeoutException($"The execution did not complete in the allotted time ({taskTimeout} seconds) and has been terminated prior to completion");
```

src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs
```diff
@@ -202,10 +202,10 @@ private async Task GenerateEnvironmentSpecificTelemetryAsync(BcdeExecutionTeleme
 
                         return string.Join(Environment.NewLine, aptListResult.Split(Environment.NewLine).Where(x => x.Contains("libssl")));
                     });
-
-                    if (!getLibSslPackages.Wait(taskTimeout))
+                    await getLibSslPackages.WaitAsync(taskTimeout);
+                    if (getLibSslPackages.IsFaulted)
```
JamieMagee, line 206: I am not sure if this is correct. I think `WaitAsync` will throw `TaskCanceledException`.

JamieMagee (approved):

src/Microsoft.ComponentDetection.Common/Telemetry/TelemetryRelay.cs
RushabhBhansali: done

src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs
RushabhBhansali: removed this  custom exception.

src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs
RushabhBhansali: REmovved the duplicate exception.

PR #365, opened 2022-11-15
Adding TargetFrameworks to the list of data retreived from NuGet packages

JamieMagee:
Successor to #79 

Original PR description:
> Per a conversation about adding target framework information to the component detection data, this is a PR that shows what that could look like.
>
>Things get slightly tricky here since packages declare target frameworks through the dependency declarations in the nuspec, as well as the folder structure of the package. This hasn't been through much testing beyond the unit tests, but they've been augmented to test some potential patterns.
>
>I mainly want to get some feedback on this before I spend much more time on testing. I can potentially test this against all NuGet.org packages pretty easily, which also gives us a test oracle since they calculate target frameworks as well.
>
>Things that might need consideration:
>* Addition of NuGet.Packaging as a dependency
>* how the data is plumbed through the NuGetComponent (I followed the Authors example as an array of >additional strings)
>* Removal of async from ProcessFile (nothing in there was async when I finished)
>* Shifting more responsibility to NuGetNuspecUtilities even through it's not strictly nuspec-related >anymore.
>* Using the nuspec reader rather than bespoke XML parsing
>
>I'm open to any and all feedback here.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/nuget/NuGetNuspecUtilities.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/nuget/NuGetNuspecUtilities.cs
@@ -1,8 +1,10 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
-using System.Threading.Tasks;
+using Microsoft.ComponentDetection.Contracts;
+using NuGet.Packaging;
 
 namespace Microsoft.ComponentDetection.Detectors.NuGet
 {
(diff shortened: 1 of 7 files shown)


src/Microsoft.ComponentDetection.Detectors/nuget/NuGetNuspecUtilities.cs
```diff
@@ -45,20 +91,68 @@ public static async Task<byte[]> GetNuspecBytesAsync(Stream nupkgStream)
             finally
             {
                 // make sure that no matter what we put the stream back to the beginning
-                nupkgStream.Seek(0, SeekOrigin.Begin);
+                nupkgComponentStream.Stream.Seek(0, SeekOrigin.Begin);
+            }
+        }
+
+        public static (string Name, string Version, string[] Authors, HashSet<string> TargetFrameworks) GetNuspecDataFromNuspecStream(Stream sourceStream)
```
marklio, line 98: There is a question of whether we should assume a directory containing a .nuspec file is an expanded .nupkg. There may be adjacent directories and subdirectories that could contain relevant target framework data that perhaps should be analyzed as in GetNuGetPackageDataFromNupkg above.

marklio:
Thanks for rebasing/bumping this, @JamieMagee ! When I first wrote this, this was very much intended to be a draft to start a conversation on how target framework info could be added to component detection. I'm very happy to engage in whatever capacity is most helpful in seeing progress here.

SokoFromNZ:
Hi guys
Any progress on this in the near future? I actually would love to read more properties from the .nupkg/.nuspec...
- copyright
- urls
- licencse
- authors
- ...basically all the main props in a nuspec

cobya:
@JamieMagee is this still needed?

SokoFromNZ:
It definitely is needed by me :)

FernandoRojo:
Closing due to staleness/large number of conflicts. There is active work ongoing to improve detection for NuGet in this area that will cover the issues raised by this PR.

PR #412, opened 2023-01-25
refactor: migrate away from MEF to Dependency Injection

### LINKED ISSUE #400: Migrate away from Managed Extensibility Framework (MEF)
Currently, we use Managed Extensibility Framework (MEF) to handle injection of dependencies through Component Detection. MEF was the best way to do Dependency Injection (DI) in .NET Framework, and still is the best way to do application extensibility via plugins. However, in Modern .NET, the most common way to do DI is via the `Microsoft.Extensions.DependencyInjection` package. I suggest migrating away from MEF and to `Microsoft.Extensions.DependencyInjection`-style DI for a few main reasons:

1. We don't need the full functionality of MEF
2. `Microsoft.Extensions.DependencyInjection` has better support, documentation, and developer mindshare
3. `Microsoft.Extensions.DependencyInjection` has less 'magic'
	1. No [loading random DLLs](https://github.com/microsoft/component-detection/blob/fb6b4a31e2a8793a421d2b51e06ba243cc75827b/src/Microsoft.ComponentDetection.Orchestrator/Services/DetectorRegistryService.cs#L96)
	2. No decorating classes with `[Export]` and properties with `[Import]`

The pattern I suggest migrating to is outline in the [Options pattern guidance for .NET library authors](https://learn.microsoft.com/en-us/dotnet/core/extensions/options-library-authors#post-configuration) documentation. I suggest performing this migration in a few key steps

## 1. Add dependencies and scaffolding #370 

We need to add a reference to `Microsoft.Extensions.DependencyInjection.Abstractions` in `Microsoft.ComponentDetection.Orchestrator` and `Microsoft.Extensions.DependencyInjection` to `Microsoft.ComponentDetection`. The `Microsoft.Extensions.DependencyInjection.Abstractions` package contains interfaces and abstract classes required for Dependency Injection which can be implemented by any library. We need to use this package specifically in our library (`Microsoft.ComponentDetection.Orchestrator`) because we don't know which DI library our caller will be using. In our CLI tool, `Microsoft.ComponentDetection`, we'll use the Microsoft implementation (`Microsoft.Extensions.DependencyInjection`) of the abstractions provided in  `Microsoft.Extensions.DependencyInjection.Abstractions`

We also need to add an extension method to `IServiceCollection` called `AddComponentDetection`. This is where we'll eventually configure all the required services. This extension should initially be private to our assembly, so we can rapidly iterate on the design, without the changes being semver breaking changes.

```C#
internal static class ServiceCollectionExtensions
{
    /// <summary>
    /// Register Component Detection services.
    /// </summary>
    /// <param name="services">The <see cref="IServiceCollection" /> to register the services with.</param>
    /// <returns>The <see cref="IServiceCollection" /> so that additional calls can be chained.</returns>
    public static IServiceCollection AddComponentDetection(this IServiceCollection services) => services;
}
```

This is already complete in #370 

## 2. Add class constructors

`Microsoft.Extensions.DependencyInjection` works on constructor injection, whereas MEF works on property injection. As we currently use MEF, most of our classes don't have the required constructors (only the default parameterless constructors created by Roslyn).

For example, the [`DockerService`](https://github.com/microsoft/component-detection/blob/fb6b4a31e2a8793a421d2b51e06ba243cc75827b/src/Microsoft.ComponentDetection.Common/DockerService.cs) class would need to have the following constructor added

```C#
[Export(typeof(IDockerService))]
public class DockerService : IDockerService
{
    public DockerService(ILogger logger)
    {
        this.Logger = logger;
    }

    [Import]
    public ILogger Logger { get; set; }
}
```

This needs to be done for all classes that are currently annotated with the `[Export]` attribute.

## 3. Replace `DetectorTestUtility`

`DetectorTestUtility` provides a fluent API to easily set up detector unit tests. Unfortunately, it uses property injection instead of constructor injection to set up the classes.

We need to refactor `DetectorTestUtility` to use constructor injection, while still providing the mocking capabilities and fluent API if possible.

## 4. Wire up classes

In our `ServiceCollectionExtensions` class, we need to wire up all the classes that are currently annotated with the `[Export]` attribute. For example:

```C#
public static IServiceCollection AddComponentDetection(this IServiceCollection services)
{
    services.AddSingleton<IComponentDetector, DockerfileComponentDetector>();
    services.AddSingleton<IComponentDetector, NpmComponentDetector>();
    // etc.

    return services;
}
```

## 4. Migrate `Microsoft.ComponentDetection`

At this point we can migrate `Microsoft.ComponentDetection` to use `ServiceCollection` and `AddComponentDetection`. For example:

```C#
var serviceCollection = new ServiceCollection()
    .AddLogging(configure =>
        // ...
    )
    .AddComponentDetection();
```

See [Dependency injection in .NET](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection) for more guidance.

## 5. Make `ServiceCollectionExtensions` public

At this point `Microsoft.ComponentDetection` should be using `Microsoft.Extensions.DependencyInjection` to handle dependency injection. So, if not done already, we should expose `ServiceCollectionExtensions` (and `AddComponentDetection` as part of our public API, so consumers of Component Detection (like https://github.com/microsoft/sbom-tool) can use it too.

## 6. Remove MEF attributes

We should remove any MEF attributes, like `[Export]` and `[Import]`.

## 7. Remove public properties

At this point we need to make a breaking API change by removing the public properties that MEF used to configure dependencies. For example, [`DockerService`](https://github.com/microsoft/component-detection/blob/fb6b4a31e2a8793a421d2b51e06ba243cc75827b/src/Microsoft.ComponentDetection.Common/DockerService.cs) would change from:

```C#
[Export(typeof(IDockerService))]
public class DockerService : IDockerService
{
    public DockerService(ILogger logger)
    {
        this.Logger = logger;
    }

    [Import]
    public ILogger Logger { get; set; }
}
```

to

```C#
[Export(typeof(IDockerService))]
public class DockerService : IDockerService
{
    private readonly ILogger logger;

    public DockerService(ILogger logger)
    {
        this.logger = logger;
    }
}
```

## 8. Remove MEF dependencies

Finally, we need to remove the nuget packages that provide the MEF functionality. These are:

- `System.Composition.AttributedModel`
- `System.Composition.Convention`
- `System.Composition.Hosting`
- `System.Composition.Runtime`
- `System.Composition.TypedParts`

JamieMagee:
This is a quite a large PR, that breaks down all the steps required to close #400. I strongly recommend reviewing this commit-by-commit and reading the commit message. 

<details>
<summary>1. refactor: add constructors to injected classes</summary>

Currently, Managed Extensibility Framework (MEF) does property injection. In migrating to .NET's own Dependency Injection framework `Microsoft.Extensions.DependencyInjection` we need to migrate to constructor injection. This change adds 2 constructors for every class that has properties injected into it:

- A parameterless constructor (for MEF)
- A constructor with parameters for all the required dependencies (for .NET DI)

This is an intermediate step that allows MEF to work, while work continues on the migration to MEF.

</details>

<details>
<summary>2. test: add DetectorTestUtilityBuilder</summary>

This is the equivalent of `DetectorTestUtilityCreator` and `DetectorTestUtility`, but for .NET DI.

</details>

<details>
<summary>3. test: migrate component detector tests to use `DetectorTestUtilityBuilder`</summary>

This change migrates all of the component detector tests to us `DetectorTestUtilityBuilder`, based on .NET DI, instead of `DetectorTestUtility`, based on MEF.

This also adds a new base class for all component detector tests called `BaseDetectorTest` which constructs the `DetectorTestUtilityBuilder` in its constructor.

</details>

<details>
<summary>4. test: delete DetectorTestUtility and DetectorTestUtilityCreator</summary>

No longer required after all component detector tests have been migrated to use `DetectorTestUtilityBuilder` instead

</details>

<details>
<summary>5. test: use dependency injection constructor for service tests</summary>

This change uses the constructor that has parameters for all required properties of a service, instead of using object initializers (which actually use the parameterless constructor)

</details>

<details>
<summary>6. feat: add all services to AddComponentDetection</summary>

This change adds all of the services that will be managed by .NET's dependency injection framework to the `AddComponentDetection` method. This allows you easily configure all the services that Component Detection requires

```csharp
var serviceProvider = new ServiceCollection()
  .AddComponentDetection()
  .BuildServiceProvider();
var orchestrator = serviceProvider.GetRequiredService<Orchestrator>();
var result = await orchestrator.LoadAsync(args);
```

</details>

<details>
<summary>7. refactor: use AddComponentDetection to setup services</summary>

This is one of the final changes in migrating away from Managed Extensibility Framework (MEF) and to .NET's Dependency Injection (DI).

These changes include:
- Deleting classes that are no longer required for handling property injection:
  - `IDetectorDependencies` and `DetectorDependencies` which held references to all common services
  - `InjectionParameters` which had `IDetectorDependencies` injected, and contained static references to all common services
  - `IDetectorRegistryService` and `DetectorRegistryService` which loaded additional classes from DLLs
    - This functionality designed to be useful before Component Detection was open sourced. Now that it is, detectors can be added more easily. This functionality was not used internally.
    - This means that `IComponentGovernanceOwnedDetectors` can also be deleted
- Updating injected types
  - `BcdeScanExecutionService` and `DetectorListingCommandService` had `IDetectorRegistryService` injected. Instead, they now have `IEnumerable<IComponentDetector>` injected directly.
- `Init` method for `FileWritingService` and `Logger` moved to their respective interfaces, `IFileWritingService` and `ILogger`
  - `Orchestrator` had the concrete types `FileWritingService` and `Logger` injected. However, it's best practice to inject interfaces instead. In making this change, some methods that were available in the concrete type had to be added to the interface
  - `VerbosityMode` needed to be moved from `Common` to `Contracts` to facilitate this

</details>

<details>
<summary>8. refactor: init TelemetryRelay during orchestrator setup</summary>

As `TelemetryRelay` is a static class, it cannot be instantiated using Dependency Injection. Instead, we inject an `IServiceProvider` into `Orchestrator` and call `Init` on `TelemetryRelay` with the required `IEnumerable<ITelemetryService>`

</details>

<details>
<summary>9. refactor: only one GraphTranslationService</summary>

The parameter `IEnumerable<Lazy<IGraphTranslationService, GraphTranslationServiceMetadata>>` was designed to allow loading of alternative graph translation services, dynamically, from DLLs. That functionality never materialized, and MEF is being removed, so I am simplifying that constructor parameter to be simply `IGraphTranslationService`.

If alternate graph translation services are required in the future, the constructor can be updated to take a `IDictionary<IGraphTranslationService, GraphTranslationServiceMetadata>>`. This can be added to the `IServiceCollection` as follows:

```csharp
services.AddSingleton<DefaultGraphTranslationService>();
services.AddSingleton<FancyNewGraphTranslationService>();

services.AddSingleton<IDictionary<IGraphTranslationService, GraphTranslationServiceMetadata>>(sp =>
  new Dictionary<IGraphTranslationService, GraphTranslationServiceMetadata>>
  {
    { sp.GetRequiredService<DefaultGraphTranslationService>(), new GraphTranslationServiceMetadata { Priority = 1 } },
    { sp.GetRequiredService<FancyNewGraphTranslationService>(), new GraphTranslationServiceMetadata { Priority = 0 } },
  });
```

</details>

<details>
<summary>10. refactor: remove System.Composition usage</summary>

This includes:
- Removal of `[Export]` and `[Import]` attributes
- Conversion of public properties to private readonly fields
- Removal of parameterless constructors

</details>

<details>
<summary>11. refactor: migrate YarnLockFileFactory and YarnLockFileParser to be managed by dependency injection</summary>

</details>

<details>
<summary>12. test: final test fixes</summary>

</details>

<details>
<summary>13. refactor: remove System.CompositionModel dependencies</summary>

🎉 

</details>

<details>
<summary>14. refactor: fix misc analyzer warnings</summary>

</details>


src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs
```diff
@@ -25,41 +25,38 @@ public class Orchestrator
...
-        var parserResult = this.ArgumentHelper.ParseArguments<BaseArguments>(args, true);
+        var parserResult = this.argumentHelper.ParseArguments<BaseArguments>(args, true);
         parserResult.WithParsed(x => baseArguments = x);
         if (parserResult.Tag == ParserResultType.NotParsed)
         {
             // Blank args for this part of the loader, all things are optional and default to false / empty / null
             baseArguments = new BaseArguments();
         }
 
+        TelemetryRelay.Instance.Init(this.serviceProvider.GetRequiredService<IEnumerable<ITelemetryService>>());
```
JamieMagee, line 59: Leave a comment here about why we need to call this.

cobya (approved):
Reviewed offline and changes look good.

Issue #320, opened 2022-10-21
Use upstream `SemanticVersioning` package


JamieMagee:
We currently use a fork[^1] of the `SemanticVersioning` NuGet package[^2]. This was due to a couple of issues we opened upstream[^3][^4]. As these are now resolved, we should investigate if we can migrate to the upstream version, instead of maintaining our own fork.

At the same time, we also use the `NuGet.Versioning` package[^5], which provides some of the same functionality. We should see if we can migrate fully to `NuGet.Versioning` instead.

[^1]: https://github.com/microsoft/component-detection/tree/main/src/Microsoft.ComponentDetection.Detectors/rust/SemVer
[^2]: https://www.nuget.org/packages/SemanticVersioning
[^3]: https://github.com/adamreeve/semver.net/issues/46
[^4]: https://github.com/adamreeve/semver.net/issues/47
[^5]: https://www.nuget.org/packages/NuGet.Versioning

PR #426, opened 2023-02-10
Changed to upstream Semantic Versioning package

Omotola:
This PR updates the semver package used in the rust detector to the upstream version . [Issue 320](https://github.com/microsoft/component-detection/issues/320)


src/Microsoft.ComponentDetection.Detectors/rust/DependencySpecification.cs
```diff
@@ -1,9 +1,9 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using Microsoft.ComponentDetection.Detectors.Rust.Contracts;
-using Semver;
 
-using Range = Microsoft.ComponentDetection.Detectors.Rust.SemVer.Range;
+using Range = SemanticVersioning.Range;
+using Version = SemanticVersioning.Version;
```
JamieMagee, line 6: Is this required? Can the imports be satisfied with `using SemanticVersioning;` instead?

Omotola: I tried that, but there is an ambiguous reference error.
image

JamieMagee (approved):
Needs conflicts resolved, but otherwise LGTM

PR #429, opened 2023-02-14
Add output log of skipped / failed components

cobya:
Add output log of skipped or failed components to the end of the build post detector summary. A log group will be created for each detector which has failed components.

image


src/Microsoft.ComponentDetection.Common/DependencyGraph/ComponentRecorder.cs
```diff
@@ -96,10 +114,17 @@ public sealed class SingleFileComponentRecorder : ISingleFileComponentRecorder
 
         private readonly ConcurrentDictionary<string, DetectedComponent> detectedComponentsInternal = new ConcurrentDictionary<string, DetectedComponent>();
 
+        /// <summary>
+        /// Dictionary of components which had an error during parsing and a dummy data value that only allocates 1 byte.
+        /// </summary>
+        private readonly ConcurrentDictionary<string, byte> skippedComponentsInternal = new ConcurrentDictionary<string, byte>();
```
melotic, line 120: Why not just use a `ConcurrentBag<string>`?

src/Microsoft.ComponentDetection.Common/DependencyGraph/ComponentRecorder.cs
```diff
@@ -173,6 +204,19 @@ public DetectedComponent GetComponent(string componentId)
...
         }
 
+        public void RegisterPackageParseFailure(string skippedComponent)
+        {
+            if (skippedComponent == null)
+            {
+                throw new ArgumentNullException(paramName: nameof(skippedComponent));
+            }
+
+            lock (this.registerSkippedLock)
```
melotic, line 214: This lock adds unnecessary overhead, the concurrent data structures are already thread-safe (and often times lock-free!)

src/Microsoft.ComponentDetection.Common/DependencyGraph/ComponentRecorder.cs
```diff
@@ -62,6 +62,24 @@ public IEnumerable<DetectedComponent> GetDetectedComponents()
...
 
+    public IEnumerable<string> GetSkippedComponents()
+    {
+        IEnumerable<string> skippedComponents;
+        if (this.singleFileRecorders == null)
+        {
+            return Enumerable.Empty<string>();
+        }
+
+        skippedComponents = this.singleFileRecorders
```
melotic, line 73: You can simpify this query: 

```csharp
this.singleFileRecorders.Select(x => x.GetSkippedComponents().Keys).SelectMany(x => x).Distinct().ToImmutableList();
```

src/Microsoft.ComponentDetection.Common/DependencyGraph/ComponentRecorder.cs
cobya: This is so we aren't iterating the entire list when searching for duplicates on list addition - this is fine for repos with a small number of components but can get expensive for large repositories.

src/Microsoft.ComponentDetection.Common/DependencyGraph/ComponentRecorder.cs
JamieMagee: You can also return directly here. No need to split declaration, initialization, and return statement.

JamieMagee:
There are now some places where we log in the detector that we're not able to parse a package, and some places where we don't. We not also provide a summary of skipped or unparseable packages at the end of the run.

Is it necessary to log directly in the detector if we're providing a summary at the end?

src/Microsoft.ComponentDetection.Common/DependencyGraph/ComponentRecorder.cs
melotic: forgot .NET doesn't like concurrent hash sets :(

cobya:
> There are now some places where we log in the detector that we're not able to parse a package, and some places where we don't. We not also provide a summary of skipped or unparseable packages at the end of the run.
> 
> Is it necessary to log directly in the detector if we're providing a summary at the end?

The purpose of having the summary at the end is not to provide detailed exception messages of why it has been skipped - that should be done in the detector. The summary is to quickly show users how many components were skipped so they can investigate further.

src/Microsoft.ComponentDetection.Common/DependencyGraph/ComponentRecorder.cs
cobya: Done.

src/Microsoft.ComponentDetection.Common/DependencyGraph/ComponentRecorder.cs
cobya: Done.

JamieMagee (approved):

Issue #373, opened 2022-11-23
Investigate performance of `PipComponentDetector`


JamieMagee:
We're receiving some support emails about poor performance for extremely large Python projects. We should profile this detector, identify any hotspots, and optimize them. Potentially related to #108

RushabhBhansali:
Related support request to investigate:
https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/2010996

melotic:
Might be closed by #629

PR #435, opened 2023-02-16
perf: add buffering to `FileWritingService`

melotic:
This adds buffering to the `FileWritingService`. Previously, every log message synchronously called `File.AppendAllText`.

This should help increase performance for #373 and across the board.

Profiling Component Detection lead to serious performance gains:

Before:
image

After:
image

### DIFF
--- a/src/Microsoft.ComponentDetection.Common/FileWritingService.cs
+++ b/src/Microsoft.ComponentDetection.Common/FileWritingService.cs
@@ -1,14 +1,16 @@
 namespace Microsoft.ComponentDetection.Common;
 using System;
+using System.Collections.Concurrent;
 using System.IO;
 using Microsoft.ComponentDetection.Common.Exceptions;
 
-public class FileWritingService : IFileWritingService
+public sealed class FileWritingService : IFileWritingService
 {
     public const string TimestampFormatString = "yyyyMMddHHmmssfff";
 
     private readonly object lockObject = new object();
     private readonly string timestamp = DateTime.Now.ToString(TimestampFormatString);
+    private readonly ConcurrentDictionary<string, StreamWriter> bufferedStreams = new();
 
     public string BasePath { get; private set; }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.ComponentDetection.Common/IFileWritingService.cs
```diff
@@ -1,8 +1,10 @@
 ﻿namespace Microsoft.ComponentDetection.Common;
+
+using System;
 using System.IO;
 
 // All file paths are relative and will replace occurrences of {timestamp} with the shared file timestamp.
-public interface IFileWritingService
+public interface IFileWritingService : IDisposable
```
JamieMagee, line 7: What about `IAsyncDisposable`? `StreamWriter` implements `IDisposable` and `IAsyncDisposable`

src/Microsoft.ComponentDetection.Common/FileWritingService.cs
```diff
@@ -26,10 +28,13 @@ public void AppendToFile(string relativeFilePath, string text)
     {
         relativeFilePath = this.ResolveFilePath(relativeFilePath);
 
-        lock (this.lockObject)
+        if (!this.bufferedStreams.TryGetValue(relativeFilePath, out var streamWriter))
```
JamieMagee, line 31: I know there are `async` overloads for writing to streams. I assume it would be a major refactoring to use them here? Can you create an issue for us to follow up with?

src/Microsoft.ComponentDetection/Program.cs
```diff
@@ -32,6 +32,9 @@
 
     Console.WriteLine($"Execution finished, status: {exitCode}.");
 
+    // Manually dispose to flush logs as we force exit
+    await serviceProvider.DisposeAsync();
```
JamieMagee, line 36: `await using` instead of manually calling dispose?

melotic: We cannot use `await using` as dispose will not be called when we call `Environment.Exit`

src/Microsoft.ComponentDetection.Common/FileWritingService.cs
melotic: It would be a huge refactor as every log statement would have to become async. We should just switch to a proper logging library like Serilog or the extension logging library instead, like we discussed offline.

melotic:
Benchmarking with [`hyperfine`](https://github.com/sharkdp/hyperfine) yield an 8% performance increase:

| Command       | Mean [s]      | Min [s] | Max [s] |
|---------------|---------------|---------|---------|
| `origin/main` | 8.604 ± 0.889 | 7.576   | 10.621  |
| buffering     | 7.920 ± 0.157 | 7.710   | 8.259   |

src/Microsoft.ComponentDetection.Common/FileWritingService.cs
melotic: #436

JamieMagee (approved):

PR #457, opened 2023-03-06
feat: add `--PrintManifest` option

### LINKED ISSUE #246: Is there a way to have scan manifest printed to stdout ?
This would make https://github.com/microsoft/component-detection/pull/220 far more usable.

melotic:
This adds the ability to pass `--PrintManifest` to CD and will then output the scan manifest JSON to stdout.

This PR also updates Serilog to only output to standard error, so that the users can simply redirect stdout.

Closes #246

### DIFF
--- a/src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs
+++ b/src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs
@@ -68,7 +68,11 @@ public class Orchestrator
         var reloadableLogger = (ReloadableLogger)Log.Logger;
         reloadableLogger.Reload(configuration =>
             configuration
-                .WriteTo.Console(standardErrorFromLevel: LogEventLevel.Debug)
+                .WriteTo.Console(standardErrorFromLevel: args.Contains(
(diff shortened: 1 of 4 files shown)


src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs
```diff
@@ -68,7 +68,7 @@ public async Task<ScanResult> LoadAsync(string[] args, CancellationToken cancell
         var reloadableLogger = (ReloadableLogger)Log.Logger;
         reloadableLogger.Reload(configuration =>
             configuration
-                .WriteTo.Console()
+                .WriteTo.Console(standardErrorFromLevel: LogEventLevel.Debug)
```
JamieMagee, line 71: This writes everything to `stderr`? I'd be more comfortable if we wrote to `stdout` when `PrintManifest` was `null` or `false`, and only wrote to `stderr` when `PrintManifest` was set to `true`. Some environments consider any output on `stderr` to be a fail condition, and unconditionally writing to `stderr` may be a breaking change.

JamieMagee (approved):

PR #458, opened 2023-03-06
refactor: simplify `PathUtilityService`

### LINKED ISSUE #454: Simplify `PathUtilityService`
[`PathUtilityService`][1] was created when Component Detection was initially ported from .NET Framework to an early version of .NET Core. Many of the file and path utilities we needed were not natively available in `System.IO`. However, in .NET 6 [`System.IO.Path`][2], [`System.IO.File`][3], and other [`System.IO` classes][4] are extremely comprehensive, and we should migrate to those instead.

[1]: https://github.com/microsoft/component-detection/blob/main/src/Microsoft.ComponentDetection.Common/PathUtilityService.cs
[2]: https://learn.microsoft.com/en-us/dotnet/api/system.io.path?view=net-6.0
[3]: https://learn.microsoft.com/en-us/dotnet/api/system.io.file?view=net-6.0
[4]: https://learn.microsoft.com/en-us/dotnet/api/system.io?view=net-7.0

melotic:
Closes #454

### DIFF
--- a/src/Microsoft.ComponentDetection.Common/PathUtilityService.cs
+++ b/src/Microsoft.ComponentDetection.Common/PathUtilityService.cs
@@ -62,5 +62,17 @@ public class PathUtilityService : IPathUtilityService
         return searchPattern.Equals(fileName, StringComparison.OrdinalIgnoreCase);
     }
 
-    public string ResolvePhysicalPath(string path) => Path.GetFullPath(path);
+    public string ResolvePhysicalPath(string path)
+    {
+        var directoryInfo = new DirectoryInfo(path);
+        if (directoryInfo.Exists)
+        {
+            return this.ResolvePathFromInfo(directoryInfo);
+        }
+
+        var fileInfo = new FileInfo(path);
+        return fileInfo.Exists ? this.ResolvePathFromInfo(fileInfo) : null;
+    }
+
+    private string ResolvePathFromInfo(FileSystemInfo info) => info.LinkTarget ?? info.FullName;
 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.ComponentDetection.Common/PathUtilityService.cs
```diff
@@ -117,176 +53,14 @@ public bool MatchesPattern(string searchPattern, string fileName)
...
-        {
-            this.logger.LogWarning("Detected execution in a Windows container. Currently windows containers < 1809 do not support symlinks well, so disabling symlink resolution/dedupe behavior");
-            return true;
-        }
-        else
-        {
-            return false;
-        }
-    }
+    public string ResolvePhysicalPath(string path) => Path.GetFullPath(path);
```
JamieMagee, line 65: Does `GetFullPath` follow symlinks?

JamieMagee (approved):
Approved pending resolving comments.

src/Microsoft.ComponentDetection.Common/PathUtilityService.cs
melotic: Yes: 
image

melotic: It does not. I've updated it to properly create either a `FileInfo` or `DirectoryInfo` class, then we can use the [`LinkTarget`](https://learn.microsoft.com/en-us/dotnet/api/system.io.filesysteminfo.linktarget?view=net-7.0) property.

JamieMagee (approved):

PR #459, opened 2023-03-06
fix: support concurrent runs of CD

### LINKED ISSUE #403: %temp%\GovCompDisc_Log_(stamp).log naming conflicts fail concurrent builds
The SBOM generator is failing in our concurrent builds due to the log file being the same name.
Here is an example of what I'm seeing in our binlog:

image

We have the fix for [Issue#134](https://github.com/microsoft/sbom-tool/issues/134) but we are still seeing concurrent issues when running the SBOM tool.

Is it possible to update the code to have a more unique value in it?  Like, a process id or a guid?
Please let me know if I can provide any more details.

melotic:
This adds the current PID of component detection to the log file name to support concurrent executions of component detection without the two separate instances attempting to use the same file.

We add the PID before the timestamp to that the log files are still sortable.

Fixes #403

### DIFF
--- a/src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs
+++ b/src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs
@@ -63,7 +63,7 @@ public class Orchestrator
 
         var logFile = Path.Combine(
             baseArguments.Output ?? Path.GetTempPath(),
-            $"GovCompDisc_Log{DateTime.Now:yyyyMMddHHmmssfff}.log");
+            $"GovCompDisc_Log_{Environment.ProcessId}_{DateTime.Now:yyyyMMddHHmmssfff}.log");
 
         var reloadableLogger = (ReloadableLogger)Log.Logger;
         reloadableLogger.Reload(configuration =>


src/Microsoft.ComponentDetection.Orchestrator/Orchestrator.cs
```diff
@@ -63,7 +63,7 @@ public async Task<ScanResult> LoadAsync(string[] args, CancellationToken cancell
 
         var logFile = Path.Combine(
             baseArguments.Output ?? Path.GetTempPath(),
-            $"GovCompDisc_Log{DateTime.Now:yyyyMMddHHmmssfff}.log");
+            $"GovCompDisc_Log_{Environment.ProcessId}_{DateTime.Now:yyyyMMddHHmmssfff}.log");
```
JamieMagee, line 66: Don't we need the timestamp first? Otherwise it's `GovCompDisc_Log_12345_20230306...` instead of `GovCompDisc_Log_20230306_12345...`

melotic: Ah yes you're right.

JamieMagee (approved):

PR #460, opened 2023-03-06
perf: fetch from PyPi in parallel and cache project responses

melotic:
This PR makes two performance related changes:

* Fetch JSON response from PyPi in parallel
* Add second cache method to `IPyPiClient`.
  * Use `System.Text.Json` and not `Newtonsoft.Json` as we can deserialize the content stream asynchronously instead of reading the entire JSON response into memory
  * Cache the actual deserialized `PythonProject` object instead of the JSON response and deserializing it every time it's needed

Benchmarking with hyperfine yields some gains:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `justinperez/pip-perf` | 6.221 ± 0.385 | 5.999 | 6.907 | 1.00 |
| `origin/main` | 8.205 ± 2.007 | 6.047 | 10.262 | 1.32 ± 0.33 |

I tried to get a successful benchmarking run without outliers, but each run had at least one outlier.

This should help #373  

---

Profiling also shows that `PythonVersion.ctor` is slow, so perhaps future work can optimize this or make it lazy. Unfortunately, most of the overhead comes from downloading the wheels from PyPI to resolve dependencies of dependencies (and so on). While not trivial, we can also later parallelize those downloads.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/pip/IPyPiClient.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/pip/IPyPiClient.cs
@@ -18,6 +18,8 @@ using Polly;
 
 namespace Microsoft.ComponentDetection.Detectors.Pip;
 
+using System.Text.Json;
+
 public interface IPyPiClient
 {
     Task<IList<PipDependencySpecification>> FetchPackageDependenciesAsync(string name, string version, PythonProjectRelease release);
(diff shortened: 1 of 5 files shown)


melotic:
We should also add caching to the `PythonVersion` class as profiling says this an incredibly expensive operation due to the amount of instances created and the expensive Regex. Testing with a [small script](https://gist.github.com/melotic/7d346b2fa235599ad3af42c1f6d266a4) on the data from our verification tests yields:

```
1317 of 1959 are duplicates
67.23% of the versions can be cached
```

This will probably have a much greater benefit and be cleaner than the current optimizations in this PR.

PR #473, opened 2023-03-10
fix: `Requires-Dist` variants & optional deps

### LINKED ISSUE #241: Python: Dependencies are looked up from incorrect urls
When running component-detection on one of my projects I am seeing output like the following.

```
[INFO] Getting Python data from https://pypi.org/pypi/pywinpty>=1.1.0/json
[WARN] Received 404 Not Found from https://pypi.org/pypi/pywinpty>=1.1.0/json
[WARN] Dependency Package pywinpty>=1.1.0 not found in Pypi. Skipping package
```

It seems like component-detection is in some situations passing an incorrect url which contains the version specifier and obviously does not resolve. This seems to happen for transitive dependencies that are constrained to a specific version by some other direct or transitive dependency.

melotic:
Properly handles `Rquires-Dist` like so:

```
Requires-Dist: numpy<1.27.0,>=1.19.5
```

Also, properly handles dependencies with optional dependencies in pip `METADATA` or in `requirements.txt`. We ignore them for now.

```
Requires-Dist: PyJWT[crypto] (<3,>=1.0.0)
```

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/pip/PipDependencySpecification.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/pip/PipDependencySpecification.cs
@@ -12,7 +12,7 @@ public class PipDependencySpecification
 {
     // Extracts name and version from a Requires-Dist string that is found in a metadata file
     public static readonly Regex RequiresDistRegex = new Regex(
-        @"Requires-Dist:\s*(?:(.*?)\s*\((.*?)\)|([^\s;]*))",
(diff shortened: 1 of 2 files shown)


melotic:
I've tested this on a customer's repo who was reporting failures like `https://pypi.org/pypi/PyJWT[crypto]/json` and `https://pypi.org/pypi/numpy%3C1.27.0,%3E=1.19.5/json`.

This does not fix incorrectly resolving `cv2`, however.

JamieMagee:
> This does not fix incorrectly resolving `cv2`, however.

Can you remove the `Fixes #241` so the issue stays open?

Issue #486, opened 2023-03-23
The NuGetComponentDetector fails with NullReferenceException if the nuspec file does not include a version property


jcfiorenzano:
Example of a nuspec that cause the problem:

```
<package>
    <metadata>
        <id>MyComponent</id>
        <title>MyTittle</title>
        <authors>test</authors>
        <owners>test</owners>
        <description>
            test
        </description>
        <summary>test</summary>
        <licenseUrl>http://test.com</licenseUrl>
        <copyright>Copyright</copyright>
        <requireLicenseAcceptance>false</requireLicenseAcceptance>
    </metadata>
    <files>
        <file exclude="*.nupkg"/>
    </files>
</package>
```

JamieMagee:
We should catch and log this, but not throw an exception.

melotic:
@jcfiorenzano do you have a stack trace?

PR #492, opened 2023-03-30
Fix: adding conditional operator to prevent NullReferenceException in NugetComponentDetector

daniel-akili:
Added conditional operator for name and version variables to prevent NullReferenceException when nuspec file does not contain name/version
#486

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/nuget/NuGetComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/nuget/NuGetComponentDetector.cs
@@ -121,8 +121,15 @@ public class NuGetComponentDetector : FileComponentDetector
 
             var name = metadataNode["id"]?.InnerText;
             var version = metadataNode["version"]?.InnerText;
-
             var authors = metadataNode["authors"]?.InnerText.Split(",").Select(author => author.Trim()).ToArray();
+
+            if (name == null)
+            {
+                this.Logger.LogInformation("Error Parsing 'Name' of NuGet Component", stream.Location);
+                singleFileComponentRecorder.RegisterPackageParseFailure(stream.Location);
+
+                return;
+            }
 
             if (!NuGetVersion.TryParse(version, out var parsedVer))
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.ComponentDetection.Detectors/nuget/NuGetComponentDetector.cs
```diff
@@ -119,8 +119,8 @@ private async Task ProcessFileAsync(ProcessRequest processRequest)
             XmlNode packageNode = doc["package"];
             XmlNode metadataNode = packageNode["metadata"];
 
-            var name = metadataNode["id"].InnerText;
-            var version = metadataNode["version"].InnerText;
+            var name = metadataNode["id"]?.InnerText;
```
melotic, line 122: If the name is null, we shouldn't try to register the component. If `version` is null, the `NuGetVersion.TryParse` call fails and the component is not registered.

Also if `name` is null, `ArgumentNullException` will be thrown when constructing the `NugetComponent`.

https://github.com/microsoft/component-detection/blob/1aba67efcf21c011ac73a5a73c342771e96a375e/src/Microsoft.ComponentDetection.Contracts/TypedComponent/NugetComponent.cs#L13

https://github.com/microsoft/component-detection/blob/1aba67efcf21c011ac73a5a73c342771e96a375e/src/Microsoft.ComponentDetection.Contracts/TypedComponent/TypedComponent.cs#L33-L38

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetComponentDetector.cs
```diff
@@ -119,11 +119,18 @@ private async Task ProcessFileAsync(ProcessRequest processRequest)
...
-            var name = metadataNode["id"].InnerText;
-            var version = metadataNode["version"].InnerText;
-
+            var name = metadataNode["id"]?.InnerText;
+            var version = metadataNode["version"]?.InnerText;
             var authors = metadataNode["authors"]?.InnerText.Split(",").Select(author => author.Trim()).ToArray();
 
+            if (name == null)
+            {
+                this.Logger.LogInformation("Error Parsing 'Name' of NuGet Component", stream.Location);
```
melotic, line 128:
```suggestion
                this.Logger.LogInformation("Could not parse name from Nuspec {NuspecLocation}", stream.Location);
```

melotic (approved):

JamieMagee (approved):

PR #497, opened 2023-04-06
Add support for Nuget lockfiles

Porges:
These have been supported since VS2017 but they are not yet parsed.


src/Microsoft.ComponentDetection.Detectors/nuget/NuGetComponentDetector.cs
```diff
@@ -105,6 +115,12 @@ private async Task ProcessFileAsync(ProcessRequest processRequest)
             else if ("paket.lock".Equals(stream.Pattern, StringComparison.OrdinalIgnoreCase))
             {
                 this.ParsePaketLock(processRequest);
+                return;
```
Porges, line 118: Fixed what seems like a bug here

test/Microsoft.ComponentDetection.Detectors.Tests/Utilities/TestLogger.cs
```diff
@@ -0,0 +1,26 @@
+namespace Microsoft.ComponentDetection.Detectors.Tests.Utilities;
+
+using System;
+using Microsoft.Extensions.Logging;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+internal class TestLogger<T> : ILogger<T>, IDisposable
```
melotic, line 7: Could you help me understand the need for this code? We have some tests that check the log output, and we have a utility that handles that functionality, which can be found at https://github.com/microsoft/component-detection/blob/dbcb973d640bc0529d58c646f94700df3f5147df/test/Microsoft.ComponentDetection.TestsUtilities/ExtensionMethods.cs#L27.

bilby91:
Is there any plans to merge this ?

Issue #476, opened 2023-03-13
Add support for NPM lockfileVersion v3


melotic:
GitHub recently pushed updates to their runners, upgrading their npm version from  `8.19.3` to `9.3.1`[^1]. It seems also that Azure DevOps has updated their runners. With NPM 9, the `package-lock.json` format has changed by default, and it is not backwards compatible with other version of NPM. Running `npm install`, with NPM 9 automatically upgrades the `package-lock.json` to lockfile version 3.

> `"lockfileVersion": 3` The lockfile version used by npm v7, without backwards compatibility affordances. This is used for the hidden lockfile at node_modules/.package-lock.json, and will likely be used in a future version of npm, once support for npm v6 is no longer relevant. [^2]

When CD attempts to process a `package-lock.json` with a v3 lockfile, we throw an exception because of the new `package-lock.json` format:

```
[09:20:18 INF] Could not parse Jtokens from C:\Users\justinperez\Documents\RPE.Partner.Service.Customers.UI\package-lock.json file.
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.ComponentDetection.Detectors.Npm.NpmComponentDetectorWithRoots.ProcessIndividualPackageJTokens(ISingleFileComponentRecorder singleFileComponentRecorder, JToken packageLockJToken, IEnumerable`1 packageJsonComponentStream, Boolean skipValidation) in C:\component-detection\src\Microsoft.ComponentDe
tection.Detectors\npm\NpmComponentDetectorWithRoots.cs:line 153
   at Microsoft.ComponentDetection.Detectors.Npm.NpmComponentDetectorWithRoots.<>c__DisplayClass28_0.<OnFileFoundAsync>b__1(JToken token) in C:\component-detection\src\Microsoft.ComponentDetection.Detectors\npm\NpmComponentDetectorWithRoots.cs:line 120
   at Microsoft.ComponentDetection.Detectors.Npm.NpmComponentDetectorWithRoots.ProcessAllPackageJTokensAsync(IComponentStream componentStream, JTokenProcessingDelegate jtokenProcessor) in C:\component-detection\src\Microsoft.ComponentDetection.Detectors\npm\NpmComponentDetectorWithRoots.cs:line 144
[...]

PR #501, opened 2023-04-10
Accept custom individual component file paths

grvillic:
## Context
Individual custom component file locations can't be added due to removal in DefaultGraphTranslationService. The service should verify they are under root and they are relative.#476 

## Solution
The DefaultGraphTranslationService already  adds the file location where the detector was executed to the components, allow other custom locations to be added under same root (making them relative beforehand)

### DIFF
--- a/src/Microsoft.ComponentDetection.Contracts/IComponentRecorder.cs
+++ b/src/Microsoft.ComponentDetection.Contracts/IComponentRecorder.cs
@@ -1,4 +1,4 @@
-﻿namespace Microsoft.ComponentDetection.Contracts;
+namespace Microsoft.ComponentDetection.Contracts;
 using System.Collections.Generic;
 using Microsoft.ComponentDetection.Contracts.BcdeModels;
 
@@ -45,6 +45,9 @@ public interface ISingleFileComponentRecorder
 
     DetectedComponent GetComponent(string componentId);
 
+    /// <summary>
+    /// Any file added here will be reported as a location on ALL components found in current graph.
+    /// </summary>
     void AddAdditionalRelatedFile(string relatedFilePath);
 
     IReadOnlyDictionary<string, DetectedComponent> GetDetectedComponents();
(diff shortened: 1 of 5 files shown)


src/Microsoft.ComponentDetection.Contracts/IComponentRecorder.cs
```diff
@@ -45,6 +45,9 @@ public interface ISingleFileComponentRecorder
 
     DetectedComponent GetComponent(string componentId);
 
+    /// <summary>
+    /// Any file added here will be reported as a location on ALL components found in current graph.
+    /// </summary>
     void AddAdditionalRelatedFile(string relatedFilePath);
```
grvillic, line 51: In our wrapper we are using this method, but this inadvertently adds it to ALL components found in that file. There currently isn't a way to customize additional locations per component, as we are clearing everything in DefaulGraphTranslationService.cs

test/Microsoft.ComponentDetection.Orchestrator.Tests/Services/DefaultGraphTranslationServiceTests.cs
```diff
@@ -0,0 +1,100 @@
...
+        var expectedNpmComponent = new NpmComponent("npm-component", "1.2.3");
+        var expectedNugetComponent = new NuGetComponent("nugetComponent", "4.5.6");
+        var detectedNpmComponent = new DetectedComponent(expectedNpmComponent);
+        var detectedNugetComponent = new DetectedComponent(expectedNugetComponent);
+
+        // Any Related File will be reported for ALL components found in this graph
+        singleFileComponentRecorder.AddAdditionalRelatedFile(Path.Join(this.sourceDirectory.FullName, relatedFilePath));
+
+        // Registering components in same manifest with different custom paths
+        detectedNpmComponent.AddComponentFilePath(Path.Join(this.sourceDirectory.FullName, npmCustomPath));
```
grvillic, line 72: No detector in this repo calls this `AddComponentFilePath` directly, as the file location gets automatically added by the graph. So it should be a NoOp change for this repo.

melotic (approved):

test/Microsoft.ComponentDetection.Orchestrator.Tests/Services/DefaultGraphTranslationServiceTests.cs
grvillic, line 72: No detector in this repo calls this `AddComponentFilePath` directly, as the file location gets automatically added by the graph. It should be a NoOp change for this repo.

PR #504, opened 2023-04-11
refactor(npm): separate lockfile 3 detector

melotic:
Separates the lock file 3 specific code into a new detector.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/npm/NpmLockfile3Detector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/npm/NpmLockfile3Detector.cs
@@ -1,0 +1,160 @@
+﻿namespace Microsoft.ComponentDetection.Detectors.Npm;
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.ComponentDetection.Contracts;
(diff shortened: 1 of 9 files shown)


src/Microsoft.ComponentDetection.Detectors/npm/NpmLockfile3Detector.cs
```diff
@@ -0,0 +1,160 @@
+﻿namespace Microsoft.ComponentDetection.Detectors.Npm;
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.ComponentDetection.Contracts;
+using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json.Linq;
+
+public class NpmLockfile3Detector : NpmLockfileDetectorBase, IExperimentalDetector
```
JamieMagee, line 10: Do you need to add this to the dependency injection configuration?

melotic: Yep! Missed that

JamieMagee (approved):
The diffs make this hard to review, but I think this is mostly just a copy-paste of existing code.

PR #508, opened 2023-04-13
feat: add detector experiments

melotic:
This adds the ability to conduct _experiments_ on a set of detectors, allowing us to compare the detected components of detectors.

We take diffs of the components with respect to:
* Component IDs
* Root components
* Is a development dependency

The output diff is serialized to a JSON model and written to disk. A future PR will send this data as telemetry.

### Nuget Detector

**Control Group**: `NuGetComponentDetector` and `NuGetProjectModelProjectCentricComponentDetector`
**Experimental Group**: `NuGetPackagesConfigDetector` and `NuGetProjectModelProjectCentricComponentDetector`

### NPM With Roots Detector

**Control Group**: `NpmComponentDetector`
**Experimental Group**: `NpmLockfile3` detector.

### DIFF
--- a/src/Microsoft.ComponentDetection.Orchestrator/Experiments/Models/ExperimentDiff.cs
+++ b/src/Microsoft.ComponentDetection.Orchestrator/Experiments/Models/ExperimentDiff.cs
@@ -1,0 +1,119 @@
+﻿namespace Microsoft.ComponentDetection.Orchestrator.Experiments.Models;
+
+using System.Collections.Generic;
+using System.Linq;
+
+/// <summary>
+/// A model for the difference between two sets of <see cref="ExperimentComponent"/> instances.
+/// </summary>
+public class ExperimentDiff
+{
+    /// <summary>
+    /// Creates a new <see cref="ExperimentDiff"/>.
+    /// </summary>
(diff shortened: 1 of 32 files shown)


src/Microsoft.ComponentDetection.Orchestrator/Experiments/Models/ExperimentDiff.cs
```diff
@@ -0,0 +1,119 @@
...
+public class ExperimentDiff
+{
+    /// <summary>
+    /// Creates a new <see cref="ExperimentDiff"/>.
+    /// </summary>
+    /// <param name="oldComponents">A set of components from the control group.</param>
+    /// <param name="newComponents">A set of components from the experimental group.</param>
+    public ExperimentDiff(
+        IEnumerable<ExperimentComponent> oldComponents,
+        IEnumerable<ExperimentComponent> newComponents)
```
JamieMagee, line 18: nit: use the same terminology all the way through. You switch between old/new and control/experimental.

JamieMagee (approved):
Only tiny comments on naming. Otherwise, looks good. Approving so as not to block.

PR #452, opened 2023-03-01
feat: switch CLI log level arguments

melotic:
Removes the `--Verbosity` command line argument and replaces it with `--LogLevel`.

We can now pass in:

* Verbose
* Debug
* Information
* Warning
* Error
* Fatal

Closes #442

Issue #442, opened 2023-02-21
Change verbosity mode argument names with Serilog's logging levels


melotic:
As discussed in #441 with the switch to Serilog & Microsoft.Extensions.Logging, we should change the name of the `Verbosity` flag in the [command line arguments](https://github.com/microsoft/component-detection/blob/main/src/Microsoft.ComponentDetection.Contracts/VerbosityMode.cs#L3-L8) to match the names of the Serilog [`LogEventLevel`](https://github.com/serilog/serilog/blob/dev/src/Serilog/Events/LogEventLevel.cs)

We should switch the `--Verbosity` argument to the more accurate `--LogLevel`. This parameter will accept the following log levels (From Serilog's `LogEvenLevel` enum):

* Verbose
* Debug
* Information
* Warning
* Error
* Fatal

The equivalents are:

| `--Verbosity` | `--LogLevel` |
|--------|--------|
| `Quiet`| `Error` |
| `Normal`| `Information` |
| `Verbose`| `Debug`|

PR #518, opened 2023-04-21
refactor: migrate to `Spectre.Console.Cli`

### LINKED ISSUE #233: Use `Spectre.Console` to write table output
Currently we use our own custom table formatter for CLI output, [`TabularStringFormat`][1]. A popular library in the .NET ecosystem that provides a better tested alternative is [`Spectre.Console`][2].

In addition to tabular output, it also provides richer output for logs, exceptions, status indicators and more.

[1]: https://github.com/microsoft/component-detection/blob/main/src/Microsoft.ComponentDetection.Common/TabularStringFormat.cs
[2]: https://spectreconsole.net/

JamieMagee:
Closes #233, #442, #452


src/Microsoft.ComponentDetection.Orchestrator/Commands/BaseSettings.cs
```diff
@@ -24,9 +24,9 @@ public abstract class BaseSettings : CommandSettings
     public Guid CorrelationId { get; set; }
 
     [Description("Flag indicating what level of logging to output to console during execution. Options are: Verbose, Normal, or Quiet.")]
-    [DefaultValue(VerbosityMode.Normal)]
-    [CommandOption("--Verbosity")]
-    public VerbosityMode Verbosity { get; set; }
+    [DefaultValue(LogEventLevel.Information)]
+    [CommandOption("--LogLevel")]
+    public LogEventLevel LogLevel { get; set; }
```
melotic, line 29: #452

src/Microsoft.ComponentDetection.Orchestrator/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -144,25 +142,4 @@ public static IServiceCollection AddComponentDetection(this IServiceCollection s
 
         return services;
     }
-
-    public static IServiceCollection ConfigureLoggingProviders(this IServiceCollection services)
```
melotic, line 157: We discussed this previously offline, and this isnt needed.

melotic (approved):

src/Microsoft.ComponentDetection/Program.cs
```diff
@@ -2,52 +2,70 @@
...
-    await serviceProvider.DisposeAsync();
+        config.SetInterceptor(new Interceptor(resolver, resolver.Resolve(typeof(ILogger<Interceptor>)) as ILogger<Interceptor>));
+        config.SetInterceptor(new LogLevelInterceptor());
+        config.SetInterceptor(new PrintManifestInterceptor());
 
-    await Log.CloseAndFlushAsync();
+        config.Settings.ApplicationName = "component-detection";
 
-    // force an exit, not letting any lingering threads not responding.
-    Environment.Exit(exitCode);
```
grvillic, line 47: Do we need to handle threads properly with the new approach? Just double checking we are freeing all resources before we terminate the tool's execution.

melotic: The `ServiceCollection` is disposed inside of the `TypeRegistrar`.

Additionally, each `Command` returns an exit code, which is propagated up by `Spectre.Console.Cli` to the `RunAsync`, which is returned.

grvillic: Also verify the exit code remains the same based on the internal values we are giving it from the orchestrator execution.

melotic: > Also verify the exit code remains the same based on the internal values we are giving it from the orchestrator execution.

This will have to be done in internal, as it has it's own separate orchestrator that will be migrated seperately.

src/Microsoft.ComponentDetection.Orchestrator/Commands/ScanSettings.cs
```diff
@@ -0,0 +1,63 @@
...
+{
+    [CommandOption("--DirectoryExclusionList")]
+    [Description("Filters out specific directories following a minimatch pattern.")]
+    [TypeConverter(typeof(SemicolonDelimitedConverter))]
+    public IEnumerable<string> DirectoryExclusionList { get; set; }
+
+    [CommandOption("--IgnoreDirectories", IsHidden = true)]
+    [Description(
+        "Filters out specific directories, providing individual directory paths separated by semicolon. Obsolete in favor of DirectoryExclusionList")]
+    [TypeConverter(typeof(CommaDelimitedConverter))]
```
melotic, line 22: docs say `paths separated by semicolon` but this uses a `CommaDelimitedConverter`.

PR #519, opened 2023-04-21
fix(experiment): don't process if empty or tainted

melotic:
The logging in `ExperimentService` was broken as it contained two identical `{Count}` parameters.

In addition, if the results of an experiment are completely empty (_i.e._ no components detected in the control group and experiment group), we should skip processing as an empty diff is indicative of a perfect detector.

Lastly, certain detectors running invalidates the experiment results. For instance, in the NPM Lockfile v3 experiment, if the lockfile v2 detector runs, we would be comparing the `NpmComponentDetector` (which would detect _N_ components), to the `NpmLockfile3Detector` (which would detect _**0**_ components). So, the experiment is tainted and we should not process it.

### DIFF
--- a/src/Microsoft.ComponentDetection.Orchestrator/Experiments/ExperimentService.cs
+++ b/src/Microsoft.ComponentDetection.Orchestrator/Experiments/ExperimentService.cs
@@ -35,6 +35,8 @@ public class ExperimentService : IExperimentService
     /// <inheritdoc />
     public void RecordDetectorRun(IComponentDetector detector, IEnumerable<DetectedComponent> components)
     {
+        this.FilterExperiments(detector, components.Count());
+
         foreach (var (config, experimentResults) in this.experiments)
         {
             if (config.IsInControlGroup(detector))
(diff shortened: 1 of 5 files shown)


src/Microsoft.ComponentDetection.Orchestrator/Experiments/ExperimentService.cs
```diff
@@ -59,21 +61,46 @@ public void RecordDetectorRun(IComponentDetector detector, IEnumerable<DetectedC
...
 
+    private void FilterExperiments(IComponentDetector detector, int count) =>
+        this.experiments.RemoveAll(experiment =>
+        {
+            var shouldRemove = !experiment.Config.ShouldRecord(detector, count);
+
+            if (shouldRemove)
+            {
+                this.logger.LogDebug(
+                    "Removing {Experiment} from active experiments because {Id} has {Count} components",
```
JamieMagee, line 72: `has {Count} components` is true right now, but maybe not in general? IMO it's fine for now.

JamieMagee (approved):

src/Microsoft.ComponentDetection.Orchestrator/Experiments/ExperimentService.cs
melotic: Oop. This can be confusing if we filter experiments based on something other than components detected.

PR #529, opened 2023-04-26
fix: make `ExperimentResults` thread safe

melotic:
Detectors execute in parallel, and it's possible to concurrently modify the `ExperimentResults`. I've only run into this once after weeks of testing, so the impact of this is pretty low.

I wrapped the `ExperimentResults` in locks, as I presume it to be more efficient as detectors finish executing in variable amount of time (overhead + number of files found), so the contention on the locks should be quite low. The locks are only accessed from multiple threads once per detector run.


melotic:
Going to refactor this and open a new PR to use `ConcurrentDictioanry` instead.

Issue #231, opened 2022-08-26
Migrate from `Newtonsoft.Json` to `System.Text.Json`


JamieMagee:
For a long time `Newtonsoft.Json` was the best JSON library for .NET. However, since `System.Text.Json` was added in .NET Core 3.0, it's clearly the future. It's a core part of .NET, so there's no need for an additional dependency, and it's far more performant than `Newtonsoft.Json`

A couple of things to be aware of:
- Should we explicitly add [`JsonPropertName` attributes][3] to object properties?
- Should we standardize on [Camel Case][4] (`exampleVariableName`) over [PascalCase][5] (`ExampleVariableName`)?
  - Currently we use a mix of both
  - `Newtonsoft.Json` can accept both by default, can `System.Text.Json` too?
- Migrate to asynchronous methods instead of synchronous?
  - [`SerializeAsync`][6]
  - [`DeserializeAsync`][7]

References:
- [`Newtonsoft.Json` to `System.Text.Json` migration guide][1]
- [What's next for `System.Text.Json`][2]

[1]: https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to?pivots=dotnet-6-0
[2]: https://devblogs.microsoft.com/dotnet/whats-next-for-system-text-json/
[3]: https://docs.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonpropertynameattribute?view=net-6.0
[4]: https://en.wikipedia.org/wiki/Camel_case
[5]: https://en.wiktionary.org/wiki/Pascal_case
[6]: https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.serializeasync?view=netcore-3.1
[7]: https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.deserializeasync?view=netcore-3.1

pn04:
Hi @JamieMagee , I would like to take a stab at it, if this has not been picked up already.
Could you answer some of the queries I have, please



> Should we explicitly add [`JsonPropertyName` attributes](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonpropertynameattribute?view=net-6.0) to object properties?

[...]

PR #537, opened 2023-04-27
refactor: use `System.Text.Json` to serialize manifest

JamieMagee:
For very large component manifests users were seeing `System.OutOfMemoryException` when serializing to JSON.

`Newtonsoft.Json` has been largely superseded by `System.Text.Json`[^1] (we want to make this change everywhere in #231) so this PR switches from using `Newtonsoft.Json` to `System.Text.Json` for serializing manifests. During testing this resulted in ~15% less memory allocated during serialization.

[^1]: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/migrate-from-newtonsoft?pivots=dotnet-6-0


jcfiorenzano (approved):

JamieMagee:
@jcfiorenzano. I thought this could be a quick win, but it's not possible without all the work that #231 requires. Closing for now.

PR #543, opened 2023-04-28
ci: add smoke test

melotic:
## Description

This PR adds a new GitHub Actions workflow for running smoke tests on various repositories using different package managers. The smoke tests will be executed on every push, pull request, and daily at midnight.

### Smoke Tests Workflow

The new workflow is named "Smoke Tests" and consists of the following steps:

1. Checkout Component Detection.
2. Setup .NET environment.
3. Setup NuGet cache.
4. Install Apache Ivy.
5. Checkout the smoke test repository.
6. Run the smoke test, executing the component detection tool 10 times.

The workflow uses a matrix strategy to run the tests across various repositories using different package managers, including:

- CocoaPods: [realm/realm-swift](https://github.com/realm/realm-swift)
- Gradle: [microsoft/ApplicationInsights-Java](https://github.com/microsoft/ApplicationInsights-Java)
- Go: [kubernetes/kubernetes](https://github.com/kubernetes/kubernetes)
- Maven: [apache/kafka](https://github.com/apache/kafka)
- NPM: [axios/axios](https://github.com/axios/axios)
- NuGet: [PowerShell/PowerShell](https://github.com/PowerShell/PowerShell)
- Pip: [django/django](https://github.com/django/django)
- Pnpm: [pnpm/pnpm](https://github.com/pnpm/pnpm)
- Poetry: [Textualize/rich](https://github.com/Textualize/rich)
- Ruby: [rails/rails](https://github.com/rails/rails)
- Rust: [alacritty/alacritty](https://github.com/alacritty/alacritty)
- Yarn: [gatsbyjs/gatsby](https://github.com/gatsbyjs/gatsby)

If a smoke test fails during a scheduled run, the workflow will create a new GitHub issue with the details of the failed tests. [Example](https://github.com/microsoft/component-detection/issues/542)

## Motivation

This PR aims to improve the reliability of Component Detection by adding smoke tests to ensure that the tool works correctly on a diverse set of repositories and package managers. Additionally, the recent outage proved that our CI is not rigorous enough to ensure that Component Detection actually works E2E. The smoke tests simply ensures that there are no crashes, (or crashes are extremely unlikely).

## Custom GitHub Runner

After the outage, my investigation discovered that the concurrency bugs that caused the LSI were not reproducible in our [Verification Tests](https://github.com/microsoft/component-detection/tree/cd6405a827b256fa6d0881ad495cd7dd5a0a821d/test/Microsoft.ComponentDetection.VerificationTests/resources). Running Component Detection on this proved inefficient, regardless of CPU cores. Additionally, GitHub runners have two cores, which was not enough to properly trigger the bug.

To ensure we trigger any concurrency bugs, I've created a custom 1ES Hosted Pool for GitHub Runners that has 8 cores ([D8ads_v5](https://learn.microsoft.com/en-us/azure/virtual-machines/dasv5-dadsv5-series#dasv5-series)), and dynamically scales up to 20 machines. Only this workflow will run on it, as startup times are higher than GitHub's as we have to provision the machines. I've configured an automatic scale-up that will learn our usage and forecasts our demand for these build machines. These machines will be provisioned before our forecasted demand for them, reducing time.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -85,7 +85,7 @@ jobs:
             await github.rest.issues.create({
               owner: context.repo.owner,
               repo: context.repo.repo,
-              title: 'TEST: Smoke Test Failure',
+              title: 'Smoke Test Failure',
               body: issue_body,
               labels: ['bug']
             })
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -1,0 +1,93 @@
+name: Smoke Tests
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: "0 0 * * *" # every day at midnight
+
+jobs:
+  smoke-test:
+    runs-on: ["self-hosted", "1ES.Pool=1ES-OSE-GH-Pool"]
+    strategy:
+      matrix:
+        language:
+          [
+            { name: "CocoaPods", repo: "realm/realm-swift" },
+            { name: "Gradle", repo: "microsoft/ApplicationInsights-Java" },
+            { name: "Go", repo: "kubernetes/kubernetes" },
+            { name: "Maven", repo: "apache/kafka" },
+            { name: "NPM", repo: "axios/axios" },
+            { name: "NuGet", repo: "PowerShell/PowerShell" },
+            { name: "Pip", repo: "django/django" },
+            { name: "Pnpm", repo: "pnpm/pnpm" },
+            { name: "Poetry", repo: "Textualize/rich" },
+            { name: "Ruby", repo: "rails/rails" },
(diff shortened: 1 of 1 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,91 @@
+name: Smoke Tests
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: "0 0 * * *" # every day at midnight
```
JamieMagee, line 7: Every day is fine but change the time. GitHub struggles when everyone schedules on the hour.

59 minutes past the hour is a good time 😂

JamieMagee: Oh, wait. These are our own runners :facepalm:

PR #544, opened 2023-04-28
Added a new detector MvnPomCliComponentDetector.

AbhinavAbhinav11:
This detector just finds all the .pom files and parses the dependencies (which are present similar to maven structure like<dependency> </dependency>) in those files. It does a simple text parsing and does not create the dependency graph.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/maven/MvnPomCliComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/maven/MvnPomCliComponentDetector.cs
@@ -1,0 +1,112 @@
+namespace Microsoft.ComponentDetection.Detectors.Maven;
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+using System.Xml;
+using Microsoft.ComponentDetection.Contracts;
+using Microsoft.ComponentDetection.Contracts.BcdeModels;
+using Microsoft.ComponentDetection.Contracts.Internal;
+using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using Microsoft.Extensions.Logging;
+
+public class MvnPomCliComponentDetector : FileComponentDetector
+{
+    public MvnPomCliComponentDetector(
+     IComponentStreamEnumerableFactory componentStreamEnumerableFactory,
+     IObservableDirectoryWalkerFactory walkerFactory,
+     ILogger<MvnCliComponentDetector> logger)
+    {
+        this.ComponentStreamEnumerableFactory = componentStreamEnumerableFactory;
+        this.Scanner = walkerFactory;
+        this.Logger = logger;
+    }
+
+    public override string Id => "MvnPomCli";
+
+    public override IList<string> SearchPatterns => new List<string>() { "*.pom" };
+
+    public override IEnumerable<string> Categories => new[] { Enum.GetName(typeof(DetectorClass), DetectorClass.Maven) };
+
+    public override IEnumerable<ComponentType> SupportedComponentTypes => new[] { ComponentType.Maven };
+
+    public override int Version => 2;
+
+    protected override async Task OnFileFoundAsync(ProcessRequest processRequest, IDictionary<string, string> detectorArgs)
+    {
+        await this.ProcessFileAsync(processRequest);
+    }
+
+    private async Task ProcessFileAsync(ProcessRequest processRequest)
+    {
+        var singleFileComponentRecorder = processRequest.SingleFileComponentRecorder;
+        var stream = processRequest.ComponentStream;
+
+        try
+        {
+            byte[] pomBytes = null;
+
+            if ("*.pom".Equals(stream.Pattern, StringComparison.OrdinalIgnoreCase))
+            {
+                using (var contentStream = File.Open(stream.Location, FileMode.Open))
+                {
+                    pomBytes = new byte[contentStream.Length];
+                    await contentStream.ReadAsync(pomBytes.AsMemory(0, (int)contentStream.Length));
+
+                    using var pomStream = new MemoryStream(pomBytes, false);
+                    var doc = new XmlDocument();
+                    doc.Load(pomStream);
+
+                    var nsmgr = new XmlNamespaceManager(doc.NameTable);
+                    nsmgr.AddNamespace("ns", "http://maven.apache.org/POM/4.0.0");
+
+                    var dependencies = doc.SelectSingleNode("//ns:project/ns:dependencies", nsmgr);
+                    if (dependencies == null)
+                    {
+                        return;
+                    }
+
+                    foreach (XmlNode node in dependencies.ChildNodes)
+                    {
+                        this.RegisterComponent(node, nsmgr, singleFileComponentRecorder);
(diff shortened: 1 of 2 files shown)


src/Microsoft.ComponentDetection.Detectors/maven/MvnPomCliComponentDetector.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.IO;
+using System.Threading.Tasks;
+using System.Xml;
+using Microsoft.ComponentDetection.Contracts;
+using Microsoft.ComponentDetection.Contracts.BcdeModels;
+using Microsoft.ComponentDetection.Contracts.Internal;
+using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using Microsoft.Extensions.Logging;
+
+public class MvnPomCliComponentDetector : FileComponentDetector
```
melotic, line 14: Could you also implement `IDefaultOffComponentDetector`? We don't want a new detector to be enabled by default (only explicitly opted-in) before we've had the chance to fully verify it correctly works.

src/Microsoft.ComponentDetection.Detectors/maven/MvnPomCliComponentDetector.cs
```diff
@@ -0,0 +1,112 @@
...
+    private async Task ProcessFileAsync(ProcessRequest processRequest)
+    {
+        var singleFileComponentRecorder = processRequest.SingleFileComponentRecorder;
+        var stream = processRequest.ComponentStream;
+
+        try
+        {
+            byte[] pomBytes = null;
+
+            if ("*.pom".Equals(stream.Pattern, StringComparison.OrdinalIgnoreCase))
```
melotic, line 50: This check is not necessary. Only files that match the `SearchPatterns` will call `ProcessFileAsync`

src/Microsoft.ComponentDetection.Detectors/maven/MvnPomCliComponentDetector.cs
```diff
@@ -0,0 +1,112 @@
...
+                    var nsmgr = new XmlNamespaceManager(doc.NameTable);
+                    nsmgr.AddNamespace("ns", "http://maven.apache.org/POM/4.0.0");
+
+                    var dependencies = doc.SelectSingleNode("//ns:project/ns:dependencies", nsmgr);
+                    if (dependencies == null)
+                    {
+                        return;
+                    }
+
+                    foreach (XmlNode node in dependencies.ChildNodes)
```
melotic, line 70:
```suggestion
                    foreach (var node in dependencies.ChildNodes)
```

src/Microsoft.ComponentDetection.Detectors/maven/MvnPomCliComponentDetector.cs
```diff
@@ -0,0 +1,112 @@
...
+                    {
+                        this.RegisterComponent(node, nsmgr, singleFileComponentRecorder);
+                    }
+                }
+            }
+            else
+            {
+                return;
```
melotic, line 78: You can eliminate this else 😁

src/Microsoft.ComponentDetection.Detectors/maven/MvnPomCliComponentDetector.cs
```diff
@@ -0,0 +1,112 @@
...
+
+    private void RegisterComponent(XmlNode node, XmlNamespaceManager nsmgr, ISingleFileComponentRecorder singleFileComponentRecorder)
+    {
+        var groupIdNode = node.SelectSingleNode("ns:groupId", nsmgr);
+        var artifactIdNode = node.SelectSingleNode("ns:artifactId", nsmgr);
+        var versionNode = node.SelectSingleNode("ns:version", nsmgr);
+
+        if (groupIdNode == null || artifactIdNode == null || versionNode == null)
+        {
+            return;
```
melotic, line 97: Maybe we should add a log here?

melotic:
Thanks for the contribution! I left some feedback on your PR. We also need unit tests to verify this new detector. You can view an example [here](https://github.com/microsoft/component-detection/blob/cd6405a827b256fa6d0881ad495cd7dd5a0a821d/test/Microsoft.ComponentDetection.Detectors.Tests/MvnCliDetectorTests.cs)

src/Microsoft.ComponentDetection.Detectors/maven/MvnPomCliComponentDetector.cs
AbhinavAbhinav11: Done

src/Microsoft.ComponentDetection.Detectors/maven/MvnPomCliComponentDetector.cs
AbhinavAbhinav11: Done. Thanks!

src/Microsoft.ComponentDetection.Detectors/maven/MvnPomCliComponentDetector.cs
AbhinavAbhinav11: Done.

JamieMagee:
@AbhinavAbhinav11 what the need for this detector have over the [`MvnCliComponentDetector`][1]? The only thing I can see is the ability to get a list of direct dependencies without invoking the Maven CLI.

[1]: https://github.com/microsoft/component-detection/blob/f1241afebb0975e26329bc68e73601665556fa48/src/Microsoft.ComponentDetection.Detectors/maven/MvnCliComponentDetector.cs#L17

src/Microsoft.ComponentDetection.Detectors/maven/MvnPomCliComponentDetector.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+        catch (Exception e)
+        {
+            // If something went wrong, just ignore the component
+            this.Logger.LogError(e, "Error parsing pom maven component from {PomLocation}", stream.Location);
+            singleFileComponentRecorder.RegisterPackageParseFailure(stream.Location);
+        }
+    }
+
+    private void RegisterComponent(XmlNode node, XmlNamespaceManager nsmgr, ISingleFileComponentRecorder singleFileComponentRecorder)
```
melotic, line 89: This method looks the same as `RegisterComponent(XmlNode node, XmlNamespaceManager nsmgr, ISingleFileComponentRecorder singleFileComponentRecorder)` in `MavenFileParserService`. We should refactor this and DRY.

FernandoRojo:
This PR is being closed as it is effectively superseded by the current MvnCLI detector
 
 1. The MavenWithFallbackDetector already covers the static parsing use case — and does it far better (variable resolution, parent POM inheritance, CLI-first with fallback, telemetry).
 2. Scanning *.pom artifacts is architecturally wrong — it would detect dependencies of downloaded packages rather than the user's project dependencies, No other detector works this way.
 
The only novel idea (scanning .pom files from local Maven repos) doesn't align with how component-detection works — it should scan source projects, not artifact caches, similar to how we do not rely on nupkg detection any longer.

PR #552, opened 2023-05-04
refactor: migrate to `System.Text.Json` in `Microsoft.ComponentDetection.Common`

JamieMagee:
Part of #231

- [Scan telemetry using `Newtonsoft.Json`][1]
- [Scan telemetry using `System.Text.Json`][2]

The main different is `System.Text.Json` converts potentially unsafe characters to unicode escape sequences[^1]. If we want to change this behaviour we need to use the `UnsafeRelaxedJsonEscaping` encoder. I don't think it matters much one way or the other.

[1]: https://gist.github.com/JamieMagee/8a6116e64a662ca6a8e65af9e7552167#file-after-json
[2]: https://gist.github.com/JamieMagee/8a6116e64a662ca6a8e65af9e7552167#file-before-json
[^1]: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/character-encoding

### DIFF
--- a/src/Microsoft.ComponentDetection.Common/DockerService.cs
+++ b/src/Microsoft.ComponentDetection.Common/DockerService.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


melotic (approved):
yay!

PR #557, opened 2023-05-05
refactor: remove more `Newtonsoft.Json`

JamieMagee:
Part of #231


melotic:
@JamieMagee I had to revert the changes to port the `VcpkgComponentDetector` to use `System.Text.JSON`.

`System.Test.JSON` does not support handling of missing members in .NET 6, we'd have to either upgrade .NET versions (and lose LTS :(), add the full SPDX definition to `VcpkgSBOM`, or use some library that has a full `SPDX` contract.

JamieMagee:
Closing this for now as there are too many features that are missing from `System.Text.Json` that we are missing in .NET 6. We can attempt this again once .NET 8 is released.

PR #562, opened 2023-05-10
fix(experiments): add ctor to `ExperimentDiff`

melotic:
This is needed for JSON deserialization. Also updated the xml docs to follow the "standard" text.+


JamieMagee:
Can you use the `init` auto-property to achieve the same effect?

JamieMagee (approved):

melotic:
> Can you use the `init` auto-property to achieve the same effect?

Yea, but then you can not ensure that all the properties are non-null, which isn't a problem but just prevents null refs in the future.

PR #596, opened 2023-06-06
feat: use `Newtonsoft.Json` streaming methods

JamieMagee:
As we're unable to move to `System.Text.Json` with the feature set that is available in .NET 6 (See #231 for more information), this PR optimizes our use of `Newtonsoft.Json` by using streaming methods for our most allocation-heavy methods where possible.

In tests against real world repositories this dropped memory usage by up to 37% when writing the manifest JSON.

There is also some refactoring and tech debt cleanup in tests and documentation.


melotic (approved):

melotic:
LGTM! Although I don't think this affects the manfiest, might be good to just double check that the format is still the same.

PR #621, opened 2023-06-23
Add a new detector: CondaLockComponentDetector

arzano:
This PR is adding support for Conda using conda-lock (see [https://github.com/conda/conda-lock](https://github.com/conda/conda-lock)). Full graph generation is supported this way for Conda without requiring to have Conda installed. 

The new detector has been tested locally and been verified to work as expected:
image


src/Microsoft.ComponentDetection.Detectors/conda/CondaDependencyResolver.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.ComponentDetection.Contracts;
+using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using Microsoft.ComponentDetection.Detectors.Poetry.Contracts;
+
+public static class CondaDependencyResolver
+{
+    /// <summary>
+    /// Iterate through all dependencis that are explicitly referenced in the conda environment
```
pjanowski, line 12:
```suggestion
    /// Iterate through all dependencies that are explicitly referenced in the conda environment
```

src/Microsoft.ComponentDetection.Contracts/TypedComponent/CondaLockComponent.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Collections.Generic;
+
+public class CondaLockComponent : TypedComponent
+{
+    public CondaLockComponent(string name, string version, string category, Dictionary<string, string> dependencies, Dictionary<string, string> hash, string manager, bool optional, string platform, string url)
+    {
+        this.Name = this.ValidateRequiredInput(name, nameof(this.Name), nameof(ComponentType.Conda));
+        this.Version = this.ValidateRequiredInput(version, nameof(this.Version), nameof(ComponentType.Conda));
+        this.Category = category;
+        this.Dependencies = dependencies;
```
grvillic, line 12: The component contracts should not contain the dependencies themselves, that is responsibility of the `ISingleFileComponentRecorder`, see how we are leveraging `singleFileComponentRecorder.RegisterUsage` method calls, where we specify the current component we found + the parent component.

src/Microsoft.ComponentDetection.Contracts/TypedComponent/CondaLockComponent.cs
```diff
@@ -0,0 +1,47 @@
...
+        this.Hash = hash;
+        this.Manager = manager;
+        this.Optional = optional;
+        this.Platform = platform;
+        this.Url = url;
+        this.Version = version;
```
grvillic, line 18: This is not necessary, you are already setting the validated value in line 10.
```suggestion
```

src/Microsoft.ComponentDetection.Contracts/TypedComponent/CondaLockComponent.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    public Dictionary<string, string> Hash { get; set; }
+
+    public string Manager { get; set; }
+
+    public string Name { get; set; }
+
+    public bool Optional { get; set; }
+
+    public string Platform { get; set; }
```
grvillic, line 38: Are any of these equivalent: Category, Build, Channel, Manager and Namespace? I see you used very different property names as the existing Conda component we had, are these new ones lockfile specific? We should use the most generic terminology applicable for Conda.

src/Microsoft.ComponentDetection.Detectors/conda/CondaDependencyResolver.cs
```diff
@@ -0,0 +1,110 @@
...
+    /// <summary>
+    /// Iterate through all dependencies that are explicitly referenced in the conda environment
+    /// and build as well as register the full dependency tree for each of them.
+    /// </summary>
+    /// <param name="condaLock">The full condaLock object.</param>
+    /// <param name="explicitDependencies">The list of all dependencies that are explicitly specified in the conda environment.</param>
+    /// <param name="singleFileComponentRecorder">The SingleFileComponentRecorder.</param>
+    public static void RecordDependencyGraphFromFile(CondaLock condaLock, List<string> explicitDependencies, ISingleFileComponentRecorder singleFileComponentRecorder)
+    {
+        if (condaLock != null && condaLock.Package != null)
```
grvillic, line 20: nit: equivalent expression
```suggestion
        if (condaLock?.Package != null)
```

src/Microsoft.ComponentDetection.Detectors/conda/CondaDependencyResolver.cs
```diff
@@ -0,0 +1,110 @@
...
+    /// <param name="isExplicitDependency">True if the package is explicitly referenced.</param>
+    /// <param name="dependencyPlatform">The platform of the dependency that should expanded.</param>
+    /// <param name="parentId">The id of the parent component of the dependency that should expanded.</param>
+    /// <param name="condaLock">The full condaLock object.</param>
+    /// <param name="singleFileComponentRecorder">The SingleFileComponentRecorder.</param>
+    public static void RecordPackage(CondaPackage package, bool isExplicitDependency, string dependencyPlatform, string parentId, CondaLock condaLock, ISingleFileComponentRecorder singleFileComponentRecorder)
+    {
+        var condaLockComponent = CreateCondaLockComponentFromPackage(package);
+        var pipComponent = CreatePipComponentFromPackage(package);
+        var componentId = package.Manager == "pip" ? pipComponent.Id : condaLockComponent.Id;
```
grvillic, line 71: nit: you can set a boolean variable
`var isPipManager = package.Manager.Equals("pip", StringComparison.OrdinalIgnoreCase);` and use it both in here and line 74

src/Microsoft.ComponentDetection.Detectors/conda/CondaDependencyResolver.cs
```diff
@@ -0,0 +1,110 @@
...
+    /// Converts a CondaPackage to a CondaLockComponent.
+    /// </summary>
+    /// <param name="package">The CondaPackage to convert.</param>
+    /// <returns>The CondaLockComponent.</returns>
+    public static CondaLockComponent CreateCondaLockComponentFromPackage(CondaPackage package)
+        => new CondaLockComponent(
+            package.Name,
+            package.Version,
+            package.Category,
+            package.Dependencies,
```
grvillic, line 94: As mentioned in other file, dependencies are not reported directly as part of the component. The graph is generated by the `SingleFileComponentRecorder`, I would only expect you do loop over the dependencies property and register them individually with current component as parent ID

src/Microsoft.ComponentDetection.Detectors/conda/CondaLockComponentDetector.cs
```diff
@@ -0,0 +1,126 @@
...
+    public override int Version { get; } = 1;
+
+    public override IEnumerable<string> Categories => new List<string> { "Python" };
+
+    /// <inheritdoc/>
+    protected override Task OnFileFoundAsync(ProcessRequest processRequest, IDictionary<string, string> detectorArgs)
+    {
+        var singleFileComponentRecorder = processRequest.SingleFileComponentRecorder;
+
+        this.Logger.LogDebug("Found cond-lock file: {YamlFile}", processRequest.ComponentStream.Location);
```
grvillic, line 42: typo `conda-lock`

src/Microsoft.ComponentDetection.Detectors/conda/CondaLockComponentDetector.cs
```diff
@@ -0,0 +1,126 @@
...
+            var condaLock = this.ParseCondaLock(processRequest);
+
+            // Parse conda environments to get explicit dependencies
+            var explicitDependencies = new List<string>();
+            if (condaLock != null && condaLock.Metadata != null && condaLock.Metadata.Sources != null)
+            {
+                foreach (var environment in condaLock.Metadata.Sources)
+                {
+                    var environmentFilePath = Path.Combine(processRequest.ComponentStream.Location, "..", environment);
+                    explicitDependencies.AddRange(this.ParseExplicitDependencies(environmentFilePath));
```
grvillic, line 55: Is the environment file necessary to capture the dependencies? Shouldn't those be declared in conda-lock file directly or why the need to open a separate file? I don't have much context on this manifest, so it may be necessary I just want to understand the reasoning behind that behavior.

src/Microsoft.ComponentDetection.Detectors/conda/CondaLockComponentDetector.cs
```diff
@@ -0,0 +1,126 @@
...
+            {
+                if (item is string)
+                {
+                    // Add all conda dependencies
+                    explicitDependencies.Add(item.ToString());
+                }
+                else if (item is Dictionary<object, object> pipDependencies)
+                {
+                    // Add all pip dependencies
+                    if (pipDependencies.First().Key.ToString() == "pip" &&
```
grvillic, line 111: This looks a bit fragile. Are we sure this is a dictionary? I see we doing `pipDependencies.First()` which makes we think it isn't an actual dictionary/list. Also string checks can be fragile, maybe use case insensitive check like the other example I shared, something like `key.Equals("pip", StringComparison.OrdinalIgnoreCase)`

src/Microsoft.ComponentDetection.Detectors/conda/CondaLockComponentDetector.cs
```diff
@@ -0,0 +1,126 @@
...
+            var deserializer = new DeserializerBuilder()
+                .IgnoreUnmatchedProperties()
+                .Build();
+            var condaEnvironment = deserializer.Deserialize<CondaEnvironment>(new StreamReader(environmentFilePath));
+            foreach (var item in condaEnvironment.Dependencies)
+            {
+                if (item is string)
+                {
+                    // Add all conda dependencies
+                    explicitDependencies.Add(item.ToString());
```
grvillic, line 106: Is it possible to declare multiple dependencies with same name in CondaEnvironment? If so, we may want to deduplicate entries here using a `HashSet<string>` to ensure we don't keep dups.

grvillic (changes requested):

src/Microsoft.ComponentDetection.Contracts/TypedComponent/CondaLockComponent.cs
arzano: Great call out! 

As discussed offline also, we don't actually need all of these information for component governance. That's why I refactored the code to only use the minimal set of required properties (the same way as the PipComponent works). This especially has the benefit that we will avoid adding duplicate packages. The previous implementation detected 215 Conda packages in the verification tests. A lot of them have been duplicates. Now, with the new implementation we are down to 88 packages in the verification list. This is the number of deduplicate packages in the conda-lock file.

src/Microsoft.ComponentDetection.Contracts/TypedComponent/CondaLockComponent.cs
arzano: I refactored the code to use the existing CondaComponent. So this should be resolved.

src/Microsoft.ComponentDetection.Detectors/conda/CondaDependencyResolver.cs
arzano: Great call out! Adapted.

src/Microsoft.ComponentDetection.Detectors/conda/CondaDependencyResolver.cs
arzano: As discussed in the other file, the code has been refactored accordingly. :)

src/Microsoft.ComponentDetection.Detectors/conda/CondaLockComponentDetector.cs
arzano: All dependencies are in the conda-lock file. The initial idea here was to use the environment.yml file to identify the packages that have been referenced directly to set the isReferencedDirectly flag correctly. 

I fully agree that parsing another file complicates the code though. I refactored the code to identify directly referenced packages by searching for packages that don't have any ancestors. This way we can avoid parsing the additional file.

src/Microsoft.ComponentDetection.Detectors/conda/CondaLockComponentDetector.cs
arzano: Fully agreed. As discussed in the other comment, I refactored the code to not parse the additional file anymore. That way we can get rid of this logic.

src/Microsoft.ComponentDetection.Detectors/conda/CondaLockComponentDetector.cs
arzano: Great questiion!

The conda-lock file will be deduplicated. It will contain multiple entries with the same package name for different platforms though. That's why I refactored the code to only store the minimal required information in CondaComponent (similar to how this is done in PipComponent). That way, the deduplication works totally fine now. 

This can be seen in the verification tests also. While we detected 215 Conda packages in the first implementation, now we are down to 88 detected packages. I manually verified that this is the deduplicated number of packages in the example conda-lock file.

src/Microsoft.ComponentDetection.Detectors/conda/CondaLockComponentDetector.cs
arzano: Great catch! Fixed.

src/Microsoft.ComponentDetection.Detectors/conda/CondaDependencyResolver.cs
arzano: Great catch as well! Adopted.

src/Microsoft.ComponentDetection.Contracts/TypedComponent/CondaLockComponent.cs
arzano: Great call out! 

As discussed in another comment, this has been refactored and the existing Conda contract is used now while passing along the minimal amount of information that are required for component governance (similar to how PipComponent is doing this.).

grvillic (approved):
Thank you for the contribution!

Issue #663, opened 2023-07-17
Yarn v3+ support and performance


JamieMagee:
Currently, Component Detection explicitly supports Yarn v1 and v2. We've had some customer reports of poor performance of our yarn detection.

Unfortunately, yarn support has gotten a little more complex recently. Yarn v1 is now considered [yarn 'classic'][1]. [Yarn 2 and 3][2] are a complete rewrite of yarn (and are sometimes called 'berry'). The lockfile version has incremented a lot.

| Yarn version       | lockfile version |
|--------------------|------------------|
| 1+                 | 1                |
| [`2.0.0-rc.2`][3]  | 2                |
| [`2.0.0-rc.5`][4]  | 3                |
| [`2.0.0-rc.15`][5] | 4                |
| [`3.1.0-rc.12`][6] | 5                |
| [`3.2.0-rc.12`][7] | 6                |
| [`4.0.0-rc.1`][8]  | 7                |
| [`4.0.0-rc.47`][9] | 8                |

We need to evaluate whether or not these lockfile changes are relevant to Component Detection, and the performance of our yarn detection for large projects in general.

[1]: https://classic.yarnpkg.com/lang/en/
[2]: https://yarnpkg.com/
[3]: https://github.com/yarnpkg/berry/commit/2f9e8073d15745f9d53e6b8b42fa9c81eb143d54
[4]: https://github.com/yarnpkg/berry/commit/b19708918a098f0453b0951f99123205c5b90c54
[5]: https://github.com/yarnpkg/berry/commit/aaa92b944673e05c9e3eeceedcd19ff98c480c82
[6]: https://github.com/yarnpkg/berry/commit/ccb0788bcb86005773fa06605d904f37db4c0a16
[7]: https://github.com/yarnpkg/berry/commit/d024a60e45018ec7e6fb3002f555c9be51f10d34
[8]: https://github.com/yarnpkg/berry/commit/1fbd616d346e71563b0712d7d75c645dd6cd812a
[9]: https://github.com/yarnpkg/berry/commit/21fcc792bf83825aa051c51c96cdc683ee85d20b

AB#2099326

lvivski:
Hi @JamieMagee, I see you had several PRs already merged. Can Component Detection already be used with new Yarn versions?

MLoughry:
[...]

PR #664, opened 2023-07-18
feat: add `YarnBerryTypeConverter` a custom YAML converter for `yarn.lock`

JamieMagee:
This is a first step for #663 

There's a comment in our code that there is no off-the-shelf YAML parser that works for `yarn.lock` files:

https://github.com/microsoft/component-detection/blob/018aa3973f3fecd61b46e8be858898f450c6c134/src/Microsoft.ComponentDetection.Detectors/yarn/Parsers/YarnBlockFile.cs#L13-L14

While it's true that no .NET off-the-shelf parser supports `yarn.lock` files out-of-the-box, using `YamlDotNet` with a custom converter, we are actually able to parse `yarn.lock` files much faster than using regex, as we currently do.

This isn't hooked up anywhere yet, but I wanted to get it out for review ASAP.


src/Microsoft.ComponentDetection.Detectors/yarn/Contracts/YarnBerryLockfileEntry.cs
JamieMagee, line 1: TODO: add documentation about which lockfile versions each property was added in.

src/Microsoft.ComponentDetection.Detectors/yarn/Contracts/YarnBerryLockfile.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace Microsoft.ComponentDetection.Detectors.Yarn.Contracts;
+
+using System.Collections.Generic;
+using YamlDotNet.Serialization;
+
+/// <summary>
+/// Represents the yarn.lock file for yarn v2+.
+/// </summary>
+public sealed record YarnBerryLockfile
```
melotic, line 9: Can we add a link to the spec for the yarn lockfile?

melotic (approved):

src/Microsoft.ComponentDetection.Detectors/yarn/Contracts/YarnBerryLockfile.cs
JamieMagee: There is no spec :(

PR #668, opened 2023-07-20
Fix Rust detector exception for non-source dependency

arlosi:
Remove Rust detector exception for a Source dependency having a dependency on a non-Source dependency.

This is seen in the Rust project's lockfile: https://github.com/rust-lang/rust/blob/1554942cdc0099e081d5cd673d45f8a09cfd0eb0/Cargo.lock#L6 where `rustc-std-workspace-alloc` has no `source`, dispite being a dependency of `addr2line` (which does have a `source`).


melotic:
LGTM. Would it be possible to update our [verification tests](https://github.com/microsoft/component-detection/tree/main/test/Microsoft.ComponentDetection.VerificationTests/resources/rust) to have this scenario as well?

Issue #629, opened 2023-06-26
Migrate to PyPI simple API PEP 691


JamieMagee:
Currently our pip detector uses [PyPI's JSON API][1]. However this API is deprecated in favour of the 'simple' API outlined in [PEP 691][2]. We should migrate our detector from the deprecated API to the simple API, and perform a couple of cleanup items along the way.

## Current state

In a simplified description, the current [`PipComponentDetector`][3] first calls [`PythonResolver.ResolveRootsAsync`][4], which in turn calls [`PyPiClient.FetchPackageDependenciesAsync`][5] which finally makes an HTTP request to `https://pypi.org/pypi/{package-name}/json`. This happens for each top-level package listed in `requirements.txt`, and the valid versions are saved in a dictionary.

Once the top-level dependencies are registered, [`PythonResolver.ProcessQueueAsync`][6] walks the top-level dependencies and calls [`PyPiClient.FetchPackageDependenciesAsync`][7] to download a [Python wheel][8]. From the Python wheel, the [`METADATA` file is extracted][9] to read the dependencies for the specific package. This process is then repeated for each node in the dependency graph, until the entire graph is resolved.

This currently has 3 major problems:
1. We're using the old PyPI API, which is deprecated
2. `PyPiClient` is doing too much
	- It not only makes HTTP calls to pypi.org, but also extracts files from wheels and adds packages to the graph
	- This makes it _very_ difficult to unit test
	- It needs to be split into a client and a service that calls the client
3. `PyPiClient` caches the entire `HttpResponseMessage`
	- This includes the entire content of the response, the headers, as well as the request
	- Instead the cache should return:
		- A `SimpleProject` object for calls to `https://pypi.org/simple/{package-name}/`
		- A `Stream` for calls to `https://files.pythonhosted.org/packages/...` (or maybe just a `Stream` for the `METADATA` file specifically. TBD)

## Step 1: `SimplyPyPiClient`

Step one of this refactoring involves writing a `SimplePyPiClient` class to replace the current `PyPiClient`. It should expose two methods:
- The first should accept a package name, and return an object representing the JSON response from `https://pypi.org/simple/{package-name}/`
	- [PEP 691][2] has full details for this API, but do note that the `application/vnd.pypi.simple.v1+html` content type is required to receive JSON responses
- The second should accept a path under `https://files.pythonhosted.org/packages/` and return a `Stream` of the wheel file (or perhaps a `Stream` of the `METADATA` file in the wheel. TBD.)

[...]

PR #672, opened 2023-07-25
New PypiClient using the new Simple Api

Omotola:
This PR adds a new client that calls the new simple pypi api [PEP 691](https://peps.python.org/pep-0691/#why-not-break-out-other-pieces-of-information-from-the-filename). The client contains methods to retrieve and cache the project json and wheel files from the api. #629 

 There will be follow up prs for a new pythonresolver and detector that can be tested alongside the old pypi detector


src/Microsoft.ComponentDetection.Detectors/pip/ISimplePyPiClient.cs
```diff
@@ -0,0 +1,310 @@
...
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using Polly;
+
+[assembly: InternalsVisibleTo("Microsoft.ComponentDetection.Detectors.Tests")]
+
+namespace Microsoft.ComponentDetection.Detectors.Pip;
+
+public interface ISimplePyPiClient
```
JamieMagee, line 21: nit: `ISimplePyPiClient` and `SimplePyPiClient` should be in different files.

src/Microsoft.ComponentDetection.Detectors/pip/ISimplePyPiClient.cs
```diff
@@ -0,0 +1,310 @@
...
+{
+    Task<Stream> FetchPackageFileStreamAsync(string name, string version, PythonProjectRelease release);
+
+    Task<SimplePypiProject> GetSimplePypiProjectAsync(PipDependencySpecification spec);
```
JamieMagee, line 25: Can you add some documentation to each of these methods using the xmldoc format? i.e.

```
/// <summary>
/// Some description
/// <summary>
/// <para, name="name">Some parameter description</param>
```

Copilot can be useful to write these quickly.

JamieMagee: Ah, just saw you have them on the implementation below. The standard in C# is to have the documentation on the interface, and then put `/// <inheritdoc />` on the implementation.

src/Microsoft.ComponentDetection.Detectors/pip/ISimplePyPiClient.cs
```diff
@@ -0,0 +1,310 @@
...
+    // retries used so far for calls to pypi.org
+    private long retries;
+
+    public SimplePyPiClient() => this.cacheTelemetry = new SimplePypiCacheTelemetryRecord()
+    {
+        NumProjectFileCacheHits = 0,
+        NumSimpleProjectCacheHits = 0,
+        FinalProjectFileCacheSize = 0,
+        FinalSimpleProjectCacheSize = 0,
+    };
```
JamieMagee, line 76: No need to explicitly initialize these. They are `int` type, which are 0 by default. Also, I would just initialize `cacheTelemetry` in the declaration, not the constructor, because the constructor doesn't actually affect the `cacheTelemetry` property.

src/Microsoft.ComponentDetection.Detectors/pip/SimplePypiProject.cs
```diff
@@ -0,0 +1,11 @@
+namespace Microsoft.ComponentDetection.Detectors.Pip;
+
+using System.Collections.Generic;
+
+/// <summary>
+/// A project from the new simple pypi api.
+/// </summary>
+public class SimplePypiProject
```
JamieMagee, line 8: nit:
```suggestion
public sealed record SimplePypiProject
```

src/Microsoft.ComponentDetection.Detectors/pip/SimplePypiProject.cs
```diff
@@ -0,0 +1,11 @@
...
+{
+    public IList<SimplePypiProjectRelease> Files { get; set; }
```
JamieMagee, line 10: nit:
```suggestion
    public IList<SimplePypiProjectRelease> Files { get; init; }
```

src/Microsoft.ComponentDetection.Detectors/pip/SimplePypiProjectRelease.cs
```diff
@@ -0,0 +1,15 @@
...
+/// A specific release of a project from the new simple pypi api.
+/// </summary>
+public class SimplePypiProjectRelease
+{
+    public string FileName { get; set; }
+
+    public double Size { get; set; }
+
+    public Uri Url { get; set; }
+}
```
JamieMagee, line 15: nit:
```suggestion
public sealed record SimplePypiProjectRelease
{
    public string FileName { get; init; }

    public double Size { get; init; }

    public Uri Url { get; init; }
}
```

src/Microsoft.ComponentDetection.Detectors/pip/ISimplePyPiClient.cs
```diff
@@ -0,0 +1,310 @@
...
+
+        var responseContent = await response.Content.ReadAsStringAsync();
+        if (string.IsNullOrEmpty(responseContent))
+        {
+            return pythonProject;
+        }
+
+        try
+        {
+            pythonProject = JsonConvert.DeserializeObject<SimplePypiProject>(responseContent);
```
melotic, line 194: Is it possible to refactor the Retry logic so we can utilize the async features of System.Text.Json? HttpClient has extension methods ([`HttpClient.GetFromJsonAsync`](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.json.httpclientjsonextensions.getfromjsonasync)) to make a GET request and deserialize json asynchronously

test/Microsoft.ComponentDetection.Detectors.Tests/Utilities/SimplePypiClientTests.cs
```diff
@@ -0,0 +1,366 @@
...
+using Moq.Protected;
+using Newtonsoft.Json;
+
+[TestClass]
+public class SimplePyPiClientTests
+{
+    private readonly SimplePyPiClient simplePypiClient;
+
+    public SimplePyPiClientTests() => this.simplePypiClient = new SimplePyPiClient(
+        new EnvironmentVariableService(),
```
JamieMagee, line 28: Mock this as well

src/Microsoft.ComponentDetection.Detectors/pip/ISimplePyPiClient.cs
```diff
@@ -0,0 +1,310 @@
...
+using System.Net.Http.Headers;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.ComponentDetection.Common.Telemetry.Records;
+using Microsoft.ComponentDetection.Contracts;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
```
JamieMagee, line 14: Can you switch out `Newtonsoft.Json` for `System.Text.Json`?

src/Microsoft.ComponentDetection.Detectors/pip/ISimplePyPiClient.cs
```diff
@@ -0,0 +1,310 @@
...
+public sealed class SimplePyPiClient : ISimplePyPiClient, IDisposable
+{
+    // Values used for cache creation
+    private const long CACHEINTERVALSECONDS = 60;
+    private const long DEFAULTCACHEENTRIES = 4096;
+
+    // max number of retries allowed, to cap the total delay period
+    public const long MAXRETRIES = 15;
+
+    private static readonly HttpClientHandler HttpClientHandler = new HttpClientHandler() { CheckCertificateRevocationList = true };
```
JamieMagee, line 37: Is `CheckCertificateRevocationList` not default?

src/Microsoft.ComponentDetection.Detectors/pip/ISimplePyPiClient.cs
```diff
@@ -0,0 +1,310 @@
...
+                   return false;
+               }
+
+               var statusCode = (int)message.StatusCode;
+
+               // only retry if server doesn't classify the call as a client error!
+               var isRetryable = statusCode != 300 && statusCode != 406 && (statusCode < 400 || statusCode > 499);
+               return !message.IsSuccessStatusCode && isRetryable;
+           })
+           .WaitAndRetryAsync((int)MAXRETRIES - 1, i => RETRYDELAY, (result, timeSpan, retryCount, context) =>
```
JamieMagee, line 248: Does this API return any `Retry-After` header that we can use instead of a static value?

src/Microsoft.ComponentDetection.Detectors/pip/ISimplePyPiClient.cs
```diff
@@ -0,0 +1,310 @@
...
+        {
+            NumProjectFileCacheHits = 0,
+            NumSimpleProjectCacheHits = 0,
+            FinalProjectFileCacheSize = 0,
+            FinalSimpleProjectCacheSize = 0,
+        };
+        this.logger = logger;
+    }
+
+    public static HttpClient HttpClient { get; internal set; } = new HttpClient(HttpClientHandler);
```
JamieMagee, line 91: Is it possible to inject this with dependency injection? See https://learn.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests

test/Microsoft.ComponentDetection.Detectors.Tests/Utilities/SimplePypiClientTests.cs
```diff
@@ -0,0 +1,366 @@
...
+
+    [TestMethod]
+    public async Task GetSimplePypiProject_DuplicateEntries_CallsGetAsync_OnceAsync()
+    {
+        var pythonSpecs = new PipDependencySpecification { DependencySpecifiers = new List<string> { "==1.0.0" }, Name = "boto3" };
+        var pythonProject = this.SampleValidApiJsonResponse("boto3", "0.0.1");
+        var mockHandler = this.MockHttpMessageHandler(pythonProject, HttpStatusCode.OK);
+        SimplePyPiClient.HttpClient = new HttpClient(mockHandler.Object);
+
+        Func<Task> action = async () => await this.simplePypiClient.GetSimplePypiProjectAsync(pythonSpecs);
```
JamieMagee, line 56: nit 
```suggestion
        var action = async () => await this.simplePypiClient.GetSimplePypiProjectAsync(pythonSpecs);
```

test/Microsoft.ComponentDetection.Detectors.Tests/Utilities/SimplePypiClientTests.cs
```diff
@@ -0,0 +1,366 @@
...
+        var pythonSpecs = new PipDependencySpecification { DependencySpecifiers = new List<string> { "==1.0.0" } };
+        var pythonProject = new SimplePypiProject()
+        {
+            Files = new List<SimplePypiProjectRelease> { new SimplePypiProjectRelease() },
+        };
+
+        var mockHandler = this.MockHttpMessageHandler(JsonConvert.SerializeObject(pythonProject), HttpStatusCode.OK);
+        SimplePyPiClient.HttpClient = new HttpClient(mockHandler.Object);
+
+        Func<Task> action = async () =>
```
JamieMagee, line 81: nit: 
```suggestion
        var action = async () =>
```

test/Microsoft.ComponentDetection.Detectors.Tests/Utilities/SimplePypiClientTests.cs
```diff
@@ -0,0 +1,366 @@
...
+        {
+            pythonSpecs.Name = Guid.NewGuid().ToString();
```
JamieMagee, line 83: Extract this line above. It doesn't need to be in the function.

test/Microsoft.ComponentDetection.Detectors.Tests/Utilities/SimplePypiClientTests.cs
```diff
@@ -0,0 +1,366 @@
...
+
+    [TestMethod]
+    public async Task GetSimplePypiProject_InvalidSpec_NotThrowAsync()
+    {
+        var pythonSpecs = new PipDependencySpecification { DependencySpecifiers = new List<string> { "==1.0.0" }, Name = "randomName" };
+
+        var mockHandler = this.MockHttpMessageHandler("404 Not Found", HttpStatusCode.NotFound);
+        SimplePyPiClient.HttpClient = new HttpClient(mockHandler.Object);
+
+        Func<Task> action = async () => await this.simplePypiClient.GetSimplePypiProjectAsync(pythonSpecs);
```
JamieMagee, line 127:
```suggestion
        var action = async () => await this.simplePypiClient.GetSimplePypiProjectAsync(pythonSpecs);
```

JamieMagee: And all the way down too

src/Microsoft.ComponentDetection.Detectors/pip/ISimplePyPiClient.cs
Omotola: Doesn't seem like it does. I don't see it mentioned in the docs

test/Microsoft.ComponentDetection.Detectors.Tests/Utilities/SimplePypiClientTests.cs
Omotola: I think it actually should be in the function since the test is to make sure the request isnt pulling a cached value, and having the spec name in the function makes sure the api url (constructed using the specname) is different each time the function is called.

src/Microsoft.ComponentDetection/Program.cs
```diff
@@ -24,6 +24,7 @@
         .CreateBootstrapLogger();
 
     var serviceProvider = new ServiceCollection()
+        .AddHttpClient()
```
melotic, line 27: This should be added in the `AddComponentDetection` method, as other code that uses CD as a library will fail to inject the `IHttpFactory`

src/Microsoft.ComponentDetection.Detectors/pip/ISimplePyPiClient.cs
Omotola: synced with Jamie. might be a little complicated to redo the retry logic due to the status codes and retry criteria

PR #677, opened 2023-07-27
ci: fix issues sync workflow

JamieMagee:
Used the wrong token 🤦 

Also allow manual invocation of this workflow.


melotic (approved):
thanks!

PR #683, opened 2023-07-31
Users/oakeredolu/newpipservice

Omotola:
Added a new python resolver for the new [simple pypi client](https://github.com/microsoft/component-detection/pull/672). The main changes from the old python resolver is that this python resolver handles converting the project we get from simple pypi into internal contract ( `PythonProjectRelease`), and it extracts and resolves dependencies from the python wheel. 

Also moved `PythonResolverState` to its own file and made it public so its accessible from the new resolver.

#672


melotic:
It's hard to review this since it also includes changes from #672

src/Microsoft.ComponentDetection.Detectors/pip/SimplePythonResolver.cs
```diff
@@ -0,0 +1,374 @@
...
+
+    /// <summary>
+    /// Uses regex to extract the version from the file name.
+    /// </summary>
+    /// <param name="fileName"> the name of the file from simple pypi. </param>
+    /// <returns> returns a string representing the release version. </returns>
+    private string GetVersionFromFileName(string fileName)
+    {
+        fileName = fileName.Replace(".tar", "-tar");
+        var version = Regex.Match(fileName, @"-(\d+\.((\d+)|(\d+\.(\d+))))-").Groups[1];
```
melotic, line 209: I think we can get rid of the string replace and instead use this regex:

```regex
-(\d+\.\d+\.\d+)(.tar|-)
```

src/Microsoft.ComponentDetection.Detectors/pip/SimplePythonResolver.cs
```diff
@@ -0,0 +1,374 @@
...
+                {
+                    content.Add(line);
+                }
+            }
+        }
+
+        // Pull the packages that aren't conditional based on "extras"
+        // Right now we just want to resolve the graph as most comsumers will
+        // experience it
+        foreach (var deps in content.Where(x => !x.Contains("extra ==")))
```
melotic, line 281: Is `extra==` also valid?

src/Microsoft.ComponentDetection.Detectors/pip/SimplePythonResolver.cs
```diff
@@ -0,0 +1,374 @@
...
+    /// Given a state, node, and new spec, will reprocess a new valid version for the node.
+    /// </summary>
+    /// <param name="state"> The PythonResolverState. </param>
+    /// <param name="node"> The PipGraphNode. </param>
+    /// <param name="newSpec"> The PipDependencySpecification. </param>
+    /// <returns> Returns true if the node can be reprocessed else false. </returns>
+    private async Task<bool> InvalidateAndReprocessAsync(
+    PythonResolverState state,
+    PipGraphNode node,
+    PipDependencySpecification newSpec)
```
melotic, line 299: nit: indent these

src/Microsoft.ComponentDetection.Detectors/pip/SimplePythonResolver.cs
```diff
@@ -0,0 +1,374 @@
...
+            return string.Empty;
+        }
+    }
+
+    /// <summary>
+    /// Uses regex to extract the version from the file name.
+    /// </summary>
+    /// <param name="fileName"> the name of the file from simple pypi. </param>
+    /// <returns> returns a string representing the release version. </returns>
+    private string GetVersionFromFileName(string fileName)
```
melotic, line 206: Can we add tests for this function?

src/Microsoft.ComponentDetection.Detectors/pip/SimplePythonResolver.cs
Omotola: Changed it to this `-(\d+\.\d+(\.\d)*)(.tar|-)` to match both versions like 1.0 and 1.0.0

src/Microsoft.ComponentDetection.Detectors/pip/SimplePythonResolver.cs
Omotola: added 2 more tests for this

src/Microsoft.ComponentDetection.Detectors/pip/SimplePythonResolver.cs
Omotola: Not sure, but the lines with the `extra==` seem to be valid packages but are conditional/optional which is why we skip them. I found an example in the metadata file image below. It shows botocore is already required and then botocore[crt] is an extra
image

melotic: I think this check needs to be more resilient. I've seen a variety of packages with strangely formatted `Requires-Dist`. I think in the Python Version class we have some regexes to parse parts of these out.

Omotola: Still a bit unclear. The Requires-Dist lines are still parsed with the `RequiresDistRegex` in `PipDependencySpecification` 
`Requires-Dist:\s*([^\s;\[<>=!~]+)(?:\[[^\]]+\])?(?:\s*\(([^)]+)\))?([^;]*)`

melotic: Discussed offline, looks fine.

melotic (approved):

PR #684, opened 2023-07-31
Users/oakeredolu/newpipdetector

Omotola:
Added a new experimental pip detector that uses the [simple pypi resolver](https://github.com/microsoft/component-detection/pull/684) and [client](https://github.com/microsoft/component-detection/pull/672). It has been tested locally and it detects the same components as the baseline detector. #672 
image

**Tested with more components:**
image

Diff is 0
image


src/Microsoft.ComponentDetection.Detectors/pip/SimplePipComponentDetector.cs
```diff
@@ -0,0 +1,133 @@
...
+        ILogger<SimplePipComponentDetector> logger)
+    {
+        this.ComponentStreamEnumerableFactory = componentStreamEnumerableFactory;
+        this.Scanner = walkerFactory;
+        this.pythonCommandService = pythonCommandService;
+        this.pythonResolver = pythonResolver;
+        this.Logger = logger;
+    }
+
+    public override string Id => "Pip";
```
melotic, line 32: This needs to a different ID, it's the same as the old pip detector

src/Microsoft.ComponentDetection.Detectors/pip/SimplePipComponentDetector.cs
```diff
@@ -0,0 +1,133 @@
...
+
+    public override IList<string> SearchPatterns => new List<string> { "setup.py", "requirements.txt" };
+
+    public override IEnumerable<string> Categories => new List<string> { "Python" };
+
+    public override IEnumerable<ComponentType> SupportedComponentTypes { get; } = new[] { ComponentType.Pip };
+
+    public override int Version { get; } = 6;
```
melotic, line 40:
```suggestion
    public override int Version { get; } = 1;
```

src/Microsoft.ComponentDetection.Detectors/pip/SimplePipComponentDetector.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reactive.Linq;
+using System.Threading.Tasks;
+using Microsoft.ComponentDetection.Contracts;
+using Microsoft.ComponentDetection.Contracts.Internal;
+using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using Microsoft.Extensions.Logging;
+
+public class SimplePipComponentDetector : FileComponentDetector, IExperimentalDetector
```
melotic, line 13: Can we mark this ass DefaultOff for now, merge this, iterate over performance concerns, then we can:

1. Leave as DefaultOff and let some customers test this new detector for us
2. Run as experimental

Omotola: updated it

src/Microsoft.ComponentDetection.Detectors/pip/SimplePipComponentDetector.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reactive.Linq;
+using System.Threading.Tasks;
+using Microsoft.ComponentDetection.Contracts;
+using Microsoft.ComponentDetection.Contracts.Internal;
+using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using Microsoft.Extensions.Logging;
+
+public class SimplePipComponentDetector : FileComponentDetector, IDefaultOffComponentDetector, IExperimentalDetector
```
melotic, line 13: Remove `IExperimentalDetector`

PR #688, opened 2023-07-31
feat: disable detector experiments by default

### LINKED ISSUE #682: Make Detector Experiments Opt-in
We currently run **always** run experiments. Instead, they should be explicitly opt-in, with an Environment Variable, or a manually set flag.

1P use cases will set this flag, as 3P customers don't care about our experiments to validate detectors.

AB#2087261

melotic:
Closes #682

### DIFF
--- a/src/Microsoft.ComponentDetection.Orchestrator/Experiments/DetectorExperiments.cs
+++ b/src/Microsoft.ComponentDetection.Orchestrator/Experiments/DetectorExperiments.cs
@@ -15,5 +15,5 @@ public static class DetectorExperiments
     private static bool EnvironmentEnabled =>
         !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("CD_DETECTOR_EXPERIMENTS"));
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.ComponentDetection.Orchestrator/Experiments/DetectorExperiments.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    /// <summary>
+    /// Manually enables detector experiments.
+    /// </summary>
+    public static bool Enable { get; set; }
+
+    private static bool EnvironmentEnabled =>
+        !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("CD_DETECTOR_EXPERIMENTS"));
+
+    internal static bool RunExperiments => Enable || EnvironmentEnabled;
```
grvillic, line 18: nit: renaming this to `AreExperimentsEnabled` to be concise about what it does.

grvillic (approved):

PR #692, opened 2023-08-02
Feature: Add conan detector that parses conan.lock files of conan package manager version 1.x

ashokgowtham:
This PR adds support for identifying components from conan.lock files created by conan package manager. 

Doesn't require conan package manager to have been installed.

### DIFF
--- a/docs/detectors/conan.md
+++ b/docs/detectors/conan.md
@@ -3,7 +3,7 @@
 Conan detection relies on a conan.lock file being present.
 
 ## Detection strategy
-Conan detection is performed by parsing every <em>conan.lock</em> found under the scan directory.
+Conan detection is performed by parsing every **conan.lock** found under the scan directory.
 
 ## Known limitations
 Conan detection will not work if lock files are not being used or not yet generated. So ensure to run the conan build to generate the lock file(s) before running the scan.
--- a/docs/detectors/conan.md
+++ b/docs/detectors/conan.md
@@ -8,4 +8,4 @@ Conan detection is performed by parsing every **conan.lock** found under the sca
 ## Known limitations
 Conan detection will not work if lock files are not being used or not yet generated. So ensure to run the conan build to generate the lock file(s) before running the scan.
 
-Full dependency graph generation is not supported. However, dependency relationships identified/present in the <em>conan.lock</em> file is captured.
+Full dependency graph generation is not supported. However, dependency relationships identified/present in the **conan.lock** file is captured.
--- a/docs/detectors/conan.md
+++ b/docs/detectors/conan.md
@@ -1,0 +1,11 @@
+# Conan Detection
+## Requirements
+Conan detection relies on a conan.lock file being present.
+
+## Detection strategy
+Conan detection is performed by parsing every <em>conan.lock</em> found under the scan directory.
+
+## Known limitations
+Conan detection will not work if lock files are not being used or not yet generated. So ensure to run the conan build to generate the lock file(s) before running the scan.
+
+Full dependency graph generation is not supported. However, dependency relationships identified/present in the <em>conan.lock</em> file is captured.
--- a/src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
@@ -13,8 +13,6 @@ using Microsoft.Extensions.Logging;
 
 public class ConanLockComponentDetector : FileComponentDetector, IDefaultOffComponentDetector
 {
-    private const string ConanLockSearchPattern = "conan.lock";
-
     public ConanLockComponentDetector(
         IComponentStreamEnumerableFactory componentStreamEnumerableFactory,
         IObservableDirectoryWalkerFactory walkerFactory,
@@ -27,7 +25,7 @@ public class ConanLockComponentDetector : FileComponentDetector, IDefaultOffComp
 
     public override string Id => "ConanLock";
 
-    public override IList<string> SearchPatterns => new List<string> { ConanLockSearchPattern };
+    public override IList<string> SearchPatterns => new List<string> { "conan.lock" };
 
     public override IEnumerable<ComponentType> SupportedComponentTypes => new[] { ComponentType.Conan };
 
@@ -40,11 +38,9 @@ public class ConanLockComponentDetector : FileComponentDetector, IDefaultOffComp
         var singleFileComponentRecorder = processRequest.SingleFileComponentRecorder;
         var conanLockFile = processRequest.ComponentStream;
 
-        await Task.CompletedTask;   // Satisfy the async signature of the overridden method.
-
         try
         {
-            var conanLock = JsonSerializer.Deserialize<ConanLock>(conanLockFile.Stream);
+            var conanLock = await JsonSerializer.DeserializeAsync<ConanLock>(conanLockFile.Stream);
             this.RecordLockfileVersion(conanLock.Version);
             if (conanLock.Version != "0.4")
             {
--- a/src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
@@ -46,6 +46,11 @@ public class ConanLockComponentDetector : FileComponentDetector, IDefaultOffComp
         {
             var conanLock = JsonSerializer.Deserialize<ConanLock>(conanLockFile.Stream);
             this.RecordLockfileVersion(conanLock.Version);
+            if (conanLock.Version != "0.4")
+            {
+                this.Logger.LogWarning("Unsupported conan.lock file version '{ConanLockVersion}'. Failed to process conan.lock file '{ConanLockLocation}'", conanLock.Version, conanLockFile.Location);
+                return;
+            }
 
             if (!conanLock.HasNodes())
             {
@@ -58,8 +63,8 @@ public class ConanLockComponentDetector : FileComponentDetector, IDefaultOffComp
             if (packagesDictionary.ContainsKey("0"))
             {
                 packagesDictionary.Remove("0", out var rootNode);
-                explicitReferencedDependencies = rootNode.Requires;
-                developmentDependencies = rootNode.BuildRequires;
+                explicitReferencedDependencies = rootNode.Requires ?? Array.Empty<string>();
+                developmentDependencies = rootNode.BuildRequires ?? Array.Empty<string>();
             }
 
             foreach (var (packageIndex, package) in packagesDictionary)
--- a/src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
@@ -54,27 +54,25 @@ public class ConanLockComponentDetector : FileComponentDetector, IDefaultOffComp
 
             var packagesDictionary = conanLock.GraphLock.Nodes;
             var explicitReferencedDependencies = Array.Empty<string>();
-            string rootComponentId = null;
+            var developmentDependencies = Array.Empty<string>();
             if (packagesDictionary.ContainsKey("0"))
             {
                 packagesDictionary.Remove("0", out var rootNode);
-                var rootComponent = rootNode.ToComponent();
-                rootComponentId = rootComponent.Id;
-                singleFileComponentRecorder.RegisterUsage(new DetectedComponent(rootComponent), true);
                 explicitReferencedDependencies = rootNode.Requires;
+                developmentDependencies = rootNode.BuildRequires;
             }
 
             foreach (var (packageIndex, package) in packagesDictionary)
             {
-                var isExplicitReferencedDependency = explicitReferencedDependencies.Contains(packageIndex);
-                var parentId = isExplicitReferencedDependency ? rootComponentId : null;
-                singleFileComponentRecorder.RegisterUsage(new DetectedComponent(package.ToComponent()), isExplicitReferencedDependency, parentId);
+                singleFileComponentRecorder.RegisterUsage(
+                    new DetectedComponent(package.ToComponent()),
+                    isExplicitReferencedDependency: explicitReferencedDependencies.Contains(packageIndex),
+                    isDevelopmentDependency: developmentDependencies.Contains(packageIndex));
             }
 
-            var packages = packagesDictionary.Values;
             foreach (var (conanPackageIndex, package) in packagesDictionary)
             {
-                var parentPackages = packages.Where(package => package.Requires?.Contains(conanPackageIndex) == true);
+                var parentPackages = packagesDictionary.Values.Where(package => package.Requires?.Contains(conanPackageIndex) == true);
                 foreach (var parentPackage in parentPackages)
                 {
                     singleFileComponentRecorder.RegisterUsage(new DetectedComponent(package.ToComponent()), false, parentPackage.ToComponent().Id, isDevelopmentDependency: false);
--- a/src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
@@ -42,11 +42,6 @@ public class ConanLockComponentDetector : FileComponentDetector, IDefaultOffComp
         {
             var conanLock = await JsonSerializer.DeserializeAsync<ConanLock>(conanLockFile.Stream);
             this.RecordLockfileVersion(conanLock.Version);
-            if (conanLock.Version != "0.4")
-            {
-                this.Logger.LogWarning("Unsupported conan.lock file version '{ConanLockVersion}'. Failed to process conan.lock file '{ConanLockLocation}'", conanLock.Version, conanLockFile.Location);
-                return;
-            }
 
             if (!conanLock.HasNodes())
             {
--- a/src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
@@ -1,0 +1,90 @@
+﻿namespace Microsoft.ComponentDetection.Detectors.Conan;
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Microsoft.ComponentDetection.Contracts;
+using Microsoft.ComponentDetection.Contracts.Internal;
+using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using Microsoft.ComponentDetection.Detectors.Conan.Contracts;
+using Microsoft.Extensions.Logging;
+
+public class ConanLockComponentDetector : FileComponentDetector, IDefaultOffComponentDetector
+{
+    private const string ConanLockSearchPattern = "conan.lock";
+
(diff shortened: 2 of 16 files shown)


ashokgowtham:

docs/detectors/conan.md
```diff
@@ -0,0 +1,11 @@
...
+## Requirements
+Conan detection relies on a conan.lock file being present.
+
+## Detection strategy
+Conan detection is performed by parsing every <em>conan.lock</em> found under the scan directory.
+
+## Known limitations
+Conan detection will not work if lock files are not being used or not yet generated. So ensure to run the conan build to generate the lock file(s) before running the scan.
+
+Full dependency graph generation is not supported. However, dependency relationships identified/present in the <em>conan.lock</em> file is captured.
```
melotic, line 11:
```suggestion
Full dependency graph generation is not supported. However, dependency relationships identified/present in the **conan.lock** file is captured.
```

src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Threading.Tasks;
+using Microsoft.ComponentDetection.Contracts;
+using Microsoft.ComponentDetection.Contracts.Internal;
+using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using Microsoft.ComponentDetection.Detectors.Conan.Contracts;
+using Microsoft.Extensions.Logging;
+
+public class ConanLockComponentDetector : FileComponentDetector, IDefaultOffComponentDetector
+{
+    private const string ConanLockSearchPattern = "conan.lock";
```
melotic, line 16: nit: inline this in `SearchPatterns`

src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
```diff
@@ -0,0 +1,93 @@
...
+    protected override async Task OnFileFoundAsync(ProcessRequest processRequest, IDictionary<string, string> detectorArgs)
+    {
+        var singleFileComponentRecorder = processRequest.SingleFileComponentRecorder;
+        var conanLockFile = processRequest.ComponentStream;
+
+        await Task.CompletedTask;   // Satisfy the async signature of the overridden method.
+
+        try
+        {
+            var conanLock = JsonSerializer.Deserialize<ConanLock>(conanLockFile.Stream);
```
melotic, line 47: Can we use `JsonSerializer.DeserializeAsync` and remove the above `await Task.CompletedTask` above?

src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
```diff
@@ -0,0 +1,93 @@
...
+            this.RecordLockfileVersion(conanLock.Version);
+            if (conanLock.Version != "0.4")
```
melotic, line 49: I don't think we should cancel detection if the lockfile versions don't match. While there might be a new lockfile version, there is a chance that we can still parse it. We should attempt to parse it and let detection fail if its a breaking change

src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
```diff
@@ -0,0 +1,93 @@
...
+                return;
+            }
+
+            var packagesDictionary = conanLock.GraphLock.Nodes;
+            var explicitReferencedDependencies = Array.Empty<string>();
+            var developmentDependencies = Array.Empty<string>();
+            if (packagesDictionary.ContainsKey("0"))
+            {
+                packagesDictionary.Remove("0", out var rootNode);
+                explicitReferencedDependencies = rootNode.Requires ?? Array.Empty<string>();
```
melotic, line 66: nit: can we put this into a hashset for O(1) contains?

src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
```diff
@@ -0,0 +1,93 @@
...
+                    isDevelopmentDependency: developmentDependencies.Contains(packageIndex));
+            }
+
+            foreach (var (conanPackageIndex, package) in packagesDictionary)
+            {
+                var parentPackages = packagesDictionary.Values.Where(package => package.Requires?.Contains(conanPackageIndex) == true);
+                foreach (var parentPackage in parentPackages)
+                {
+                    singleFileComponentRecorder.RegisterUsage(new DetectedComponent(package.ToComponent()), false, parentPackage.ToComponent().Id, isDevelopmentDependency: false);
+                }
```
melotic, line 84: Do these need to be seperate foreach loops? Can they be merged?

src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
```diff
@@ -0,0 +1,93 @@
...
+            }
+
+            var packagesDictionary = conanLock.GraphLock.Nodes;
+            var explicitReferencedDependencies = Array.Empty<string>();
+            var developmentDependencies = Array.Empty<string>();
+            if (packagesDictionary.ContainsKey("0"))
+            {
+                packagesDictionary.Remove("0", out var rootNode);
+                explicitReferencedDependencies = rootNode.Requires ?? Array.Empty<string>();
+                developmentDependencies = rootNode.BuildRequires ?? Array.Empty<string>();
```
melotic, line 67: nit: can we put these into hashset for O(1) contains?

src/Microsoft.ComponentDetection.Detectors/conan/Contracts/ConanLock.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    [JsonPropertyName("profile_build")]
+    public string ProfileBuild { get; set; }
+
+    [JsonPropertyName("graph_lock")]
+    public ConanLockGraph GraphLock { get; set; }
+
+    public override bool Equals(object obj) => obj is ConanLock conanLock && this.Version == conanLock.Version && this.ProfileHost == conanLock.ProfileHost && this.ProfileBuild == conanLock.ProfileBuild && this.GraphLock.Equals(conanLock.GraphLock);
+
+    public override int GetHashCode() => HashCode.Combine(this.Version, this.ProfileHost, this.ProfileBuild, this.GraphLock);
```
melotic, line 22: I dont think these are necessary

src/Microsoft.ComponentDetection.Detectors/conan/Contracts/ConanLockGraph.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    [JsonPropertyName("revisions_enabled")]
+    public bool RevisionsEnabled { get; set; }
+
+    [JsonPropertyName("nodes")]
+    public Dictionary<string, ConanLockNode> Nodes { get; set; }
+
+    public override bool Equals(object obj) => obj is ConanLockGraph graph && this.RevisionsEnabled == graph.RevisionsEnabled && this.Nodes.Count == graph.Nodes.Count && !this.Nodes.Except(graph.Nodes).Any();
+
+    public override int GetHashCode() => HashCode.Combine(this.RevisionsEnabled, this.Nodes);
```
melotic, line 18: same here

src/Microsoft.ComponentDetection.Detectors/conan/Contracts/ConanLockNode.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    [JsonPropertyName("requires")]
+    public string[] Requires { get; set; }
+
+    [JsonPropertyName("build_requires")]
+    public string[] BuildRequires { get; set; }
+
+    public override bool Equals(object obj) => obj is ConanLockNode node && this.Context == node.Context && this.Modified == node.Modified && this.Options == node.Options && this.PackageId == node.PackageId && this.Path == node.Path && this.Previous == node.Previous && this.Reference == node.Reference && Enumerable.SequenceEqual(this.Requires, node.Requires);
+
+    public override int GetHashCode() => HashCode.Combine(this.Context, this.Modified, this.Options, this.PackageId, this.Path, this.Previous, this.Reference, this.Requires);
```
melotic, line 39: here too

test/Microsoft.ComponentDetection.Detectors.Tests/ConanLockComponentDetectorTests.cs
```diff
@@ -0,0 +1,271 @@
...
+                            Requires = new[] { "0" },
+                        }
+                    },
+                },
+            },
+        };
+        var jsonString = JsonSerializer.Serialize(sampleConanLock, new JsonSerializerOptions { DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingDefault });
+        var expectedJsonString = @"{""version"":""0.4"",""profile_build"":""someProfileBuild"",""graph_lock"":{""revisions_enabled"":true,""nodes"":{""0"":{""context"":"""",""modified"":false,""options"":"""",""package_id"":""SomePackageId"",""path"":""D:/SomePath"",""prev"":""SomeOtherId"",""ref"":"""",""requires"":[""0""]}}}}";
+
+        Assert.AreEqual(expectedJsonString, jsonString);
```
melotic, line 150: nit: use FluentAssertions

```suggestion
	jsonString.Should().Be(expectedJsonString)
        Assert.AreEqual(expectedJsonString, jsonString);
```

test/Microsoft.ComponentDetection.Detectors.Tests/ConanLockComponentDetectorTests.cs
```diff
@@ -0,0 +1,271 @@
...
+                            Previous = "SomeOtherId",
+                            Reference = string.Empty,
+                            Requires = new[] { "0" },
+                        }
+                    },
+                },
+            },
+        };
+        var jsonString = JsonSerializer.Serialize(sampleConanLock, new JsonSerializerOptions { DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingDefault });
+        var expectedJsonString = @"{""version"":""0.4"",""profile_build"":""someProfileBuild"",""graph_lock"":{""revisions_enabled"":true,""nodes"":{""0"":{""context"":"""",""modified"":false,""options"":"""",""package_id"":""SomePackageId"",""path"":""D:/SomePath"",""prev"":""SomeOtherId"",""ref"":"""",""requires"":[""0""]}}}}";
```
melotic, line 148: Are we just testing JSON deserialization? I'm not sure if this is necessary.

melotic:
Thanks for the contribution @ashokgowtham! Some minor code review comments and this looks good to go :)

docs/detectors/conan.md
```diff
@@ -0,0 +1,11 @@
+# Conan Detection
+## Requirements
+Conan detection relies on a conan.lock file being present.
+
+## Detection strategy
+Conan detection is performed by parsing every <em>conan.lock</em> found under the scan directory.
```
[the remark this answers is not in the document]
ashokgowtham: Good catch!

src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
ashokgowtham: Would version>="0.4" make more sense? Since the previous versions is not something I've tested with.

test/Microsoft.ComponentDetection.Detectors.Tests/ConanLockComponentDetectorTests.cs
ashokgowtham: I used this test to get the JSON serialization annotations to working state. May not add much value unless we are making changes to the annotations, in which case it can help avoid breaking the contract unexpectedly.

src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
ashokgowtham: This is where we are identifying the relationships of every package, so it is a nested loop.
The nesting can be reduced with Linq or SelectMany(), but the complexity doesn't reduce.

src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
ashokgowtham: That's good catch too! updated.

src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
ashokgowtham: done

src/Microsoft.ComponentDetection.Detectors/conan/Contracts/ConanLock.cs
ashokgowtham: Agreed, only used in unit tests. Removed them.

test/Microsoft.ComponentDetection.Detectors.Tests/ConanLockComponentDetectorTests.cs
ashokgowtham: Done

src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
ashokgowtham: Of course! Definitely not a nit, it could end up performing poorly for projects with large number of dependencies.
Addressed.

src/Microsoft.ComponentDetection.Detectors/conan/Contracts/ConanLockGraph.cs
ashokgowtham: Addressed

src/Microsoft.ComponentDetection.Detectors/conan/Contracts/ConanLockNode.cs
ashokgowtham: Addressed

test/Microsoft.ComponentDetection.Detectors.Tests/ConanLockComponentDetectorTests.cs
ashokgowtham: Removed it.

src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
melotic: Ah, I meant with the loop above, not the nested for loop. There are two ` foreach (var (conanPackageIndex, package) in packagesDictionary)`.

src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                packagesDictionary.Remove("0", out var rootNode);
+                if (rootNode.Requires != null)
+                {
+                    explicitReferencedDependencies = new HashSet<string>(rootNode.Requires);
+                }
+
+                if (rootNode.BuildRequires != null)
+                {
+                    developmentDependencies = new HashSet<string>(rootNode.BuildRequires);
```
melotic, line 64: Can `rootNode.Requires` or `rootNode.BuildRequires` be null? If they're null, the constructor to `HashSet<string>` will throw an exception. Maybe `rootNode?.Requires.ToHashSet() ?? new HashSet<string>()`?

ashokgowtham: Yes `rootNode.Requires` or `rootNode.BuildRequires` can be null and that is why they are wrapped with `!= null` checks.
The `explicitReferencedDependencies` and `developmentDependencies` are initialized with empty sets to ensure they are never null.

melotic (approved):

src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
ashokgowtham: > Maybe `rootNode?.Requires`

`rootNode` _should not_ be `null` in the conan.lock format as I understand. But in case it is `null` for whatever reason then the detection will break. It would be better to just ignore the particular component instead of not scanning that file at all. So I'll add a check for that.

src/Microsoft.ComponentDetection.Detectors/conan/ConanLockComponentDetector.cs
ashokgowtham: oh, that's a good catch! Yes, that can be combined in its current state. That's a remnant of evolution of this code over multiple passes.

test/Microsoft.ComponentDetection.Detectors.Tests/ConanLockComponentDetectorTests.cs
```diff
@@ -0,0 +1,266 @@
...
+        rootComponents.ForEach(rootComponentId =>
+        {
+            try
+            {
+                graph.IsComponentExplicitlyReferenced(rootComponentId).Should().BeTrue();
+            }
+            catch
+            {
+                Assert.Fail($"Expected Component to be explicitly referenced but it is not: {rootComponentId}");
+            }
```
melotic, line 171:
```suggestion
                graph.IsComponentExplicitlyReferenced(rootComponentId).Should().BeTrue($"Expected Component to be explicitly referenced but it is not: {rootComponentId}");
```

test/Microsoft.ComponentDetection.Detectors.Tests/ConanLockComponentDetectorTests.cs
```diff
@@ -0,0 +1,266 @@
...
+                packageDependencyRoots[packageName].Select(expectedRoot =>
+                    new Func<ConanComponent, bool>(parentComponent => parentComponent.Name == expectedRoot)).ToArray());
+
+            componentNames.Add(packageName);
+        }
+
+        // Verify all packages were detected
+        foreach (var expectedPackage in packageVersions.Keys)
+        {
+            Assert.IsTrue(componentNames.Contains(expectedPackage));
```
melotic, line 252:
```suggestion
            componentNames.Should().Contain(expectedPackage);
```

test/Microsoft.ComponentDetection.Detectors.Tests/ConanLockComponentDetectorTests.cs
```diff
@@ -0,0 +1,266 @@
...
+
+    [TestMethod]
+    public async Task TestConanDetector_SupportNoDependenciesAsync()
+    {
+        var (result, componentRecorder) = await this.DetectorTestUtility
+            .WithFile("conan.lock", this.testConanLockNoDependenciesString)
+            .ExecuteDetectorAsync();
+
+        result.ResultCode.Should().Be(ProcessingResultCode.Success);
+        componentRecorder.GetDetectedComponents().Count().Should().Be(0);
```
melotic, line 264:
```suggestion
        componentRecorder.GetDetectedComponents().Should().BeEmpty();
```

ashokgowtham:
@Omotola Can you help merge this PR or let me know if you have any suggestions

melotic (approved):

melotic:
There are some build errors, then we can merge this.

https://github.com/microsoft/component-detection/actions/runs/5768854806/job/15718475416?pr=692

ashokgowtham:
Fixed the formatting issue

PR #695, opened 2023-08-03
Users/oakeredolu/pipsharedcache

Omotola:
Shared cache for the old and new pip detectors to help with performance


src/Microsoft.ComponentDetection.Detectors/pip/IPyPiClient.cs
```diff
@@ -138,61 +143,66 @@ public async Task<IList<PipDependencySpecification>> FetchPackageDependenciesAsy
...
+                    Interlocked.Increment(ref this.retries);
+                })
+                .ExecuteAsync(() =>
+                {
+                    if (Interlocked.Read(ref this.retries) >= MAXRETRIES)
+                    {
+                        return Task.FromResult<HttpResponseMessage>(null);
+                    }
+
+                    return this.GetAndCachePyPiResponseAsync(requestUri);
```
melotic, line 185: is this bypassing the cache?

Omotola: yes, it should be adding to cached instead. fixed it.

Omotola: actually I think I misunderstood your question. do you mean is the shared cache bypassing this internal cache? yes it is, i'm checking the shared cache first and if it returns nothing, then it enters this retry block and tries to get from the internal cache and if that returns nothing, it sends a request to the pypi api.

PR #703, opened 2023-08-08
feat(pip): fetch roots in parallel

### LINKED ISSUE #702: Fetch Pip Root Packages in Parallel
Currently, the Simple Pip detector iteratively solves the dependency tree by starting at the roots. However, since resolving the roots does not require keeping a state, we can download and parse the roots independently of each other -- in parallel.

Specifically, this loop code can be parallelized for huge performance:

https://github.com/microsoft/component-detection/blob/c6bbc3aa1ea7b29089a75ace0e612df4f08c56c8/src/Microsoft.ComponentDetection.Detectors/pip/SimplePythonResolver.cs#L93-L136

AB#2090980

melotic:
Achieves decent performance gains (25%!). Benchmarked with [`hyperfine`](https://github.com/sharkdp/hyperfine)

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `main` | 10.594 ± 4.735 | 5.115 | 17.982 | 1.25 ± 0.70 |
| `parallel-pip` | 8.456 ± 2.792 | 5.438 | 13.066 | 1.00 |

Fixes #702

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/pip/SimplePythonResolver.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/pip/SimplePythonResolver.cs
@@ -88,14 +88,18 @@ public class SimplePythonResolver : ISimplePythonResolver
     public async Task<IList<PipGraphNode>> ResolveRootsAsync(ISingleFileComponentRecorder singleFileComponentRecorder, IList<PipDependencySpecification> initialPackages)
     {
         var state = new PythonResolverState();
+        var syncRoot = new object();
 
         // Fill the dictionary with valid packages for the roots
-        foreach (var rootPackage in initialPackages)
+        await Parallel.ForEachAsync(initialPackages, async (rootPackage, ct) =>
         {
             // If we have it, we probably just want to skip at this phase as this indicates duplicates
-            if (state.ValidVersionMap.TryGetValue(rootPackage.Name, out _))
+            lock (syncRoot)
             {
-                continue;
+                if (state.ValidVersionMap.TryGetValue(rootPackage.Name, out _))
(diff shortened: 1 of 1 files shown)


src/Microsoft.ComponentDetection.Detectors/pip/SimplePythonResolver.cs
```diff
@@ -292,7 +300,7 @@ private async Task<IList<PipGraphNode>> ProcessQueueAsync(ISingleFileComponentRe
     private async Task<IList<PipDependencySpecification>> FetchDependenciesFromPackageStreamAsync(string name, string version, Stream packageStream)
     {
         var dependencies = new List<PipDependencySpecification>();
-        using var package = new ZipArchive(packageStream);
```
melotic, line 295: We should probably update this code to not pass around streams which can be disposed.

Omotola (approved):

src/Microsoft.ComponentDetection.Detectors/pip/SimplePythonResolver.cs
Omotola: should we create a separate issue to work on this? Seems it might be a bigger change. We would need to update the in resolver, client, cache, and tests

melotic: Yes. I'm not sure of the entirely familiar with how these streams are used, but I believe if they're always parsed as `ZipArchive`, then we should just be storing `ZipArchive`s and not the underlying `MemoryStream`.

PR #725, opened 2023-08-15
fix(nuget): null ref when parsing `paket.lock`

### LINKED ISSUE #724: Issue with NuGet Detector parsing paket.lock file
There is a bug present in version 3.3.12 of Component-Detection where if the directory contains a paket.lock file with the following contents:
`NUGET
  remote: https://api.nuget.org/v3/index.json
    GitVersion.CommandLine (5.5.1)
    Pester (4.10.1)
  remote: https://www.powershellgallery.com/api/v2
    InvokeBuild (5.6.3)`

it will throw the following error and skip the NuGet detector leading to incomplete scans:
`[14:21:07 ERR] Error parsing NuGet component from {redactedPath}
System.ArgumentNullException: Buffer cannot be null. (Parameter 'buffer')
   at System.IO.MemoryStream..ctor(Byte[] buffer, Boolean writable)
   at Microsoft.ComponentDetection.Detectors.NuGet.NuGetComponentDetector.ProcessFileAsync(ProcessRequest processRequest) in C:\repos\component-detection\src\Microsoft.ComponentDetection.Detectors\nuget\NuGetComponentDetector.cs:line 114`

AB#2095585

melotic:
Fixes #724

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/nuget/NuGetComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/nuget/NuGetComponentDetector.cs
@@ -176,7 +176,7 @@ public class NuGetComponentDetector : FileComponentDetector
                 }
                 catch (Exception e)
                 {
-                    this.Logger.LogError(e, "Failed to parse paket.lock component from line `{Line}` in {Location}", line, stream.Location);
+                    this.Logger.LogWarning(e, "Failed to parse paket.lock component from line `{Line}` in {Location}", line, stream.Location);
                     singleFileComponentRecorder.RegisterPackageParseFailure(stream.Location);
                 }
             }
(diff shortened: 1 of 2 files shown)


src/Microsoft.ComponentDetection.Detectors/nuget/NuGetComponentDetector.cs
```diff
@@ -166,10 +167,18 @@ private void ParsePaketLock(ProcessRequest processRequest)
...
+                try
+                {
+                    var name = match.Groups[1].Value;
+                    var version = match.Groups[2].Value;
+                    var component = new NuGetComponent(name, version);
+                    singleFileComponentRecorder.RegisterUsage(new DetectedComponent(component));
+                }
+                catch (Exception e)
+                {
+                    this.Logger.LogError(e, "Failed to parse paket.lock component from line `{Line}` in {Location}", line, stream.Location);
```
cobya, line 179: We should probably make this a warning and not a full error

cobya (approved):

PR #730, opened 2023-08-18
Downgrade invalid `packages.config` files from an error to warning

JamieMagee:
This prevents issues for those that have configured error logs in CI to break builds. Really, we should continue, and a single invalid `packages.config` file shouldn't break a build.

Also added documentation and a new test.

### DIFF
--- a/test/Microsoft.ComponentDetection.Detectors.Tests/nuget/NuGetPackagesConfigDetectorTests.cs
+++ b/test/Microsoft.ComponentDetection.Detectors.Tests/nuget/NuGetPackagesConfigDetectorTests.cs
@@ -25,10 +25,29 @@ public class NuGetPackagesConfigDetectorTests : BaseDetectorTest<NuGetPackagesCo
             .WithFile("packages.config", packagesConfig)
             .ExecuteDetectorAsync();
 
+        scanResult.ResultCode.Should().Be(ProcessingResultCode.Success);
         var detectedComponents = componentRecorder.GetDetectedComponents();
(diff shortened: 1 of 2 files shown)


test/Microsoft.ComponentDetection.Detectors.Tests/nuget/NuGetPackagesConfigDetectorTests.cs
```diff
@@ -25,10 +25,29 @@ public async Task Should_WorkAsync()
...
     }
+
+    [TestMethod]
+    public async Task Should_SkipWithInvalidVersionAsync()
+    {
+        var packagesConfig =
+            @"<?xml version=""1.0"" encoding=""utf-8""?>
+                <packages>
+                    <package id=""jQuery"" version=""3.1.1"" targetFramework=""net46"" />
+                    <package id=""NLog"" version=""
```
melotic, line 43: Ideally we would be able to detect `jQuery`, but NuGet throws an error because the whole file isn't valid XML :(

melotic (approved):

test/Microsoft.ComponentDetection.Detectors.Tests/nuget/NuGetPackagesConfigDetectorTests.cs
JamieMagee: Yeah, and we defer the actual detection to NuGet itself. IMO that's a good thing, because the behaviour is aligned.

PR #733, opened 2023-08-22
Add Conan package information to SBOMPackage and added MD5,SHA1

tarun06:
1) This PR adds Conan package information to SBOMPackage.
2) This PR also adds MD5, SHA1Hash information collected from Conan.Lock file.
3) PR also have support for Conan package url.
   example -> https://conan.io/center/recipes/{packageName}?version=(packageVersion)


src/Microsoft.ComponentDetection.Contracts/TypedComponent/ConanComponent.cs
```diff
@@ -9,16 +9,27 @@ private ConanComponent()
...
 
     public string Version { get; set; }
 
+    public string Md5Hash { get; set; }
+
+    public string Sha1Hash { get; set; }
+
+    public string Supplier { get; }
+
+    public string PackageSourceURL => $"https://conan.io/center/recipes/{this.Name}?version={this.Version}";
```
melotic, line 31: Are these two necessary? I definitely agree with the hashes.

The supplier is always `ConanCenter` and the `PackageSourceUrl` can be computed by the end user. I don't believe we store the source URL in any typed component.

tarun06: 1) PackageSourceURL is actually the download URL which end user can compute definitely.. But since, we have corresponding attribute "downloadLocation", I believe it is better to calculate and assign this so that end user can navigate to the url with one click. 

example of data

`{
      "name": "jwt-cpp",
      "SPDXID": "SPDXRef-Package-38296822FEC0FAB8266C8F2C99ABBDA7BB985BA19F82BDF9CD0C1A86A8D76751",
      "downloadLocation": "https://conan.io/center/recipes/jwt-cpp?version=0.6.0#9ba33a57e1b44045f50633246c660de4",
      "filesAnalyzed": false,
      "licenseConcluded": "NOASSERTION",
      "licenseDeclared": "NOASSERTION",
      "copyrightText": "NOASSERTION",
      "versionInfo": "0.6.0#9ba33a57e1b44045f50633246c660de4",
      "externalRefs": [
        {
          "referenceCategory": "PACKAGE-MANAGER",
          "referenceType": "purl",
          "referenceLocator": "pkg:conan//jwt-cpp@0.6.0#9ba33a57e1b44045f50633246c660de4#"
        }
      ],
      "supplier": "Conancenter"
 }`
 
2) I agrees that supplier is hard coded value as **Conancenter** and can be left as "NOASSERTION" .

melotic (approved):

PR #734, opened 2023-08-22
feat(go): break down telemetry on go cli command failures

annaowens:
**Changes**
- Create telemetry object after project roots command succeeds to avoid empty telemetry objects
- Remove `WasGoCliNotFound` as it was the exact inverse of `IsGoInstalled`
- Add `DidGoCliCommandFail` and `GoCliCommandError` to track `go list` failures

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
@@ -1,4 +1,4 @@
-namespace Microsoft.ComponentDetection.Detectors.Go;
+﻿namespace Microsoft.ComponentDetection.Detectors.Go;
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -142,6 +142,8 @@ private async Task<bool> UseGoCliToScanAsync(string location, ISingleFileCompone
         {
             this.Logger.LogError("Go CLI command \"go list -m -json all\" failed with error: {GoDependenciesProcessStdErr}", goDependenciesProcess.StdErr);
             this.Logger.LogError("Go CLI could not get dependency build list at location: {Location}. Fallback go.sum/go.mod parsing will be used.", location);
+            record.DidGoCliCommandFail = true;
+            record.GoCliCommandError = goDependenciesProcess.StdErr;
```
annaowens, line 146: There's a chance this is too long and will cause JSON truncation - will keep an eye on that.

melotic (approved):

PR #739, opened 2023-08-22
Fix to remove default off detectors from experiments

Omotola:
PR to remove experiments for unused/disabled detectors from experiments that run. Tested locally with the NewPipDetectorExperiment. The experiment runs only when the detector is enabled as expected.

### DIFF
--- a/src/Microsoft.ComponentDetection.Orchestrator/Experiments/ExperimentService.cs
+++ b/src/Microsoft.ComponentDetection.Orchestrator/Experiments/ExperimentService.cs
@@ -110,15 +110,9 @@ public class ExperimentService : IExperimentService
 
     public void RemoveUnwantedExperimentsbyDetectors(IEnumerable<IComponentDetector> detectors)
     {
-        var experimentsToRemove = new List<IExperimentConfiguration>();
-
-        foreach (var detector in detectors)
-        {
-            experimentsToRemove.AddRange(this.experiments
-                .Where(x => x.Key.IsInControlGroup(detector) || x.Key.IsInExperimentGroup(detector))
-                .Select(x => x.Key)
-                .ToList());
-        }
+        var experimentsToRemove = this.experiments
+            .Where(x => detectors.Any(detector => x.Key.IsInControlGroup(detector) || x.Key.IsInExperimentGroup(detector)))
+            .Select(x => x.Key).ToList();
 
         foreach (var config in experimentsToRemove.Where(config => this.experiments.TryRemove(config, out _)))
         {
--- a/src/Microsoft.ComponentDetection.Orchestrator/Experiments/ExperimentService.cs
+++ b/src/Microsoft.ComponentDetection.Orchestrator/Experiments/ExperimentService.cs
@@ -108,6 +108,24 @@ public class ExperimentService : IExperimentService
         }
     }
 
+    public void RemoveUnwantedExperimentsbyDetectors(IEnumerable<IComponentDetector> detectors)
+    {
+        var experimentsToRemove = new List<IExperimentConfiguration>();
+
+        foreach (var detector in detectors)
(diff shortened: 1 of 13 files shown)


src/Microsoft.ComponentDetection.Orchestrator/Experiments/ExperimentService.cs
```diff
@@ -108,6 +108,24 @@ private void FilterExperiments(IComponentDetector detector, int count)
         }
     }
 
+    public void RemoveUnwantedExperimentsbyDetectors(IEnumerable<IComponentDetector> detectors)
+    {
+        var experimentsToRemove = new List<IExperimentConfiguration>();
```
melotic, line 113: This allocation isn't necessary. You can just do `var experimentsToRemove = [linq query]`

src/Microsoft.ComponentDetection.Orchestrator/Services/IDetectorProcessingService.cs
```diff
@@ -8,4 +8,6 @@
 public interface IDetectorProcessingService
 {
     Task<DetectorProcessingResult> ProcessDetectorsAsync(IDetectionArguments detectionArguments, IEnumerable<IComponentDetector> detectors, DetectorRestrictions detectorRestrictions);
+
+    void SetUnusedDetectors(IEnumerable<IComponentDetector> detectors);
```
melotic, line 12: super nit: I don't like this. Is there any way we could maybe hook into the detector restriction server service instead?

src/Microsoft.ComponentDetection.Orchestrator/Experiments/ExperimentService.cs
```diff
@@ -108,6 +108,24 @@ private void FilterExperiments(IComponentDetector detector, int count)
         }
     }
 
+    public void RemoveUnwantedExperimentsbyDetectors(IEnumerable<IComponentDetector> detectors)
```
melotic, line 111: Could you add tests for this?

src/Microsoft.ComponentDetection.Orchestrator/Services/IDetectorProcessingService.cs
Omotola: update: added the disabled detectors to `DetectorRestrictions`

src/Microsoft.ComponentDetection.Orchestrator/Services/BcdeScanExecutionService.cs
```diff
@@ -39,6 +39,7 @@ public async Task<ScanResult> ExecuteScanAsync(IDetectionArguments detectionArgu
 
         var detectorRestrictions = this.GetDetectorRestrictions(detectionArguments);
         var detectors = this.detectorRestrictionService.ApplyRestrictions(detectorRestrictions, this.detectors).ToImmutableList();
```
cobya, line 41: nit: I know this isn't your variable but if we're going to be using `this.detectors` and `detectors` in the same function we should probably rename it to something like `restrictedDetectors`

src/Microsoft.ComponentDetection.Orchestrator/Services/DetectorProcessingService.cs
```diff
@@ -349,4 +352,6 @@ private void LogTabularOutput(ILogger logger, ConcurrentDictionary<string, Detec
             this.logger.LogInformation("{Line}", line);
         }
     }
+
+    public void SetUnusedDetectors(IEnumerable<IComponentDetector> detectors) => this.unusedDetectors = detectors;
```
cobya, line 356: Is this needed anymore? I don't see any references to unusedDetectors

cobya (approved):

PR #740, opened 2023-08-22
docs(go): improve docs on troubleshooting default go detection strategy

annaowens:
Improving documentation on how to ensure the default Go detection strategy runs.

This documentation will be linked in communications that encourage users to troubleshoot the reasons why they are seeing the fallback Go detection strategy running in their pipelines.


annaowens:
Closing in favor of #821

PR #751, opened 2023-08-29
feat(go): better `go.mod` scanning for go >= 1.17

### LINKED ISSUE #737: Only parse `go.sum` if go version <= 1.16
Currently, the go detector parses components by either:

1. Use the go CLI command to get the dependencies.
   a. For large projects, this can time out, or the build machine may not have go installed.
2. Fall back to scanning the aggregate of a `go.sum` and the `go.mod` file.
   a. However, the `go.sum` file contains dependencies that may not actually be used in the project, leading to overreporting. 

In go > 1.17, the `go.mod` file now contains both direct and indirect dependencies[^1]. 

Importantly, the `go.mod` file contains the go version. So, better logic to scan go components is:

1. Use the Go CLI to scan if available / not disabled.
2. Check the `go.mod` for the go version.
a. If version >= 1.17, Only scan the `go.mod` file for component
b. else, fall back to the old `go.mod` + `go.sum` logic.

[^1]: https://go.dev/doc/modules/gomod-ref

AB#2098642

melotic:
This PR adds support for `go.mod` files that were created with go versions >= 1.17.

Since go 1.17, `go.mod` files now contain all transitive dependencies[^1], _in a separate requires_ section. This means that we do not have to parse the `go.sum` file for dependencies, which can contain dependencies that aren't actually used.

Additionally, the `go.mod` file can require dependencies without a section, like:

```
require github.com/go-sql-driver/mysql v1.7.1 // indirect
```

We also now support two `go.mod` `require` sections:

```
module microsoft/component-detection

go 1.18

require (
        github.com/go-sql-driver/mysql v1.7.1
        rsc.io/quote v1.5.2
)

require (
        golang.org/x/text v0.0.0-20170915032832-14c0d48ead0c // indirect
        rsc.io/sampler v1.3.0 // indirect
)
```

Closes #737 

[^1]: https://go.dev/doc/modules/gomod-ref#go-notes

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
@@ -5,8 +5,10 @@ using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
+using System.Reactive.Linq;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
+using Microsoft.ComponentDetection.Common;
 using Microsoft.ComponentDetection.Common.Telemetry.Records;
 using Microsoft.ComponentDetection.Contracts;
 using Microsoft.ComponentDetection.Contracts.Internal;
@@ -16,11 +18,11 @@ using Newtonsoft.Json;
 
 public class GoComponentDetector : FileComponentDetector
 {
-    private static readonly Regex GoSumRegex = new Regex(
+    private static readonly Regex GoSumRegex = new(
         @"(?<name>.*)\s+(?<version>.*?)(/go\.mod)?\s+(?<hash>.*)",
         RegexOptions.Compiled | RegexOptions.ExplicitCapture | RegexOptions.IgnoreCase);
 
-    private readonly HashSet<string> projectRoots = new HashSet<string>();
+    private readonly HashSet<string> projectRoots = new();
 
     private readonly ICommandLineInvocationService commandLineInvocationService;
     private readonly IEnvironmentVariableService envVarService;
@@ -39,7 +41,7 @@ public class GoComponentDetector : FileComponentDetector
         this.Logger = logger;
     }
 
-    public override string Id { get; } = "Go";
+    public override string Id => "Go";
 
     public override IEnumerable<string> Categories => new[] { Enum.GetName(typeof(DetectorClass), DetectorClass.GoMod) };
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -159,36 +234,53 @@ private async Task<bool> UseGoCliToScanAsync(string location, ISingleFileCompone
...
+    private async Task ParseGoModFileAsync(
         ISingleFileComponentRecorder singleFileComponentRecorder,
         IComponentStream file,
         GoGraphTelemetryRecord goGraphTelemetryRecord)
     {
-        using var reader = new StreamReader(file.Stream);
-
-        var line = reader.ReadLine();
-        while (line != null && !line.StartsWith("require ("))
+        void TryRegisterDependency(string s)
```
jcfiorenzano, line 242: nit: pick a better name for this parameter, also consider moving the method outside. There are some advantages of using nested methods like sharing context but this method seems pretty independent as it is. I don't see a reason for having a nested declaration in this case.

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -159,36 +234,53 @@ private async Task<bool> UseGoCliToScanAsync(string location, ISingleFileCompone
...
+        // There can be multiple require( ) sections in go 1.17+. loop over all of them.
+        while (!reader.EndOfStream)
         {
-            if (this.TryToCreateGoComponentFromModLine(line, out var goComponent))
+            var line = await reader.ReadLineAsync();
+
+            while (line != null && !line.StartsWith("require ("))
             {
-                singleFileComponentRecorder.RegisterUsage(new DetectedComponent(goComponent));
+                if (line.StartsWith("go "))
```
jcfiorenzano, line 265: nit: what if instead of having `go ` and later `line[3..]` you declare a string constant with a descriptive name and then you can do something like `line.StartsWith(MyNewDescriptiveConst)` and in the following line you do `line[MyNewDescriptiveCons.length]`.

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -159,36 +234,53 @@ private async Task<bool> UseGoCliToScanAsync(string location, ISingleFileCompone
...
+                {
+                    goGraphTelemetryRecord.GoModVersion = line[3..].Trim();
+                }
+
+                // In go >= 1.17, direct dependencies are listed as "require x/y v1.2.3", and transitive dependencies
+                // are listed in the require () section
+                if (line.StartsWith("require "))
```
jcfiorenzano, line 272: same here

jcfiorenzano (approved):

src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
melotic: There are so many different private methods for all the use cases the go detector has to support that I think it's better to follow up with a separate PR to refactor this into different static helper classes.

PR #753, opened 2023-08-29
refactor(go): split `go.mod` and `go.sum` parsing

melotic:
Splits the parsing for `go.mod` files and `go.sum` files into two seperate helper classes.

This helps to declutter the go detector, which has 3 jobs to do. Now the go detector's main code is for go cli scanning, the happy path.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
@@ -6,7 +6,6 @@ using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using System.Reactive.Linq;
-using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.ComponentDetection.Common;
 using Microsoft.ComponentDetection.Common.Telemetry.Records;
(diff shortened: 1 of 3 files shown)


src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -78,7 +73,7 @@ public class GoComponentDetector : FileComponentDetector
             var goModFileContents = reader.ReadToEnd();
             goModFile.Stream.Dispose();
 
-            return this.CheckGoModVersion(goModFileContents, processRequest, goModFile);
+            return GoModFileParser.ShouldParseGoSumFile(goModFileContents, processRequest, goModFile, this.Logger);
```
jcfiorenzano, line 76: it seems weird that having a `GoSumFileParser` is the `GoModFileParser` that one that tells you if you have to parse the `go.sum` file or not. There are different ways that you can handle that one is that you have a method in the detector class that is the one that knows when and how the different parser should interact. Then the `GoModFileParser` will have a method to output the version. The coordinator method then is the one that contains the business logic to determine if the `go.sum` should be parsed or not. In that way you keep the `GoModFileParser` concise.

jcfiorenzano (approved):

FernandoRojo:
Closing as the underlying go detector was overhauled and replaced by the GoWithReplace detector, https://github.com/microsoft/component-detection/pull/1272/files

I will be revisiting the go detector to fix some other issues, and can introduce some of these refactor changes when performing that update.

PR #760, opened 2023-08-31
ci(build): build on windows and linux

melotic:
We have some tests that are not run on windows and vice verse. We should be running those tests in CI.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,11 @@ permissions:
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


melotic:
Cannot add MacOS support, as it doesn't include docker, which is needed for some tests: https://github.com/actions/runner-images/issues/17#issuecomment-614726536

Could add a skip tests on MacOS attribute, but I think this is good enough.

cobya (approved):

PR #776, opened 2023-09-11
feat(rust): add `cargo metadata` detector

### LINKED ISSUE #625: Use `cargo metadata` to properly detect Rust dev/build components
Our current Rust detector works by parsing the `Cargo.lock` file generated by Cargo. However, this lockfile does not include information on whether the component is a development dependency, or a build dependency.

Instead, we should use the command [`cargo metadata`](https://doc.rust-lang.org/cargo/commands/cargo-metadata.html) if `cargo` is available.

`cargo metadata` will output the fully resolved tree of a given _workspace_, in JSON, and has a `kind` field that determines whether the component is a dev dependency, build dependency, or a normal dependency.

In addition, there is also a `--format-version` option, that will prevent breakage if cargo updates the metadata format.

Looking at the [commit logs](https://github.com/rust-lang/cargo/commit/d931a95b82076a451b34f924b372e858d16f7af3#diff-ba01db28a0383baeba4f892d93ee0b9f81a8a0e3892085d2634d2c5e54a1e056), it looks like this command has been in cargo since 2018.

AB#2103260

melotic:
Closes #625


src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -0,0 +1,134 @@
...
+
+            if (!graph.TryGetValue(id, out var node))
+            {
+                this.Logger.LogWarning("Could not find {Id} at {Location} in cargo metadata output", id, location);
+                return;
+            }
+
+            foreach (var dep in node.Deps)
+            {
+                this.TraverseAndRecordComponents(recorder, location, graph, dep.Pkg, detectedComponent, dep, parent == null);
```
melotic, line 125: Not a huge fan of checking for explicit dependencies like this. feedback appreciated

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -0,0 +1,134 @@
...
+            }
+
+            var cliResult = await this.cliService.ExecuteCommandAsync(
+                "cargo",
+                null,
+                "metadata",
+                "--manifest-path",
+                componentStream.Location,
+                "--format-version=1",
+                "--locked");
```
melotic, line 74: Prevents us from updating the `Cargo.lock`: https://doc.rust-lang.org/cargo/commands/cargo-install.html#dealing-with-the-lockfile

annaowens: I suppose there is a tradeoff here that the command will fail if the lockfile needs to be updated. 

Our similar use of `-mod=readonly` in the go detector causes failures fairly often in the go detector.

melotic: We should probably add a log here printing stderr here if we fail then.

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        try
+        {
+            if (!await this.cliService.CanCommandBeLocatedAsync("cargo", null))
+            {
+                this.Logger.LogWarning("Could not locate cargo command. Skipping Rust CLI detection");
+                return;
+            }
+
+            var cliResult = await this.cliService.ExecuteCommandAsync(
```
annaowens, line 67: This can fail. Should be handled / could kick off the existing parse-based detector if it does.

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
annaowens: Agreed

annaowens: From https://doc.rust-lang.org/cargo/commands/cargo-metadata.html#manifest-options:
>Either of these flags requires that the Cargo.lock file is up-to-date. If the lock file is missing, or it needs to be updated, Cargo will exit with an error. The --frozen flag also prevents Cargo from attempting to access the network to determine if it is out-of-date.

src/Microsoft.ComponentDetection.Detectors/rust/Contracts/CargoMetadata.cs
```diff
@@ -0,0 +1,603 @@
+// <auto-generated />
```
JamieMagee, line 1: Generated by what?

Also nit, but I generally prefer single class per file. But I can live with it if it's autogenerated.

melotic: Its autogenerated from the output of `cargo metadata` from https://app.quicktype.io/?l=csharp

melotic: Granted it did not get everything right so I had to make some changes.

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -0,0 +1,134 @@
...
+                "--format-version=1",
+                "--locked");
+
+            var metadata = CargoMetadata.FromJson(cliResult.StdOut);
+            var graph = BuildGraph(metadata);
+            var root = metadata.Resolve.Root;
+
+            this.TraverseAndRecordComponents(processRequest.SingleFileComponentRecorder, componentStream.Location, graph, root, null, null);
+        }
+        catch (Exception e)
```
JamieMagee, line 82: nit: ne more specific about the exceptions. It's good to have a catch all `Exception` block, but `InvalidOperationException` is thrown if the command can't be located. There are probably some others.

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -0,0 +1,134 @@
...
+            this.Logger.LogWarning(e, "Failed to parse Cargo.toml file: {Location}", processRequest.ComponentStream.Location);
+        }
+    }
+
+    private static Dictionary<string, Node> BuildGraph(CargoMetadata cargoMetadata) => cargoMetadata.Resolve.Nodes.ToDictionary(x => x.Id);
+
+    private static (string Name, string Version) ParseNameAndVersion(string nameAndVersion)
+    {
+        var parts = nameAndVersion.Split(' ');
+        return (parts[0], parts[1]);
```
JamieMagee, line 93: Check how many parts we have here? And throw or log a warning if it's not exactly 2.

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
JamieMagee: I can't think of any other way than `parent == null`. The roots of the graph always require special handling. If I was being nit-picky I would ask you to extract it to a variable to make it explicit `var isDirectDependency = parent is null;`.

JamieMagee (approved):
Couple of nit-picky comments, but otherwise looks great!

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
melotic: The cargo format mandates that the format is `<crate-name> <version> (<registry-type>+<registry-url>)`

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
melotic: We dont have that functionality untill we merge the detectors.

PR #790, opened 2023-09-18
feat: new filesystem walker

melotic:
Removes observables and gives each file detector their own walker, dramatically improving speeds for large codebases, at the cost of a bit more memory.

Once the code is ready, I'll run benchmarks including memory usage to test.


JamieMagee:
Does this allow us to drop our dependency on `Dotnet.Glob`?

src/Microsoft.ComponentDetection.Common/DirectoryWalkerFactory.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        var actionBlock = new ActionBlock<ProcessRequest>(callback, new ExecutionDataflowBlockOptions() { MaxDegreeOfParallelism = Environment.ProcessorCount });
+
+        var walker = new FileSystemEnumerable<FileSystemInfo>(
+            root.FullName,
+            (ref FileSystemEntry entry) => entry.ToFileSystemInfo(),
+            new EnumerationOptions
+            {
+                AttributesToSkip = FileAttributes.Hidden | FileAttributes.System,
```
JamieMagee, line 52: I think we might actually need to scan hidden directories.

src/Microsoft.ComponentDetection.Common/DirectoryWalkerFactory.cs
```diff
@@ -0,0 +1,128 @@
...
+    private readonly IPathUtilityService pathUtilityService;
+    private readonly ILogger<DirectoryWalkerFactory> logger;
+
+    public DirectoryWalkerFactory(IPathUtilityService pathUtilityService, ILogger<DirectoryWalkerFactory> logger)
+    {
+        this.logger = logger;
+        this.pathUtilityService = pathUtilityService;
+    }
+
+    public async Task WalkDirectoryAsync(DirectoryInfo root, ExcludeDirectoryPredicate directoryExclusionPredicate, IComponentRecorder recorder, Func<ProcessRequest, Task> callback, IEnumerable<string> filePatterns)
```
JamieMagee, line 26: nit: XML-doc

src/Microsoft.ComponentDetection.Common/DirectoryWalkerFactory.cs
melotic: We currently do not. The original walker does not set these options, and `FileAttributes.Hidden | FileAttributes.System` is [the default](https://learn.microsoft.com/en-us/dotnet/api/system.io.enumerationoptions.attributestoskip?view=net-7.0#property-value).

https://github.com/microsoft/component-detection/blob/1c3ef853d24315f8648ec3b0b38f776b927af0f4/src/Microsoft.ComponentDetection.Common/FastDirectoryWalkerFactory.cs#L93-L98

melotic:
A new approach is needed, it seems this is dramatically slower:

image

Issue #798, opened 2023-09-20
Revert changes related to individual spdx file.


tarun06:
As per recent discussion with @sebasgomez238 on sbom-tool Repository, It is confirmed that We don't want to display information from within the SPDX file. SBOMs are representative of an individual piece of software. We just list the presence of an SPDX file in the externalDocumentRefs section. No additional details are required. Therefore, spdx related changes should be removed. 

Reference PR -> https://github.com/microsoft/sbom-tool/pull/319

AB#2106159

sebasgomez238:
@tarun06 component-detection is a separate product. No changes are necessary on their end as these features may be useful to someone else we just don’t use them in the sbom-tool. This issue can be closed.

tarun06:
Thanks, Closing it

PR #799, opened 2023-09-20
Revert changes related to spdx package component

tarun06:
Issue number #798 

As per recent discussion with @sebasgomez238 on sbom-tool Repository, It is confirmed that We don't want to display information from within the SPDX file. SBOMs are representative of an individual piece of software. We just list the presence of an SPDX file in the externalDocumentRefs section. No additional details are required. Therefore, spdx related changes should be removed.

Reference PR -> https://github.com/microsoft/sbom-tool/pull/319


tarun06:
Closing it base on comment on the issue

PR #805, opened 2023-09-25
fix: validate `--SourceDirectory`

melotic:


src/Microsoft.ComponentDetection.Orchestrator/Commands/ScanSettings.cs
```diff
@@ -60,4 +61,15 @@ public class ScanSettings : BaseSettings
         "Comma separated list of docker image names or hashes to execute container scanning on, ex: ubuntu:16.04, 56bab49eef2ef07505f6a1b0d5bd3a601dfc3c76ad4460f24c91d6fa298369ab")]
     [TypeConverter(typeof(CommaDelimitedConverter))]
     public IEnumerable<string> DockerImagesToScan { get; set; }
+
+    /// <inheritdoc />
+    public override ValidationResult Validate()
+    {
+        if (this.SourceDirectory is null)
```
grvillic, line 68: Should we also validate this isn't whitespace?

grvillic (approved):

src/Microsoft.ComponentDetection.Orchestrator/Commands/ScanSettings.cs
melotic: This should be caught by exists no?

melotic: I also believe that `Spectre.Console.Cli` will validate this

PR #808, opened 2023-09-26
fix: validate `--Output`

melotic:
This used to be validated in the Orchestrator


annaowens (approved):

src/Microsoft.ComponentDetection.Orchestrator/Commands/BaseSettings.cs
```diff
@@ -44,6 +45,11 @@ public override ValidationResult Validate()
             return ValidationResult.Error($"{nameof(this.Timeout)} must be a positive integer");
         }
 
+        if (!string.IsNullOrEmpty(this.Output) && !Directory.Exists(this.Output))
+        {
+            return ValidationResult.Error($"{nameof(this.Output)} must be a valid path");
```
JamieMagee, line 50: The description probably needs to be updated

> Output path for log files. Defaults to %TMP%

PR #814, opened 2023-09-28
Fix to reduce pip log verbosity

Omotola:
Fix to reduce pip log verbosity
image


src/Microsoft.ComponentDetection.Orchestrator/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -141,6 +143,7 @@ public static IServiceCollection AddComponentDetection(this IServiceCollection s
 
         // HttpClient
         services.AddHttpClient();
+        services.RemoveAll<IHttpMessageHandlerBuilderFilter>();
```
melotic, line 146: can we add a comment about why this is needed?

melotic (approved):

PR #816, opened 2023-09-28
Fix to hide Http MessageHandler cleanup messages

Omotola:
Based on these threads [Stack Overflow](https://stackoverflow.com/questions/63199414/asp-net-core-3-1-httpmessagehandler-cleanup-cycle), [Github](https://github.com/aspnet/HttpClientFactory/issues/165), It seems the multiple cleanup messages are default behavior as it waits to clean up the message handler resources. The suggested thing to do is hide the messages in the logs if we don't want to see them.

### DIFF
--- a/src/Microsoft.ComponentDetection/Program.cs
+++ b/src/Microsoft.ComponentDetection/Program.cs
@@ -25,6 +25,7 @@ var serviceCollection = new ServiceCollection()
     .AddComponentDetection()
     .AddLogging(l => l.AddSerilog(new LoggerConfiguration()
         .MinimumLevel.ControlledBy(Interceptor.LogLevel)
+        .MinimumLevel.Override("Microsoft.Extensions.Http.DefaultHttpClientFactory", LogEventLevel.Information)
         .Enrich.With<LoggingEnricher>()
         .Enrich.FromLogContext()
         .WriteTo.Map(


melotic:
> This means that there's a client/message handler that's been expired, but someone is still holding a reference to it. You can analyze what's keeping it alive by capturing a memory dump.

Do we dispose all instances of the `HTTPClient`? Why does the old pip detector work fine but this not?

I'm not sure this is the right solution but instead just masking the logs.

Omotola:
> Do we dispose all instances of the `HTTPClient`? Why does the old pip detector work fine but this not?

@melotic Yeah, the httpclient is only created using httpclientfactory in the SimplePypiClient, and its disposed after. The original pip detector doesn't use httpclientfactory, it manually creates the httphandler and httpclient. Jamie had mentioned that doing it that way would result in resources not being cleaned up properly afterwards. He suggested using httpclientfactory for the simplepipdetector cause it uses dependency injection to inject the httpclient, and handles the cleanup by itself.

Also if you scroll down to the end of that [github issue](https://github.com/aspnet/HttpClientFactory/issues/165) you quoted, you'll see where they concluded that cleanup message loop is by design. I also found another article [here](https://ofstack.com/ASP.NET/43070/detailed-explanation-of-httpclientfactory-class-for-in-depth-understanding-of-dotnetcore.html) that explains that by design httpclientfactory expires handlers after 2 minutes of no longer being in use and adds them to an expired queue, and then there is a cleanup job that checks that queue and has a timer that checks every 10 seconds if the handler has been garbage collected so it can cleanup other internal resources (i.e socket connections). 
In the logs we see in our builds you will notice there is the expiration message of the one handler 
image
and then the consequent cleanup loop at 10 second intervals which is just waiting for the handler to be GC
image

melotic (approved):
this will need to replaced in internal

PR #820, opened 2023-09-28
PR to populate disable detectors list

Omotola:
Readded changes to populate disabled detectors list and added some unit tests. Also modified `ScanResult` contract to help with the unit tests.

### DIFF
--- a/src/Microsoft.ComponentDetection.Orchestrator/Services/ScanExecutionService.cs
+++ b/src/Microsoft.ComponentDetection.Orchestrator/Services/ScanExecutionService.cs
@@ -38,14 +38,17 @@ public class ScanExecutionService : IScanExecutionService
         using var scope = this.logger.BeginScope("Executing BCDE scan");
 
         var detectorRestrictions = this.GetDetectorRestrictions(settings);
-        var detectors = this.detectorRestrictionService.ApplyRestrictions(detectorRestrictions, this.detectors).ToImmutableList();
+        var detectors = this.detectors.ToList();
+        var restrictedDetectors = this.detectorRestrictionService.ApplyRestrictions(detectorRestrictions, detectors).ToImmutableList();
+        detectorRestrictions.DisabledDetectors = detectors.Except(restrictedDetectors).ToList();
 
         this.logger.LogDebug("Finished applying restrictions to detectors.");
 
-        var processingResult = await this.detectorProcessingService.ProcessDetectorsAsync(settings, detectors, detectorRestrictions);
+        var processingResult = await this.detectorProcessingService.ProcessDetectorsAsync(settings, restrictedDetectors, detectorRestrictions);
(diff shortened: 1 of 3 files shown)


src/Microsoft.ComponentDetection.Orchestrator/Services/ScanExecutionService.cs
```diff
@@ -38,14 +38,17 @@ public async Task<ScanResult> ExecuteScanAsync(ScanSettings settings)
         using var scope = this.logger.BeginScope("Executing BCDE scan");
 
         var detectorRestrictions = this.GetDetectorRestrictions(settings);
-        var detectors = this.detectorRestrictionService.ApplyRestrictions(detectorRestrictions, this.detectors).ToImmutableList();
+        var detectors = this.detectors.ToList();
```
Omotola, line 41: Added this line to convert to a list cause the tests were failing with IEnumerable

src/Microsoft.ComponentDetection.Orchestrator/Services/ScanExecutionService.cs
```diff
@@ -38,14 +38,17 @@ public async Task<ScanResult> ExecuteScanAsync(ScanSettings settings)
...
+        var restrictedDetectors = this.detectorRestrictionService.ApplyRestrictions(detectorRestrictions, detectors).ToImmutableList();
+        detectorRestrictions.DisabledDetectors = detectors.Except(restrictedDetectors).ToList();
 
         this.logger.LogDebug("Finished applying restrictions to detectors.");
 
-        var processingResult = await this.detectorProcessingService.ProcessDetectorsAsync(settings, detectors, detectorRestrictions);
+        var processingResult = await this.detectorProcessingService.ProcessDetectorsAsync(settings, restrictedDetectors, detectorRestrictions);
```
grvillic, line 47: nit: This naming seems a bit confusing. We need to process all the detectors. This is just a subset of detectors, but naming it like this + the `detectorRestrictions` next to it makes it confusing as to what are we actually passing.

src/Microsoft.ComponentDetection.Orchestrator/Services/ScanExecutionService.cs
grvillic: nit: but I think the tests were not mocked properly, we should have to enlist the entries to make then work. Just your check may need to be looking for equivalent objects not the same exact object.

grvillic (approved):

src/Microsoft.ComponentDetection.Orchestrator/Services/ScanExecutionService.cs
Omotola: The initial one that fails is actually only checking if `DetectorsNotInRun` is not empty image

PR #855, opened 2023-10-13
Add Properties VCPKG

amitla1:
Added GitRepositoryOwner and GitRepositoryName to prevent fuzzy matching

### DIFF
--- a/src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
+++ b/src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
@@ -1,5 +1,6 @@
 namespace Microsoft.ComponentDetection.Contracts.TypedComponent;
 
+using System.Linq;
 using PackageUrl;
 
 public class VcpkgComponent : TypedComponent
@@ -20,9 +21,11 @@ public class VcpkgComponent : TypedComponent
         this.Triplet = triplet;
         this.Description = description;
         this.DownloadLocation = downloadLocation;
-        var locationArr = downloadLocation.Split('/');
-        this.GitRepositoryOwner = locationArr[3];
-        this.GitRepositoryName = locationArr[4].Split('@')[0];
+
+        if (downloadLocation.ToLower().Contains("https://github.com/"))
+        {
+            this.SetGitRepoProperties();
+        }
     }
 
     public string SPDXID { get; set; }
@@ -30,10 +33,6 @@ public class VcpkgComponent : TypedComponent
     public string Name { get; set; }
 
     public string DownloadLocation { get; set; }
-
-    public string GitRepositoryOwner { get; set; }
-
-    public string GitRepositoryName { get; set; }
 
     public string Triplet { get; set; }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
```diff
@@ -20,6 +20,9 @@ public VcpkgComponent(string spdxid, string name, string version, string triplet
         this.Triplet = triplet;
         this.Description = description;
         this.DownloadLocation = downloadLocation;
+        var locationArr = downloadLocation.Split('/');
+        this.GitRepositoryOwner = locationArr[3];
```
chsalgado, line 24: Magic numbers. Please add a comment showing a sample DownloadLocation

chsalgado (approved):

src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
```diff
@@ -71,4 +81,22 @@ public override PackageURL PackageUrl
...
+        */
+        var locationArr = this.DownloadLocation.Split('/');
+        if (!string.IsNullOrEmpty(locationArr[2]))
+        {
+            this.GitRepositoryOwner = locationArr[2];
+        }
+
+        if (!string.IsNullOrEmpty(locationArr[3]))
+        {
+            this.GitRepositoryName = locationArr[3].TakeWhile(ch => char.IsLetterOrDigit(ch)).ToString();
```
grvillic, line 99: I think this check may fail for repositories with valid names like `component-detection`

src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
```diff
@@ -71,4 +81,22 @@ public override PackageURL PackageUrl
...
     }
+
+    private void SetGitRepoProperties()
+    {
+        /* example download locations
+         * "git+https://github.com/leethomason/tinyxml2@9.0.0"
+         * "git+https://github.com/Microsoft/vcpkg#ports/nlohmann-json"
+        */
+        var locationArr = this.DownloadLocation.Split('/');
+        if (!string.IsNullOrEmpty(locationArr[2]))
```
grvillic, line 92: Let's do an array length check, in which we expect at least 4 fragments, if less than that this should be noop. Otherwise we can hit index out of range exceptions here.

src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
```diff
@@ -71,4 +81,22 @@ public override PackageURL PackageUrl
...
         }
     }
+
+    private void SetGitRepoProperties()
+    {
+        /* example download locations
+         * "git+https://github.com/leethomason/tinyxml2@9.0.0"
+         * "git+https://github.com/Microsoft/vcpkg#ports/nlohmann-json"
+        */
+        var locationArr = this.DownloadLocation.Split('/');
```
grvillic, line 91: qq, if downloadlocation is null will the split work?

src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
```diff
@@ -1,5 +1,6 @@
-﻿namespace Microsoft.ComponentDetection.Contracts.TypedComponent;
+namespace Microsoft.ComponentDetection.Contracts.TypedComponent;
 
+using System.Linq;
 using PackageUrl;
 
 public class VcpkgComponent : TypedComponent
```
grvillic, line 6: I expect to see some tests to validate these new changes.

src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
amitla1: I have a null or empty check before this method gets called

PR #864, opened 2023-10-17
VCPKG Properites

amitla1:
adding more checks for the new vcpkg properties.

### DIFF
--- a/src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
+++ b/src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
@@ -1,5 +1,6 @@
-﻿namespace Microsoft.ComponentDetection.Contracts.TypedComponent;
+namespace Microsoft.ComponentDetection.Contracts.TypedComponent;
 
+using System.Linq;
 using PackageUrl;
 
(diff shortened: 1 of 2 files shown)


grvillic (changes requested):
This is blocked on proper contract changes for Governance and CD public. Those NuGet packages need to be updated here in manifest for this PR to succeed.

src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
```diff
@@ -88,15 +88,17 @@ private void SetGitRepoProperties()
...
+        if (Uri.TryCreate(this.DownloadLocation, UriKind.Absolute, out var tempUri))
         {
-            this.GitRepositoryName = locationArr[3].TakeWhile(ch => char.IsLetterOrDigit(ch)).ToString();
+            var locationArr = tempUri.Segments;
+            if (locationArr.Length < 3)
+            {
+                return;
+            }
+
+            this.GitRepositoryOwner = locationArr[1].TrimEnd('/');
```
grvillic, line 100: These properties are not first class for Vcpkg, downstream clients should capture this themselves if desirable. This data is encoded in native DownloadLocation.

PR #872, opened 2023-10-20
Users/oakeredolu/throttlehttpclient

Omotola:
This pr uses semaphore to reduce the amount of http requests to 5 concurrent requests at a time in both pip and simplepip. This is to avoid httpclient timing out due to too many requests at a time. 

The fix also seems to have made the cd scan run faster. It was initially taking over 30 minutes to run on the vienna repo locally, but after throttling the requests the scan is taking about 23 minutes.

I also remove the manual disposal of httpclient, since its [not recommended](https://stackoverflow.com/questions/65363252/how-to-dispose-httpclient-after-its-finished) to manually dispose it.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/pip/SimplePypiClient.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/pip/SimplePypiClient.cs
@@ -65,7 +65,7 @@ public sealed class SimplePyPiClient : ISimplePyPiClient, IDisposable
     {
         this.environmentVariableService = environmentVariableService;
         this.logger = logger;
-        this.semaphore = new SemaphoreSlim(10, 10);
+        this.semaphore = new SemaphoreSlim(5);
     }
 
     public static HttpClient HttpClient { get; internal set; } = new HttpClient(HttpClientHandler);
@@ -287,7 +287,6 @@ public sealed class SimplePyPiClient : ISimplePyPiClient, IDisposable
         this.cacheTelemetry.Dispose();
         this.cachedProjectWheelFiles.Dispose();
         this.cachedSimplePyPiProjects.Dispose();
-        HttpClient.Dispose();
         this.semaphore.Dispose();
     }
 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.ComponentDetection.Detectors/pip/SimplePypiClient.cs
```diff
@@ -265,12 +269,14 @@ private async Task<HttpResponseMessage> RetryPypiRequestAsync(Uri uri, PipDepend
...
     private async Task<HttpResponseMessage> GetPypiResponseAsync(Uri uri)
     {
+        await this.semaphore.WaitAsync();
         this.logger.LogInformation("Getting Python data from {Uri}", uri);
         using var request = new HttpRequestMessage(HttpMethod.Get, uri);
         request.Headers.UserAgent.Add(ProductValue);
         request.Headers.UserAgent.Add(CommentValue);
         request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/vnd.pypi.simple.v1+json"));
         var response = await HttpClient.SendAsync(request);
+        this.semaphore.Release();
```
cobya, line 279: We should do this within a try-catch-finally block which throws the exception again but releases the semaphore in the finally block so exceptions don't block other threads forever.

src/Microsoft.ComponentDetection.Detectors/pip/SimplePypiClient.cs
```diff
@@ -265,12 +269,14 @@ private async Task<HttpResponseMessage> RetryPypiRequestAsync(Uri uri, PipDepend
     /// <returns> Returns the httpresponsemessage. </returns>
     private async Task<HttpResponseMessage> GetPypiResponseAsync(Uri uri)
     {
+        await this.semaphore.WaitAsync();
```
cobya, line 272: What kind of impact does this have on build times for large Python repositories?

PR #878, opened 2023-10-26
Support development dependencies for the Gradle detector

joakley-msft:
Lack of development dependency detection for Gradle is a problem for Android teams, especially in the context of Component Governance alerts. Unfortunately Gradle doesn't provide enough information to definitively identify dev dependencies in all cases, so manual configuration is required. This change adds dev dependency classification through two mechanisms

1. `buildscript-gradle.lockfile` and `settings-gradle.lockfile` contain only build-system dependencies, so always classify these as development dependencies.
2. Processing based on two new environment variables: `GRADLE_PROD_CONFIGURATIONS_REGEX` and `GRADLE_DEV_CONFIGURATIONS_REGEX`. Gradle lockfiles indicate which Gradle configuration(s) each dependency is required by. `GRADLE_PROD_CONFIGURATIONS_REGEX` allows specifying production configurations explicitly. All other configurations are considered development. Alternately, dev configurations may be specified in `GRADLE_DEV_CONFIGURATIONS_REGEX` and all others are considered production.


rygo-msft:
CG is auto-injected into all production pipelines, so I fear that maintaining an accurate configuration list in an environment variable will be near-impossible. CG operates on lockfiles as checked-in files, so I think ultimately the config needs to be a checked-in file as well.

src/Microsoft.ComponentDetection.Detectors/gradle/GradleComponentDetector.cs
```diff
@@ -47,6 +64,9 @@ protected override Task OnFileFoundAsync(ProcessRequest processRequest, IDiction
 
     private void ParseLockfile(ISingleFileComponentRecorder singleFileComponentRecorder, IComponentStream file)
     {
+        // Buildscript and Settings lockfiles are always development dependencies
+        var lockfileIsDevDependency = file.Location.EndsWith("buildscript-gradle.lockfile") || file.Location.EndsWith("settings-gradle.lockfile");
```
rygo-msft, line 68: lockfile paths are configurable. The buildscript-gradle.lockfile in particular is not a default lock path.

docs/environment-variables.md
```diff
@@ -18,4 +18,16 @@ When set to any value, enables detector experiments, a feature to compare the re
 same ecosystem. The available experiments are found in the [`Experiments\Config`](../src/Microsoft.ComponentDetection.Orchestrator/Experiments/Configs)
 folder.
 
+## `GRADLE_PROD_CONFIGURATIONS_REGEX`
+
+Enables dev-dependency categorization for the Gradle
+detector. Dependencies connected to Gradle configurations NOT matching
```
rygo-msft, line 24: This isn't quite sufficient level of granularity. A specific configuration wont be production across the board for all lockfiles in a project. Likely only one configuration in a handful of lockfiles is production and the rest are dev dependencies.

A java project will use the `classpath` configuration for example, but that would also apply to the buildscript runtime configuration.

And android project might have `releaseClasspath` as its production configuration, but a test app in the same repo would also have a `releaseClasspath` configuration.

rygo-msft: It would require a bit more work on the consumer's end, but if we were to add configuration for tagging specific lockfiles as production vs dev, we can setup production configurations to be written to a separate lockfile from dev dependencies.

src/Microsoft.ComponentDetection.Detectors/gradle/GradleComponentDetector.cs
joakley-msft: Unless I'm misreading something, buildscript-gradle.lockfile *is* a standard name, it is documented here https://docs.gradle.org/current/userguide/dependency_locking.html#lock_state_location_and_format

joakley-msft:
> CG is auto-injected into all production pipelines, so I fear that maintaining an accurate configuration list in an environment variable will be near-impossible.

I'm aware, but not sure why we couldn't set the variable in all production pipelines too?

> CG operates on lockfiles as checked-in files, so I think ultimately the config needs to be a checked-in file as well.

I'd like input from the CG/component-detection owners before proceeding further here. There are existing configurations for specific detectors with environment variables, and I don't see any precedent for file-based configuration in this tool. I'm happy to go either direction based on guidance/preference of the owners here.

docs/environment-variables.md
joakley-msft: That's fair. I can expand the configuration to allow specification of matching for prod or dev lockfiles, but will wait to hear from the repo owners to resolve the env-var vs configuration-file question from the other comment thread.

docs/environment-variables.md
```diff
@@ -18,4 +18,16 @@ When set to any value, enables detector experiments, a feature to compare the re
 same ecosystem. The available experiments are found in the [`Experiments\Config`](../src/Microsoft.ComponentDetection.Orchestrator/Experiments/Configs)
 folder.
 
+## `GRADLE_PROD_CONFIGURATIONS_REGEX`
```
melotic, line 21: nit: could we prefix these with `CD`?

src/Microsoft.ComponentDetection.Detectors/gradle/GradleComponentDetector.cs
```diff
@@ -87,4 +108,38 @@ private MavenComponent CreateMavenComponentFromFileLine(string line)
...
+        if (this.prodDepsRegex == null && this.devDepsRegex == null)
+        {
+            return false; // no regexes configured to check against
+        }
+
+        var equalsSeparatorIndex = line.IndexOf('=');
+        if (equalsSeparatorIndex == -1)
+        {
+            // We can't parse out the configuration. Maybe the project is using the one-lockfile-per-configuration format but
+            // this is deprecated in Gradle so we don't support it here, projects should upgrade to one-lockfile-per-project.
```
melotic, line 123: can we log a warning here for customers with a recommendation?

src/Microsoft.ComponentDetection.Detectors/gradle/GradleComponentDetector.cs
```diff
@@ -22,6 +28,17 @@ public class GradleComponentDetector : FileComponentDetector, IComponentDetector
         this.ComponentStreamEnumerableFactory = componentStreamEnumerableFactory;
         this.Scanner = walkerFactory;
         this.Logger = logger;
+        var envVar = Environment.GetEnvironmentVariable(ProdDepsEnvVar);
```
melotic, line 31: we have a `IEnvironmentVariableService` to allow mocking and testing of env variables

melotic: lets also update the tests to not set environment variables as well

joakley-msft:
Thanks for the review @melotic. Before I make changes to address your comments on the environment variable access and naming, could you or @JamieMagee  weigh in on Ryan's suggestion to use a configuration file instead of environment variables? I'm happy to go with either approach. I agree with @rygo-msft 's point on the benefit of a configuration file, but am not sure if it would be too inconsistent with the rest of the component-detection tool.

src/Microsoft.ComponentDetection.Detectors/gradle/GradleComponentDetector.cs
joakley-msft: I considered this, but for any projects that are still using one-lockfile-per-configuration, the warning would be horribly verbose as this function is executed a lot.

rygo-msft:
I'm still unclear how this list of developer configurations will be built and maintained in practice. A typical android project will have 20+ different configurations and that number grows even large when build flavors are used.

Because lockfiles are static files, they will be scanned even for pipelines unrelated to the gradle build. These config values will need to be copied and initialized in every pipeline somehow. This is particularly difficult to achieve as static analysis jobs are typically injected without pipeline author input. There are cases where pipeline authors cannot provide configuration these injected jobs.

joakley-msft:
I would still like to hear from the repo maintainers if they would accept a change that read from a configuration file instead of environment variables. I agree that would be somewhat preferable in this case.

Ryan, can you give an example of a case where the pipeline authors *cannot* provide configuration via environment variables though? I was assuming we would just run the task setting up the environment variables before the component detection task.

rygo-msft:
Governed Templates inject static analysis jobs that run standardized task sets and do not take user provided tasks.
Even in the case where this is an option, there are often many pipelines running from the same repo. Most of those will not run a gradle build at all, but need to be able to configure CG correctly because the analysis is on the static files.

joakley-msft:
We don't need to run a gradle build, just to be able to set up environment variables. But I wasn't familiar with the auto-injected static analysis jobs, I agree that's a concern and a good motivation for allowing configuration by file instead of environment variable.  I'll pause here to see if Anna, Jamie or others can weigh in.

src/Microsoft.ComponentDetection.Detectors/gradle/GradleComponentDetector.cs
```diff
@@ -45,8 +57,24 @@ protected override Task OnFileFoundAsync(ProcessRequest processRequest, IDiction
         return Task.CompletedTask;
     }
 
+    private List<string> ReadEnvVarStringList(IEnvironmentVariableService envVarService, string envVar)
```
cobya, line 60: We can probably move this to `EnvironmentVariableService` and allow configurable string delimiters as a parameter since it's fairly generic.

src/Microsoft.ComponentDetection.Detectors/gradle/GradleComponentDetector.cs
```diff
@@ -12,16 +13,27 @@ namespace Microsoft.ComponentDetection.Detectors.Gradle;
 
 public class GradleComponentDetector : FileComponentDetector, IComponentDetector
```
cobya, line 14: Please bump the detector version as it is determining different dependencies now that it contains dev dep info.

src/Microsoft.ComponentDetection.Detectors/gradle/GradleComponentDetector.cs
```diff
@@ -45,8 +57,24 @@ protected override Task OnFileFoundAsync(ProcessRequest processRequest, IDiction
         return Task.CompletedTask;
     }
 
+    private List<string> ReadEnvVarStringList(IEnvironmentVariableService envVarService, string envVar)
+        => (envVarService.GetEnvironmentVariable(envVar) ?? string.Empty).Split(",", StringSplitOptions.RemoveEmptyEntries).ToList();
+
     private void ParseLockfile(ISingleFileComponentRecorder singleFileComponentRecorder, IComponentStream file)
     {
+        // Buildscript and Settings lockfiles are always development dependencies
+        var lockfileRelativePath = Path.GetRelativePath(this.CurrentScanRequest.SourceDirectory.FullName, file.Location);
```
cobya, line 66: Let's move this logic to its own helper function.

src/Microsoft.ComponentDetection.Detectors/gradle/GradleComponentDetector.cs
```diff
@@ -45,8 +57,24 @@ protected override Task OnFileFoundAsync(ProcessRequest processRequest, IDiction
...
+        var lockfileIsDevDependency = lockfileRelativePath.EndsWith("buildscript-gradle.lockfile") || lockfileRelativePath.EndsWith("settings-gradle.lockfile") || this.devLockfiles.Contains(lockfileRelativePath);
```
cobya, line 67: You should just be able to compare the file names directly rather than `lockfileRelativePath.EndsWith` unless Gradle allows `abc-buildscript-gradle.lockfile` as valid names.

src/Microsoft.ComponentDetection.Detectors/gradle/GradleComponentDetector.cs
```diff
@@ -87,4 +116,23 @@ private MavenComponent CreateMavenComponentFromFileLine(string line)
...
+        var equalsSeparatorIndex = line.IndexOf('=');
+        if (equalsSeparatorIndex == -1)
+        {
+            // We can't parse out the configuration. Maybe the project is using the one-lockfile-per-configuration format but
+            // this is deprecated in Gradle so we don't support it here, projects should upgrade to one-lockfile-per-project.
+            return false;
+        }
+
+        var configurations = line[(equalsSeparatorIndex + 1)..].Split(",");
+        return configurations.Select(c => this.IsDevDependencyByConfigurationName(c)).All(dev => dev);
```
cobya, line 131: `.Select` is already returning a list of bools here. This can be simplified to `configurations.All(this.IsDevDependencyByConfigurationName)`

cobya (changes requested):

src/Microsoft.ComponentDetection.Detectors/gradle/GradleComponentDetector.cs
joakley-msft: Ah, I overlooked that detectors were versioned, thanks

cobya (approved):

PR #904, opened 2023-11-15
Add a parameter to control the summary display

### LINKED ISSUE #900: Parameter to control the summary display
In my case I have a custom application, which is using `Spectre.Console` and `component-detection` nugets to scan a folder. (so not running another executable, but directly controlling another `CommandApp<ScanCommand>` which is working as expected).

We have fine control over logging level but since you moved to `Spectre.Console` as well, `LogTabularOutput` is used to display a detection summary:

[    private void LogTabularOutput(ConcurrentDictionary<string, DetectorRunResult> providerElapsedTime, double totalElapsedTime)](https://github.com/microsoft/component-detection/blob/03a561ac9c9dd06170661c82e01859ed71c8a3af/src/Microsoft.ComponentDetection.Orchestrator/Services/DetectorProcessingService.cs#L303)

And rendering as is on the *default TextWriter (like Console.Out)* which is completely interfering with my application's output.

It is possible to hack a bit like using `AnsiConsole.AlternateScreen()` or redirecting `AnsiConsole.Profile.Out`, but that's not convenient or compatible with all use-cases (think of CI runs / redirected outputs).

So could we add a parameter to `ScanCommandSettings` like `NoSummary` to completely disable the rendering of this detection summary?

I can make a PR for that if you want.

Thanks!

AB#2123049

sailro:
Fixes #900

This PR just adds a parameter to `ScanCommandSettings`: `--NoSummary` to completely disable the rendering/logging of the detection summary.

This is useful when an application directly controls the execution of the `component-detection`. Displaying the detection summary directly on the standard output can interfere with the main application's own output.

The manifest already provides all the necessary information for the host application to display it's own status.

I did not add a test for this super simple feature. Because this is an implementation detail of the `DetectorProcessingService` that doesn't seem easily testable with the available mocks. 

Thanks!

### DIFF
--- a/src/Microsoft.ComponentDetection.Orchestrator/Commands/ScanSettings.cs
+++ b/src/Microsoft.ComponentDetection.Orchestrator/Commands/ScanSettings.cs
@@ -72,6 +72,10 @@ public class ScanSettings : BaseSettings
     [TypeConverter(typeof(CommaDelimitedConverter))]
     public IEnumerable<string> DockerImagesToScan { get; set; }
 
+    [CommandOption("--NoSummary")]
+    [Description("Do not display the detection summary on the standard output nor in the logs.")]
+    public bool NoSummary { get; set; }
+
     /// <inheritdoc />
     public override ValidationResult Validate()
     {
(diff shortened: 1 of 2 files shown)


melotic (approved):

annaowens:
@sailro Did you intentionally close this? Just following up here.

sailro:
@annaowens nope, sorry for this, I'm going to recreate another clone PR (I cannot change the origin branch as my initial repo was deleted).

sailro:
See https://github.com/microsoft/component-detection/pull/917

PR #915, opened 2023-11-27
Skip detection of workspace projects in Yarn detector

MLoughry:
Yarn v1 does not list local workspace projects in the `yarn.lock` file, but Yarn "berry" (>= v2) does. As a result, there is a difference in behavior in the component detection between the two versions. On very large mono-repo projects (such as Outlook's main client repository), this change combined with the very repetitive/uncompressed nature of the resulting scan report JSON can balloon to an unmanageable size (>80x increase in Outlook's case).

This change fixes code looking for "resolved" in Yarn berry lockfiles to look for "resolution" instead (the new entry name). Additionally, it simply skips detection for local workspace projects to maintain parity with v1 detection (these projects are still detected by the NPM detector).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,6 +61,6 @@ Analysis rulesets are defined in [analyzers.ruleset](analyzers.ruleset) and vali
 
 ### Testing
 
-L0s are defined in `MS.VS.Services.Governance.CD.*.L0.Tests`.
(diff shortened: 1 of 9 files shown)


grvillic (approved):

cobya:
@grvillic the diff is expected here since it's a local component correct?

PR #918, opened 2023-11-29
Update documentation for detector arguments

sailro:
Following changes in https://github.com/microsoft/component-detection/pull/917 for `--NoSummary`

I also documented `--PrintManifest` which was missing.

cc @annaowens


cobya (approved):
Thanks for adding this and the one we missed! 👍

annaowens (approved):
Thank you!

PR #931, opened 2023-12-08
Enable Experimental Rust and update telemetry

FernandoRojo:


src/Microsoft.ComponentDetection.Orchestrator/Services/DetectorProcessingService.cs
```diff
@@ -115,7 +112,7 @@ public class DetectorProcessingService : IDetectorProcessingService
                     exitCode = resultCode;
                 }
 
-                this.experimentService.RecordDetectorRun(detector, componentRecorder, settings);
+                this.experimentService.RecordDetectorRun(detector, componentRecorder, settings, providerElapsedTime[detector.Id + (isExperimentalDetector ? " (Beta)" : string.Empty)]);
```
grvillic, line 115: Can we have a utility function for this, it looks risky if we ever change the "(Beta)" suffix this will likely break it? On that same vein, we should check first if the detector id exists in Provider Elapsed time array, or this would break the whole scan.

FernandoRojo: updated to no longer use the array, but still pulled it out into a utility function for future references

PR #934, opened 2023-12-14
Add Author/License to LinuxComponent

sebasgomez238:
Adds Author and License properties from Syft output into the LinuxComponent.


src/Microsoft.ComponentDetection.Contracts/TypedComponent/LinuxComponent.cs
```diff
@@ -9,12 +9,14 @@ private LinuxComponent()
         /* Reserved for deserialization */
     }
 
-    public LinuxComponent(string distribution, string release, string name, string version)
+    public LinuxComponent(string distribution, string release, string name, string version, string license = null, string author = null)
```
JamieMagee, line 12: Can these be nullable? `string?`. Same for the properties below.

src/Microsoft.ComponentDetection.Detectors/linux/Contracts/SyftOutput.cs
```diff
@@ -371,6 +371,22 @@ public partial struct License
 
     public static implicit operator License(string String) => new License { String = String };
     public static implicit operator License(string[] StringArray) => new License { StringArray = StringArray };
+
+    // Override ToString method to provide a string representation of the struct
+    public override string ToString()
```
JamieMagee, line 376: This whole file is autogenerated. I would add a `SyftExtensions` file or class, and add an extension method for `License` instead.

sebasgomez238: Turns out this was actually unnecessary. I Removed the changes in this file.

src/Microsoft.ComponentDetection.Contracts/TypedComponent/LinuxComponent.cs
sebasgomez238: They can be but the analyzer kept complaining. This was the only way I could get it to not show errors. I didn't want to use "#nullable enable" for the entire file.

JamieMagee (approved):

cobya (approved):

PR #936, opened 2023-12-15
Adds an entry point for the sbom-tool use the ScanCommand class.

sebasgomez238:
Currently the ScanCommand class returns an int. In the sbom-tool we rely on the ScanResult object which forced us to go a step below and use the ScanExecution itself. This is causing us to miss out on a few features such as the ability to write a ScanManifest file (https://github.com/microsoft/sbom-tool/issues/452).

This PR adds an extra method to the ScanCommand class that returns a ScanResult object instead of an int.

Let me know if there is more suitable name the current name is "ExecuteAsync" the method I added is "ExecuteScanAsync"


cobya (approved):
Looks good to me, pending a look from Fern on why Rust is failing on the windows build I will merge.

sebasgomez238:
Thanks! @cobya I added unit tests since I saw the codecov comments.

JamieMagee (approved):

PR #938, opened 2023-12-19
Adds supplier/license info to pip components.

sebasgomez238:
This only covers the default PythonResolver. This information cannot be added to the SimplePythonResolver without making additional calls which would defeat the purpose of using the SimplePythonResolver.

Classifier field specification: https://peps.python.org/pep-0301/#distutils-trove-classification and https://packaging.python.org/en/latest/specifications/core-metadata/#classifier-multiple-use

Why we prioritize License field over Classifier License field: https://packaging.python.org/en/latest/specifications/core-metadata/#license


src/Microsoft.ComponentDetection.Contracts/TypedComponent/PipComponent.cs
```diff
@@ -10,16 +11,26 @@ private PipComponent()
...
+        this.Author = author;
+        this.License = license;
     }
 
     public string Name { get; set; }
 
     public string Version { get; set; }
 
+#nullable enable
+    [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
```
sebasgomez238, line 27: Because of the way the simplepip resolver handles building the dependency graph, it is not possible to find Author and License info for the dependencies of a package. This tag makes it so these properties aren't serialized when they are null. This avoids having fields in the graph that will always be null.

src/Microsoft.ComponentDetection.Detectors/pip/PythonResolver.cs
```diff
@@ -210,10 +214,57 @@ private void AddGraphNode(PythonResolverState state, PipGraphNode parent, string
...
+            var node = new PipGraphNode(new PipComponent(name, version, license: license, author: author));
             state.NodeReferences[name] = node;
             parent.Children.Add(node);
             node.Parents.Add(parent);
         }
     }
+
+    private string GetAuthorFromProject(PythonProject project)
+    {
+        if (!string.IsNullOrEmpty(project.Info?.Author))
```
grvillic, line 226: nit: We can use `string.IsNullOrWhitespace` on all of these entries

src/Microsoft.ComponentDetection.Detectors/pip/PythonResolver.cs
```diff
@@ -210,10 +214,57 @@ private void AddGraphNode(PythonResolverState state, PipGraphNode parent, string
...
+
+        if (!string.IsNullOrEmpty(project.Info?.AuthorEmail))
+        {
+            return project.Info.AuthorEmail;
+        }
+
+        if (!string.IsNullOrEmpty(project.Info?.MaintainerEmail))
+        {
+            return project.Info.MaintainerEmail;
+        }
```
grvillic, line 244: Why are we selecting this priority for project author? Is it documented anywhere? Some doc links in PR description would be helpful to get more context even for future reviewers.

src/Microsoft.ComponentDetection.Detectors/pip/PythonResolver.cs
```diff
@@ -210,10 +214,57 @@ private void AddGraphNode(PythonResolverState state, PipGraphNode parent, string
...
+
+        // If none of the fields are populated, return null.
+        return null;
+    }
+
+    private string GetLicenseFromProject(PythonProject project)
+    {
+        if (project.Info?.Classifiers != null)
+        {
+            var licenseClassifiers = project.Info.Classifiers.Where(x => x.StartsWith("License ::")).ToList();
```
grvillic, line 254: Classifiers is a list, an item in them could technically be null too, check for null in the array element as well.

src/Microsoft.ComponentDetection.Detectors/pip/PythonResolver.cs
```diff
@@ -210,10 +214,57 @@ private void AddGraphNode(PythonResolverState state, PipGraphNode parent, string
...
+
+            // Split the license classifiers by the " :: " and take the last part of the string
+            licenseClassifiers = licenseClassifiers.Select(x => x.Split(" :: ").Last()).ToList();
```
grvillic, line 257: is the split always like that ` :: `? If so could you link to Python doc in PR description in case there is an update done upstream that we need to include here too. Additionally, a constant for ` :: ` divider string and using it here and in line 254 would help for any future refactoring.

grvillic (changes requested):

src/Microsoft.ComponentDetection.Detectors/pip/PythonResolver.cs
sebasgomez238: Updated the PR description with the relevant docs and added the constants.

src/Microsoft.ComponentDetection.Detectors/pip/PythonResolver.cs
sebasgomez238: I moved these around a bit after reading this comment. The order I chose was really to make the field more representative of who is responsible for the contents inside of the package. The author may no longer be involved with a project so prioritizing the maintainer felt appropriate in this case. But there is no spec on this. Just what I feel would make sense if the field is going to be a part of an SBOM.

cobya: I think the new order makes sense - if we need to update it in the future we can do so.

cobya (approved):

PR #940, opened 2023-12-21
Add Supplier/License info to RustCli Cargo Components.

sebasgomez238:
Adds supplier/license info to Cargo components detected through the RustCli detector. These properties are not made visible in the graph when they are null to avoid null properties on Cargo components detected through the RustCrateDetector.

Approach: Build a dictionary with "{name} {version}" as the key and (license, author) as the value. This allows us to quickly look up the corresponding information when traversing the graph and recording the components.


src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -118,13 +125,19 @@ private static (string Name, string Version) ParseNameAndVersion(string nameAndV
         string id,
         DetectedComponent parent,
         Dep depInfo,
+        IReadOnlyDictionary<string, (string Authors, string License)> packages,
```
cobya, line 128: nit: can we name this something like packagesMetadata to be clearer since this doesn't contain data on the packages themselves?

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -84,6 +84,13 @@ protected override async Task OnFileFoundAsync(ProcessRequest processRequest, ID
 
             var metadata = CargoMetadata.FromJson(cliResult.StdOut);
             var graph = BuildGraph(metadata);
+
+            var packages = metadata.Packages.ToDictionary(
+                x => $"{x.Name} {x.Version}",
+                x => (
+                    (x.Authors == null || x.Authors.All(a => string.IsNullOrWhiteSpace(a))) ? null : string.Join(", ", x.Authors),
```
cobya, line 91: This could be changed to `(x.Authors == null || !x.Authors.Any(a => string.IsNullOrWhiteSpace(a))) ? null : string.Join(", ", x.Authors)` so it isn't always evaluating the whole array, but not strictly necessary since author arrays will have a limited size.

test/Microsoft.ComponentDetection.Detectors.Tests/RustCliDetectorTests.cs
```diff
@@ -292,4 +293,266 @@ public async Task RustCliDetector_InvalidNameAsync()
...
+            if (component.Component is CargoComponent cargoComponent)
+            {
+                cargoComponent.Author.Should().Be("Sample Author 1, Sample Author 2");
+                cargoComponent.License.Should().Be("MIT");
+            }
+        }
+    }
+
+    [TestMethod]
+    public async Task RustCliDetector_AuthorAndLicenseNullAsync()
```
cobya, line 434: Can you add a similar test for empty string values to validate the `x.Authors.All(a => string.IsNullOrWhiteSpace(a))` logic?

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
sebasgomez238: Changed to packagesMetadata as suggested

test/Microsoft.ComponentDetection.Detectors.Tests/RustCliDetectorTests.cs
sebasgomez238: Done

cobya (approved):

cobya:
The rust smoke test failure is unrelated to your changes, so I'll go ahead and merge. Thanks for the contribution!

Issue #928, opened 2023-12-05
The SBOM issue persists despite the successful pipeline for "AI.Models.SmartPicker-Official" and "AI.Models.NL2PowerFX-Official."


hravellamicrosoft:
I wanted to bring to your attention that I have identified similar issues to the ones reported in "https://github.com/microsoft/component-detection/issues/862 " for the "AI.Models.SmartPicker-Official", "AI.Flows.NextBestAction-Official" and "AI.Models.NL2PowerFX-Official" repositories regarding SBOM (Software Bill of Materials).

I have observed successful builds of SBOM manifest artefacts for the two pipelines. However, issues have arisen in the SBOM security compliance aspect.
https://dev.azure.com/powerbi/Business360%20AI/_build/results?buildId=8532158&view=logs&j=0a3c71cb-e72e-5356-5209-9a5ddb8da13b&t=ffa8cf66-1c16-5d05-86dd-8eb2d8103242
https://dev.azure.com/powerbi/Business360%20AI/_build/results?buildId=8604750&view=logs&j=0a3c71cb-e72e-5356-5209-9a5ddb8da13b&t=ffa8cf66-1c16-5d05-86dd-8eb2d8103242
https://dev.azure.com/powerbi/Business360%20AI/_build/results?buildId=8596917&view=logs&j=0a3c71cb-e72e-5356-5209-9a5ddb8da13b&t=ffa8cf66-1c16-5d05-86dd-8eb2d8103242

Could you please review and address these issues with immediate attention, as they have already exceeded the SLA?

Regards,
Hari

AB#2127636

cobya:
Relevant stack trace as Component Detection is an open source project:
```##[error]Error while parsing pip components in "/mnt/vss/_work/1/b/2/_work/1/s/setup.py"
System.IO.InvalidDataException: Number of entries expected in End Of Central Directory does not correspond to number of entries in Central Directory.
   at System.IO.Compression.ZipArchive.ReadCentralDirectory()
   at System.IO.Compression.ZipArchive.GetEntry(String entryName)
   at Microsoft.ComponentDetection.Detectors.Pip.PyPiClient.FetchPackageDependenciesAsync(String name, String version, PythonProjectRelease release)
   at Microsoft.ComponentDetection.Detectors.Pip.PythonResolver.FetchPackageDependenciesAsync(PythonResolverState state, PipDependencySpecification spec)
[...]

PR #962, opened 2024-01-11
Make Python detection more resilient to unexpected failure cases

cobya:
In the Python detectors, there are a few exceptions which are currently unhandled, leading to failures in detection which are not well presented to the user. Add some additional handling when this occurs so users have a good idea of why the error occurred and can investigate themselves to see if detection or their requirements need changes. Related issue #928 

**First case:**
Python package has multiple dependencies specified with the same package name, but differing version constraints. An exception will be hit because of the `ToDictionary` call which expects distinct values.
```
System.ArgumentException: An item with the same key has already been added. Key: portalocker
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](List`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   at Microsoft.ComponentDetection.Detectors.Pip.PythonResolver.InvalidateAndReprocessAsync(PythonResolverState state, PipGraphNode node, PipDependencySpecification newSpec)
   at Microsoft.ComponentDetection.Detectors.Pip.PythonResolver.ProcessQueueAsync(ISingleFileComponentRecorder singleFileComponentRecorder, PythonResolverState state)
   at Microsoft.ComponentDetection.Detectors.Pip.PythonResolver.ResolveRootsAsync(ISingleFileComponentRecorder singleFileComponentRecorder, IList`1 initialPackages)
   at Microsoft.ComponentDetection.Detectors.Pip.PipComponentDetector.OnFileFoundAsync(ProcessRequest processRequest, IDictionary`2 detectorArgs)
```
I'll be filing an issue here for further investigation, but for now, choose the latest specifier given to avoid skipping the dependency all together.

**Second case:**
Python package is given version constraints either from requirements.txt or package dependency version specifiers which are not resolved in PyPi. Previously only the package name was outputted to the user, meaning they don't know why a valid version wasn't found if the package exists on PyPi.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/pip/PythonResolver.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/pip/PythonResolver.cs
@@ -7,6 +7,8 @@ using System.Threading.Tasks;
 using Microsoft.ComponentDetection.Contracts;
 using Microsoft.ComponentDetection.Contracts.TypedComponent;
 using Microsoft.Extensions.Logging;
+using MoreLinq;
+using Newtonsoft.Json;
 
 public class PythonResolver : IPythonResolver
 {
@@ -43,7 +45,7 @@ public class PythonResolver : IPythonResolver
 
                 var result = project.Releases;
 
-                if (result.Keys.Any())
+                if (result is not null && result.Keys.Any())
                 {
                     state.ValidVersionMap[rootPackage.Name] = result;
 
@@ -62,8 +64,9 @@ public class PythonResolver : IPythonResolver
                 else
                 {
                     this.logger.LogWarning(
-                        "Root dependency {RootPackageName} not found on pypi. Skipping package.",
-                        rootPackage.Name);
+                        "Unable to resolve root dependency {PackageName} with version specifiers {PackageVersions} from pypi possibly due to computed version constraints. Skipping package.",
+                        rootPackage.Name,
+                        JsonConvert.SerializeObject(rootPackage.DependencySpecifiers));
                     singleFileComponentRecorder.RegisterPackageParseFailure(rootPackage.Name);
                 }
             }
@@ -113,7 +116,7 @@ public class PythonResolver : IPythonResolver
 
                     var result = project.Releases;
 
-                    if (result.Keys.Any())
+                    if (result is not null && result.Keys.Any())
                     {
                         state.ValidVersionMap[dependencyNode.Name] = result;
                         var candidateVersion = state.ValidVersionMap[dependencyNode.Name].Keys.Any()
@@ -126,8 +129,9 @@ public class PythonResolver : IPythonResolver
                     else
                     {
                         this.logger.LogWarning(
-                            "Dependency Package {DependencyName} not found in Pypi. Skipping package",
-                            dependencyNode.Name);
+                            "Unable to resolve non-root dependency {PackageName} with version specifiers {PackageVersions} from pypi possibly due to computed version constraints. Skipping package.",
+                            dependencyNode.Name,
+                            JsonConvert.SerializeObject(dependencyNode.DependencySpecifiers));
                         singleFileComponentRecorder.RegisterPackageParseFailure(dependencyNode.Name);
                     }
                 }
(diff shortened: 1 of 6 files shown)


cobya:
Related issue for further investigation #963

src/Microsoft.ComponentDetection.Detectors/pip/PythonResolver.cs
```diff
@@ -165,7 +169,24 @@ private async Task<IList<PipGraphNode>> ProcessQueueAsync(ISingleFileComponentRe
 
         node.Value = new PipComponent(pipComponent.Name, candidateVersion, license: pipComponent.License, author: pipComponent.Author);
 
-        var dependencies = (await this.FetchPackageDependenciesAsync(state, newSpec)).ToDictionary(x => x.Name, x => x);
+        var fetchedDependences = await this.FetchPackageDependenciesAsync(state, newSpec);
+
+        // Multiple dependency specification versions can be given for a single package name
+        // Until a better method is divised, choose the latest entry
```
annaowens, line 175:
```suggestion
        // Until a better method is devised, choose the latest entry
```
😝

src/Microsoft.ComponentDetection.Detectors/pip/PythonResolver.cs
```diff
@@ -165,7 +169,24 @@ private async Task<IList<PipGraphNode>> ProcessQueueAsync(ISingleFileComponentRe
...
+        var dependencies = new Dictionary<string, PipDependencySpecification>();
+        fetchedDependences.ForEach(d =>
```
annaowens, line 177: Do we have a pattern for using an id rather than a name in other detectors?

src/Microsoft.ComponentDetection.Detectors/pip/SimplePythonResolver.cs
```diff
@@ -366,7 +374,24 @@ await using (var stream = entry.Open())
 
         node.Value = new PipComponent(pipComponent.Name, candidateVersion);
 
-        var dependencies = (await this.FetchPackageDependenciesAsync(state, newSpec)).ToDictionary(x => x.Name, x => x);
+        var fetchedDependences = await this.FetchPackageDependenciesAsync(state, newSpec);
+
+        // Multiple dependency specification versions can be given for a single package name
+        // Until a better method is divised, choose the latest entry
```
annaowens, line 380:
```suggestion
        // Until a better method is devised, choose the latest entry
```

src/Microsoft.ComponentDetection.Detectors/pip/PythonResolver.cs
annaowens: I think I'm comfortable if you're planning to handle this in a future PR--these dependencies aren't getting registered today because of the exception so there's not a risk of dropping deps.

src/Microsoft.ComponentDetection.Detectors/pip/SimplePythonResolver.cs
```diff
@@ -207,8 +209,9 @@ private async Task<IList<PipGraphNode>> ProcessQueueAsync(ISingleFileComponentRe
                     else
                     {
                         this.logger.LogWarning(
-                            "Unable to resolve dependency package {DependencyName} gotten from pypi possibly due to invalid versions. Skipping package",
-                            dependencyNode.Name);
+                            "Unable to resolve non-root dependency {PackageName} with version specifiers {PackageVersions} from pypi possibly due to computed version constraints. Skipping package.",
```
annaowens, line 212: Would be nice to share some of this logic in the future.

annaowens (approved):

src/Microsoft.ComponentDetection.Detectors/pip/PythonResolver.cs
cobya: Yes, plan is to handle them in #963

PR #990, opened 2024-01-29
Add fallback logic for rust CLI detector and additional telemetry

FernandoRojo:


src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -170,4 +242,195 @@ private static (string Name, string Version) ParseNameAndVersion(string nameAndV
...
         }
     }
+
+    private IComponentStream FindCorrespondingCargoLock(IComponentStream cargoToml, ISingleFileComponentRecorder singleFileComponentRecorder)
+    {
+        var cargoLockLocation = Path.Combine(Path.GetDirectoryName(cargoToml.Location), "Cargo.lock");
+        return this.ComponentStreamEnumerableFactory.GetComponentStreams(new FileInfo(cargoToml.Location).Directory, new List<string> { "Cargo.lock" }, (name, directoryName) => false, recursivelyScanDirectories: false).FirstOrDefault();
+    }
+
+    private async Task ProcessCargoLockFallbackAsync(IComponentStream cargoTomlFile, ISingleFileComponentRecorder singleFileComponentRecorder, RustGraphTelemetryRecord record)
```
cobya, line 252: We should add test cases here

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -120,6 +171,27 @@ private static (string Name, string Version) ParseNameAndVersion(string nameAndV
...
+    private static bool IsLocalPackage(CargoPackage package) => package.Source == null;
+
+    private static bool ParseDependency(string dependency, out string packageName, out string version, out string source)
+    {
+        var match = DependencyFormatRegex.Match(dependency);
+        var packageNameMatch = match.Groups["packageName"];
+        var versionMatch = match.Groups["version"];
+        var sourceMatch = match.Groups["source"];
+
+        packageName = packageNameMatch.Success ? packageNameMatch.Value : null;
```
cobya, line 183: Is `.Value` not null if `Success` is false?

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -120,6 +171,27 @@ private static (string Name, string Version) ParseNameAndVersion(string nameAndV
...
+        version = versionMatch.Success ? versionMatch.Value : null;
+        source = sourceMatch.Success ? sourceMatch.Value : null;
+
+        if (string.IsNullOrEmpty(source))
```
cobya, line 187: Does this need to be IsNullOrWhitespace?

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -170,4 +242,195 @@ private static (string Name, string Version) ParseNameAndVersion(string nameAndV
...
+            record.FallbackCargoLockFound = false;
+            return;
+        }
+
+        record.FallbackCargoLockLocation = cargoLockFile.Location;
+        record.FallbackCargoLockFound = true;
+        using var fileStream = new FileStream(cargoLockFile.Location, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
+        using var reader = new StreamReader(fileStream);
+        var content = await reader.ReadToEndAsync();
+        var options = new TomlModelOptions
```
cobya, line 267: Can we make this a class const so we aren't initing it multiple times?

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -170,4 +242,195 @@ private static (string Name, string Version) ParseNameAndVersion(string nameAndV
...
+                // We are adding a root edge (from a local package)
+                singleFileComponentRecorder.RegisterUsage(detectedComponent, isExplicitReferencedDependency: true);
+            }
+            else
+            {
+                // we are adding an edge within the graph
+                singleFileComponentRecorder.RegisterUsage(detectedComponent, isExplicitReferencedDependency: false, parentComponentId: parentComponent.Id);
+            }
+        }
+        catch (Exception e)
```
cobya, line 425: We should avoid catching such generic exceptions. If we're only throwing FormatExceptions catch that

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
FernandoRojo: Only if none of the groups match, so in the case of it just being a package name the other two fields have a value of "" instead of null. 

But if none of them match then yes they are all null 👎

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
FernandoRojo: wouldn't hurt

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
FernandoRojo: In the majority of cases yes, but there are a few other exceptions that could mess with parsing, like out of memory exceptions or other weird edge cases, we log whatever the error is but take the safe route of catching all to not fail out of the entire detector. It's horrible practice but it matches how we do dependency parsing for our other detectors :(

cobya (approved):

PR #1006, opened 2024-02-09
Catch version spec failures in Python detection

cobya:
When an invalid version specification is returned from PyPi (i.e. `Requires-Dist: c (>dev)`) the current implementation of Python resolvers both throw the exception all the way out, leading to a complete failure of Python graph creation. In this case we should log the warning and continue executing to get as many dependencies as possible.


grvillic:
Let's bump Pip detectors version Id, since we are picking up dependency graphs we didn't use to

grvillic (approved):

PR #1009, opened 2024-02-19
feat(rust): allow specifying features to `cargo metadata`

### LINKED ISSUE #1001: Rust: allow specific features to be set for RustCLI detector
cargo.toml files can have specified features that can change the component set when building, currently the CLI detector assumes all features are enabled but we should implement some support to enable specific features to be set/unset

melotic:
Closes #1001 

This uses a new environment variable to configure the Rust CLI detector, `CD_RUST_CLI_FEATURES`. Open to any other ideas on how to pass this into the detector.


cobya:
I think it's time to start investing in a file-based configuration approach where we have a top-level service read a file from something like `.detection/config.json` where we keep these environment variables, but that would probably be outside the scope of this 🤨

PR #1015, opened 2024-02-23
Update RustCLI processing to handle virtual manifests / skip over vendor packages

FernandoRojo:


src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -99,9 +99,9 @@ protected override async Task OnFileFoundAsync(ProcessRequest processRequest, ID
 
                 if (cliResult.ExitCode != 0)
```
grvillic, line 100: We should bump detector version above

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -118,22 +118,28 @@ protected override async Task OnFileFoundAsync(ProcessRequest processRequest, ID
...
+                        {
+                            var componentKey = $"{dep.Id}";
+                            if (!visitedDependencies.Contains(componentKey))
+                            {
+                                visitedDependencies.Add(componentKey);
+                                this.TraverseAndRecordComponents(processRequest.SingleFileComponentRecorder, componentStream.Location, graph, dep.Id, null, null, packages, visitedDependencies, explicitlyReferencedDependency: true, isTomlRoot: true);
+                            }
+                        }
+                    }
+                    else if (!record.WasRustFallbackStrategyUsed)
```
grvillic, line 140: just to confirm that we want the else here. If we have a root and fallback strategy was used, we will ignore?

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -118,22 +118,28 @@ protected override async Task OnFileFoundAsync(ProcessRequest processRequest, ID
...
-                        record.FallbackReason = "Virtual Manifest";
-                    }
+                        this.Logger.LogWarning("Virtual Manifest: {Location}", processRequest.ComponentStream.Location);
 
-                    HashSet<string> visitedDependencies = new();
-                    if (!record.WasRustFallbackStrategyUsed)
+                        foreach (var dep in metadata.Resolve.Nodes)
+                        {
+                            var componentKey = $"{dep.Id}";
+                            if (!visitedDependencies.Contains(componentKey))
```
grvillic, line 133: Can we add some unit tests to validate this new logic?

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
FernandoRojo: If we determine that we're going to use the fallback strategy, we don't want to try and process the metadata. but with the added support for virtual manifests now this can just be an else instead of an else-if

src/Microsoft.ComponentDetection.Detectors/rust/Contracts/CargoMetadata.cs
```diff
@@ -453,8 +453,16 @@ public override CrateType Read(ref Utf8JsonReader reader, Type typeToConvert, Js
...
+                return CrateType.RLib;
+            case "dylib":
+                return CrateType.DyLib;
+            case "cdylib":
+                return CrateType.CdyLib;
+            case "staticlib": 
+                return CrateType.StaticLib;
         }
-        throw new Exception("Cannot unmarshal type CrateType");
+        throw new Exception($"Cannot unmarshal type CrateType - {value}");
```
annaowens, line 465: Should we return type Other and log this as an error? Guess it depends if it would break things / how often new CrateTypes are added / how many we might not know about.

FernandoRojo: I used the docs to determine the full set of allowable values, and it doesn't look like this is something that's updated often, My only worry with just logging it as an error is that if something new does get introduced in the future we would just silently be marking those as unknown when there could be some additional processing that needs to happen for a potentially new crate-type

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
FernandoRojo: added

grvillic (approved):

Issue #32, opened 2021-12-29
Python detector does not match pre-release versions explicitly specified in requirements.txt


cobya:
IPyPiClient when retrieving release information will not add pre-release versions to the list of potential versions [(see IPyPiClient)](https://github.com/microsoft/component-detection/blob/main/src/Microsoft.ComponentDetection.Detectors/pip/IPyPiClient.cs#L205) to match requirements.txt versions to in PythonResolver.cs. 

This leads to cases where pre-release versions are explicitly referenced in requirements.txt (ex. numpy 1.22.0rc3 when the current release version is 1.21.5) but the detector will output the selected version as the release version and not the pre-release.

cobya:
Closed with #1021.

PR #1021, opened 2024-03-01
Support for release candidate versions python detector

jcfiorenzano:
This PR is to fix issue #32 

Ex of requirement.txt file
```
numpy==1.23.0rc3
xyz01==0.2
```

Dependency graph results

image


src/Microsoft.ComponentDetection.Detectors/pip/IPyPiClient.cs
```diff
@@ -209,7 +209,7 @@ public async Task<PythonProject> GetProjectAsync(PipDependencySpecification spec
             {
                 var parsedVersion = PythonVersion.Create(release.Key);
                 if (release.Value != null && release.Value.Count > 0 &&
-                    parsedVersion.Valid && parsedVersion.IsReleasedPackage &&
+                    parsedVersion.Valid &&
```
cobya, line 212: Similar logic is used in SimplePip, can you verify if it needs to be updated there too?

cobya (approved):

Issue #503, opened 2023-04-11
Add support for pnpm lockfile v6


annaowens:
pnpm version 8 includes an upgrade to lockfile version 6, but we currently only support up to pnpm lockfile v5. 

Since we are unable to accurately parse the lockfile v6, we are not correctly detecting development dependencies for pnpm v8. 

We need to update our pnpm detector to handle the new v6 lockfile. At the time of writing this issue, there isn't documentation on the lockfile v6 format: https://github.com/pnpm/spec/issues/4. 

This is a similar issue to #476 -- we've already seen the same problem with NPM lockfile v3. 

AB#2121069

wterpstra:
Hey, the lockfile v6 format is available now.

https://github.com/pnpm/spec/blob/master/lockfile/6.0.md

dwelch2344:
Just came to upvote this. Would be happy to contribute, if someone would be open to some onboarding support...

vhuseinova-msft:
Just came to upvote this too. It would be great to have support for the latest lockfile formats.

CraigMacomber:
I looked at the relevant code and it doesn't seem too hard. I'm going to make an attempt at fixing this during the hackathon next week.

CraigMacomber:
Pnpm 9 (currently in alpha) moves to lock file format v7: thats going to be an issue soon even if v6 support is added.

CraigMacomber:
My work in progress for this lives in the branch https://github.com/CraigMacomber/component-detection/tree/pnpmv6

I have it working as far as I can tell. It manages to parse the included test lock file with v6 as well as a much larger test using the lock file from FluidFramework (which actually uses the shared shrinkwrap mode, file references, peer deps etc).

I still need to add unit tests v6, but otherwise I think its almost ready for review. I hope to have a PR up for review tomorrow.

CraigMacomber:
https://github.com/microsoft/component-detection/pull/1028 fixes this and is currently awaiting review.

annaowens:
@CraigMacomber Pinged the team to see if we can get some traction on the PR review. Otherwise I'll plan to take a look next week.

CraigMacomber:
> @CraigMacomber Pinged the team to see if we can get some traction on the PR review. Otherwise I'll plan to take a look next week.

"Next week" happened. Any update?

CraigMacomber:
@annaowens The PR addressing this issue been waiting for review for a response from maintainers for over a month.

[...]

PR #1028, opened 2024-03-12
Add support for pnpm lock file format 6

CraigMacomber:
This addresses https://github.com/microsoft/component-detection/issues/503 by doing the following:

- Explicitly mark code specific to version 5 as "V5"
- Add new code for V6.
- Add version detection logic to select which code to run.
- Add tests for new code.

The old code wasn't clear about what file formats it actually worked with, and a lot of its tests seems to be from versions earlier than 5 which it didn't really parse correctly (ex: the version it parses isn't present in those versions as it had a different name).
Thus to preserve existing behavior I pointed cases which failed to detect a version through the old V5 code.

The included verification tests didn't cover a lot of cases (link and file references, aliased packages, shared shrink-wrap format, disambiguated deps with peer deps) so I also tested on the package lock from https://github.com/microsoft/FluidFramework which is 44k lines and has all the mentioned edge cases.
If desired, I can add the v5 and v6 versions of that file to the verification tests directory.

This should be in a working state but I need to actually understand how the verification tests results are verifies before this should be merged.

This is my first time working in this codebase, as well as the first time in several years that I'm working in C#, so please be extra careful reviewing this: don't trust that I know what I'm doing.


src/Microsoft.ComponentDetection.Detectors/pnpm/PnpmComponentDetector.cs
```diff
@@ -104,14 +125,97 @@ private void RecordDependencyGraphFromFile(PnpmYaml yaml, ISingleFileComponentRe
...
+        foreach (var (_, (component, package)) in components)
+        {
+            foreach (var (name, version) in package.dependencies ?? Enumerable.Empty<KeyValuePair<string, string>>())
+            {
+                var pnpmDependencyPath = PnpmParsingUtilities.ReconstructPnpmDependencyPathV6(name, version);
+
+                // If this lookup fails, then pnpmDependencyPath was either parsed incorrectly or constructed incorrectly.
+                var (referenced, _) = components[pnpmDependencyPath];
+
+                singleFileComponentRecorder.RegisterUsage(referenced, parentComponentId: component.Component.Id, isExplicitReferencedDependency: false);
```
CraigMacomber, line 167: Its unclear to me if I need to respecify dev-dependency-ness here. Would doing so indicate this dependency edge is a dev dependency, or and I overwriting the dev-dependency ness of the target component with the default by not specifying it here?

src/Microsoft.ComponentDetection.Detectors/pnpm/PnpmComponentDetector.cs
```diff
@@ -30,7 +31,7 @@ public class PnpmComponentDetector : FileComponentDetector
 
     public override IEnumerable<ComponentType> SupportedComponentTypes { get; } = new[] { ComponentType.Npm };
 
-    public override int Version { get; } = 5;
+    public override int Version { get; } = 6;
```
CraigMacomber, line 34: I think I'm supposed to update this because I made changes, but I'd like confirmation that that's the correct thing to do.

src/Microsoft.ComponentDetection.Detectors/pnpm/PnpmParsingUtilities.cs
```diff
@@ -3,43 +3,52 @@
...
-            throw new ArgumentNullException(nameof(pnpmPackage));
-        }
+        var deserializer = new DeserializerBuilder()
+            .IgnoreUnmatchedProperties()
+            .Build();
+        return deserializer.Deserialize<PnpmYamlV6>(new StringReader(fileContent));
+    }
 
-        return string.Equals(bool.TrueString, pnpmPackage.dev, StringComparison.InvariantCultureIgnoreCase);
+    /// <summary>
```
CraigMacomber, line 37: Nothing in here had doc comments before, but it seems like docs would be good so I added some. he projects configuration seems to like this format so I used it, but I'm not sure what the policy on documentation is.

CraigMacomber:
It appears that the VS code dev container setup doesn't include PowerShell so I can't run the VerificationTests.ps1 script locally. The instructions in there also make explicit reference to visual studio. It also says "The automation script above will output variables that needs to be replaced in `ComponentDetectionIntegrationTests.cs` file.": as far as I can tell thats not accurate as it looks like that file pulls in the data from environment variables.

Anyway its unclear to me what the workflow is for adding more test cases to the verification script. I'd like to include my added v6 file works but I don't know how.

CraigMacomber:
Looking at the code coverage, I determined I could use a test for the shared-shrinkwrap case that workspaces use, and one for renamed/aliased dependencies, so I have added both. More testing is possible (ex: file and link cases, peer deps etc.) but I think I'm past the point of diminishing returns.

CraigMacomber:
@microsoft/ose-component-detection-maintainers : I'd like assistance dealing with the snapshot tests as noted in https://github.com/microsoft/component-detection/pull/1028#issuecomment-1992470325. If this is not practical to get assistance on, I can revert the extended snapshot test coverage which should get this passing on CI: maybe that would get this reviewed sooner?

src/Microsoft.ComponentDetection.Detectors/pnpm/PnpmComponentDetector.cs
cobya: Our component merging logic will do the following where if any detection is classified as non-dev it will categorize the component as non-dev.

First Detection Dev Dep | Second Detection Dev Dep | Final Result
false                             | false                              | false
true                              | true                               | true
true                              | false                              | false

src/Microsoft.ComponentDetection.Detectors/pnpm/PnpmComponentDetector.cs
cobya: I'd be happy to help make these changes as well if needed

src/Microsoft.ComponentDetection.Detectors/pnpm/PnpmComponentDetector.cs
cobya: When passing null for isDevelopmentDependency whatever the previous value is will stay.

src/Microsoft.ComponentDetection.Detectors/pnpm/PnpmComponentDetector.cs
```diff
@@ -104,14 +125,97 @@ private void RecordDependencyGraphFromFile(PnpmYaml yaml, ISingleFileComponentRe
...
+        // "dedicated shrinkwrap" (single package) case:
+        ProcessDependencySet(yaml);
+
+        // "shared shrinkwrap" (workspace / mono-repos) case:
+        foreach (var (_, package) in yaml.importers ?? Enumerable.Empty<KeyValuePair<string, PnpmHasDependenciesV6>>())
+        {
+            ProcessDependencySet(package);
+        }
+
+        void ProcessDependencySet(PnpmHasDependenciesV6 item)
```
cobya, line 182: nit: can these be declared as explicit functions rather than locals for RecordDependencyGraphFromFileV6?

src/Microsoft.ComponentDetection.Detectors/pnpm/PnpmParsingUtilities.cs
cobya: We appreciate any new docs being added, and yes this is the format we use 😄

src/Microsoft.ComponentDetection.Detectors/pnpm/PnpmParsingUtilities.cs
```diff
@@ -50,7 +69,65 @@ private static (string Name, string Version) ExtractNameAndVersionFromPnpmPackag
...
+    {
+        /*
+         * The format is documented at https://github.com/pnpm/spec/blob/master/dependency-path.md.
+         * At the writing it does not seem to reflect changes which were made in lock file format v6:
+         * See https://github.com/pnpm/spec/issues/5.
+         */
+
+        // Strip parenthesized suffices from package. These hold peed dep related information that is unneeded here.
+        // An example of a dependency path with these: /webpack-cli@4.10.0(webpack-bundle-analyzer@4.10.1)(webpack-dev-server@4.6.0)(webpack@5.89.0)
+        var fullPackageNameAndVersion = pnpmDependencyPath.Split("(")[0];
```
cobya, line 87: is this guaranteed to always have [0]? do we need to do index checks beforehand?

cobya:
> @microsoft/ose-component-detection-maintainers : I'd like assistance dealing with the snapshot tests as noted in [#1028 (comment)](https://github.com/microsoft/component-detection/pull/1028#issuecomment-1992470325). If this is not practical to get assistance on, I can revert the extended snapshot test coverage which should get this passing on CI: maybe that would get this reviewed sooner?

When adding new components like this, the verification test failures are expected as it is performing a diff based on the last "known good version" so adding new files will trip the failure. I'll verify before merging but you shouldn't need to worry about this for now.

src/Microsoft.ComponentDetection.Detectors/pnpm/PnpmComponentDetector.cs
CraigMacomber: That sounds like an excellent way to reduce risk in cases like this.

> I'd be happy to help make these changes as well if needed

That would be great. I don't know this codebase well (I just spent a day and a half making this as part of a hackathon), so I suspect you could do that much more efficiently than I could. I'm happy to have you take over any part or all of this work.

src/Microsoft.ComponentDetection.Detectors/pnpm/PnpmComponentDetector.cs
CraigMacomber: I choose to do it this way so they could be scoped to v6 specific code, and not require as many parameters. They could of course be moved to a larger scope, given more specific names, and have everything passed in: that just seemed messier to me. If you refactor the v5 and v6 detector to be more separated, moving these to a larger scope would be more attractive.

src/Microsoft.ComponentDetection.Detectors/pnpm/PnpmParsingUtilities.cs
CraigMacomber: My understanding is splitting a string always returns at least one part, even if there is no separator or the string is empty, so this should be fine. There likely is a more explicit way to do this (for example with a regular expression), but this should work fine.

cobya:
Closing this as #1110 has been opened with the related experiments changes.

Issue #858, opened 2023-10-16
Issue while mapping DependencyScope from maven dependency tree 


tarun06:
Method name -> MavenParsingUtilities -> GetMavenComponentStringInfo  

On using below dependency with scope -> compile, in Maven Project, 

```
<dependency>
	<groupId>org.seleniumhq.selenium</groupId>
        <artifactId>selenium-java</artifactId>
	<version>4.8.0</version>
	<scope>compile</scope>
  </dependency> 
```

The SBOM throws an exception

**System.InvalidOperationException: Invalid scope ('linux-x86_64') found for 'io.netty:netty-transport-native-epoll:jar:4.1.87.Final:linux-x86_64' found in generated dependency graph**

More Info ->

Dependency tree generated using maven dependency CLI command 

`mvn dependency:tree -B -DoutputFile=bcde.mvndeps -DoutputType=text -f <SampleProjectPath>`, contains value linux-x86_64 and osx-x86_64 which further leads to an exception while parsing dependency Scope. 

Snapshot of Dependency tree 

```|  +- org.seleniumhq.selenium:selenium-remote-driver:jar:4.8.0:compile
|  |  +- com.beust:jcommander:jar:1.82:compile
|  |  +- io.netty:netty-buffer:jar:4.1.87.Final:compile
|  |  +- io.netty:netty-codec-http:jar:4.1.87.Final:compile
|  |  |  +- io.netty:netty-codec:jar:4.1.87.Final:compile
|  |  |  \- io.netty:netty-handler:jar:4.1.87.Final:compile
|  |  +- io.netty:netty-common:jar:4.1.87.Final:compile
|  |  +- io.netty:netty-transport-classes-epoll:jar:4.1.87.Final:compile
|  |  +- io.netty:netty-transport-classes-kqueue:jar:4.1.87.Final:compile
|  |  +- io.netty:netty-transport-native-epoll:jar:4.1.87.Final:linux-x86_64
|  |  +- io.netty:netty-transport-native-kqueue:jar:4.1.87.Final:osx-x86_64
|  |  +- io.netty:netty-transport-native-unix-common:jar:4.1.87.Final:compile
```

AB#2114682

RushabhBhansali:
@tarun06 
When printing dependency tree on my end, below is the output
```
[INFO]    +- io.netty:netty-transport-native-epoll:jar:linux-x86_64:4.1.60.Final:compile
[...]

PR #1033, opened 2024-03-14
Evaluating invalid maven scope as "Compile"

RushabhBhansali:
### Summary: 
Currently maven detector is failing for invalid `<scope>` in maven dependency file. However, these dependencies are valid as per maven cli and evaluated as "Compile". Component Detector matches the maven's behavior to evaluate these "invalid" scope as "Compile"

#858 

### Testing:
- local testing
- updated unit test for invalid dependency

### DIFF
--- a/test/Microsoft.ComponentDetection.Detectors.Tests/MavenParsingUtilitiesTests.cs
+++ b/test/Microsoft.ComponentDetection.Detectors.Tests/MavenParsingUtilitiesTests.cs
@@ -1,4 +1,4 @@
-﻿namespace Microsoft.ComponentDetection.Detectors.Tests;
+namespace Microsoft.ComponentDetection.Detectors.Tests;
 
 using System;
 using FluentAssertions;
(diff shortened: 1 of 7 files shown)


test/Microsoft.ComponentDetection.Detectors.Tests/MavenParsingUtilitiesTests.cs
```diff
@@ -79,10 +79,16 @@ public void GenerateDetectedComponentAndIsDeveDependencyAndDependencyScope_Devel
     }
 
     [TestMethod]
-    public void GenerateDetectedComponentAndIsDeveDependencyAndDependencyScope_InvalidScope()
+    public void GenerateDetectedComponentAndIsDeveDependencyAndDependencyScope_InvalidScopeIsEvaluatedAsCompile()
```
grvillic, line 82: nit: typo in test name `DeveDependency`

grvillic (approved):

grvillic (commented):
We need to bump the detector versions that use this parsing utility since we are generating a different dependency graph than before.

grvillic (approved):

test/Microsoft.ComponentDetection.Detectors.Tests/MavenParsingUtilitiesTests.cs
RushabhBhansali: fixed the typo for this test and all the other tests in the class.

Issue #122, opened 2022-06-08
Go CLI detector should take into account `replace` go.mod overrides


cobya:
When running `go list -m`, the Go CLI detector does not currently take into account any `replace` values specified within the corresponding go.mod file.

Example:
go.mod
```
replace (
    github.com/miekg/dns => github.com/miekg/dns v1.1.42
)
```

If a version <1.1.42 is requested as a dependency in go.sum, the override is not taken into consideration and the version returned by Component Detection will be whatever is requested from go.sum

cobya:
Resolved with #1052 and #1045.

PR #1045, opened 2024-03-25
Replace in Go Mod

amitla1:
Currently replace statements in go.mod files are ignored, this will abide by the replace statements
https://github.com/microsoft/component-detection/issues/122

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/go/GoComponentWithReplaceDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/go/GoComponentWithReplaceDetector.cs
@@ -17,7 +17,7 @@ using Microsoft.Extensions.Logging;
 using MoreLinq;
 using Newtonsoft.Json;
 
-public class GoComponentWithReplaceDetector : FileComponentDetector
+public class GoComponentWithReplaceDetector : FileComponentDetector, IExperimentalDetector
 {
     private static readonly Regex GoSumRegex = new(
         @"(?<name>.*)\s+(?<version>.*?)(/go\.mod)?\s+(?<hash>.*)",
--- a/src/Microsoft.ComponentDetection.Detectors/go/GoComponentWithReplaceDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/go/GoComponentWithReplaceDetector.cs
@@ -14,9 +14,10 @@ using Microsoft.ComponentDetection.Contracts;
 using Microsoft.ComponentDetection.Contracts.Internal;
 using Microsoft.ComponentDetection.Contracts.TypedComponent;
 using Microsoft.Extensions.Logging;
+using MoreLinq;
 using Newtonsoft.Json;
 
-public class GoComponentDetector : FileComponentDetector
+public class GoComponentWithReplaceDetector : FileComponentDetector
 {
     private static readonly Regex GoSumRegex = new(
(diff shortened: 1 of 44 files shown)


src/Microsoft.ComponentDetection.Detectors/go/GoComponentWithReplaceDetector.cs
```diff
@@ -0,0 +1,560 @@
...
+using Microsoft.ComponentDetection.Common;
+using Microsoft.ComponentDetection.Common.Telemetry.Records;
+using Microsoft.ComponentDetection.Contracts;
+using Microsoft.ComponentDetection.Contracts.Internal;
+using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using Microsoft.Extensions.Logging;
+using MoreLinq;
+using Newtonsoft.Json;
+
+public class GoComponentWithReplaceDetector : FileComponentDetector, IExperimentalDetector
```
cobya, line 20: I'd prefer that we use an inherited class here rather than a full copy paste of the file. If the main addition is `ReplaceGoComponents` we can override the functions that call it in the inherited class

src/Microsoft.ComponentDetection.Detectors/go/GoComponentWithReplaceDetector.cs
```diff
@@ -0,0 +1,560 @@
...
+
+    public override string Id => "GoWithReplace";
+
+    public override IEnumerable<string> Categories => new[] { Enum.GetName(typeof(DetectorClass), DetectorClass.GoMod) };
+
+    public override IList<string> SearchPatterns { get; } = new List<string> { "go.mod", "go.sum" };
+
+    public override IEnumerable<ComponentType> SupportedComponentTypes { get; } = new[] { ComponentType.Go };
+
+    public override int Version => 7;
```
cobya, line 55: It would be v1 of a new detector

cobya (changes requested):

src/Microsoft.ComponentDetection.Detectors/go/GoComponentWithReplaceDetector.cs
amitla1: would rather leave it like this since the only reason its in a new files for experiments. The other detector would get deleted and that would mean reconfiguring the code again

cobya (approved):

src/Microsoft.ComponentDetection.Detectors/go/GoComponentWithReplaceDetector.cs
```diff
@@ -0,0 +1,560 @@
...
+        GoGraphTelemetryRecord goGraphTelemetryRecord)
+    {
+        using var reader = new StreamReader(file.Stream);
+
+        // There can be multiple require( ) sections in go 1.17+. loop over all of them.
+        while (!reader.EndOfStream)
+        {
+            var line = await reader.ReadLineAsync();
+
+            while (line != null && !line.StartsWith("require ("))
```
grvillic, line 322: @amitla1 - Not sure if you were aware, but I think there is no need to parse the require statements directly [go list -m](https://go.dev/ref/mod#go-list-m) does return a property called "Replace" with the module that will replace the existing entry. We are already running that command so it shouldn't add cost there, but it won't be available for the fallback scenario.

PR #1046, opened 2024-03-25
Update VCPKG component identifier in detector

grvillic:
Updating slightly the component ID to include the origin of the VCPKG package for increased accuracy, as packages can be consumed from anywhere. This component type behaves similarly to Git components for some of their resources.

### DIFF
--- a/src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
+++ b/src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
@@ -42,14 +42,19 @@ public class VcpkgComponent : TypedComponent
     {
         get
         {
+            var componentLocationPrefix = string.Empty;
+            if (!string.IsNullOrWhiteSpace(this.DownloadLocation) && !this.DownloadLocation.Trim().Equals("NONE", System.StringComparison.InvariantCultureIgnoreCase))
+            {
+                componentLocationPrefix = $"{this.DownloadLocation} : ";
+            }
+
+            var componentPortVersionSuffix = " ";
             if (this.PortVersion > 0)
             {
-                return $"{this.Name} {this.Version}#{this.PortVersion} - {this.Type}";
+                componentPortVersionSuffix = $"#{this.PortVersion} ";
             }
-            else
-            {
-                return $"{this.Name} {this.Version} - {this.Type}";
-            }
+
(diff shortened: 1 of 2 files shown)


src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
```diff
@@ -42,14 +42,19 @@ public override string Id
     {
         get
         {
-            if (this.PortVersion > 0)
+            var componentLocationPrefix = string.Empty;
+            if (!string.IsNullOrWhiteSpace(this.DownloadLocation) && !this.DownloadLocation.Trim().Equals("NONE", System.StringComparison.InvariantCultureIgnoreCase))
             {
-                return $"{this.Name} {this.Version}#{this.PortVersion} - {this.Type}";
+                componentLocationPrefix = $"{this.DownloadLocation} : ";
```
grvillic, line 48: The main change is including the downloadlocation of the component as prefix, the rest of the string stayed the same.

src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
```diff
@@ -42,14 +42,19 @@ public override string Id
...
-            else
+
+            var componentPortVersionSuffix = " ";
+            if (this.PortVersion > 0)
             {
-                return $"{this.Name} {this.Version} - {this.Type}";
+                componentPortVersionSuffix = $"#{this.PortVersion} ";
             }
+
+            return $"{componentLocationPrefix}{this.Name} {this.Version}{componentPortVersionSuffix}- {this.Type}";
```
RushabhBhansali, line 57: do we not need space after `{componentPortVdrsionSuffix}` for portVersion > 0 scenario ? 

if yes, we can write it like this ...
```
 var componentPortVersionSuffix = EMTPTY;
            if (this.PortVersion > 0)
            {
                componentPortVersionSuffix = $"#{this.PortVersion} ";
            }

            return $"{componentLocationPrefix}{this.Name} {this.Version}{componentPortVersionSuffix} - {this.Type}";
```

grvillic: Yes that space is as suffix in line 54

RushabhBhansali (approved):

PR #1052, opened 2024-03-28
Go CLI Replacement

amitla1:
Adding replacement module for the GO CLI scan not just fallback scan 
#122

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/go/GoComponentWithReplaceDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/go/GoComponentWithReplaceDetector.cs
@@ -320,28 +320,28 @@ public class GoComponentWithReplaceDetector : FileComponentDetector, IExperiment
         while (!reader.EndOfStream)
         {
             var line = await reader.ReadLineAsync();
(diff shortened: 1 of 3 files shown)


src/Microsoft.ComponentDetection.Detectors/go/GoComponentWithReplaceDetector.cs
```diff
@@ -33,7 +33,7 @@ public class GoComponentWithReplaceDetector : FileComponentDetector, IExperiment
         IObservableDirectoryWalkerFactory walkerFactory,
         ICommandLineInvocationService commandLineInvocationService,
         IEnvironmentVariableService envVarService,
-        ILogger<GoComponentDetector> logger)
+        ILogger<GoComponentWithReplaceDetector> logger)
     {
```
cobya, line 37: are you keeping the detector version from previous PR or do we need to bump it again?

amitla1: same version

cobya (approved):

PR #1101, opened 2024-05-08
Use a random name for working folder for IvyDetector

abyss638:
IvyDetector creates files in a folder named "ComponentDetection_Ivy" in the temporary folder.
However if multiple detectors run in parallel, it brings a race condition when detectors manipulate files in the same folder.

Solution: Use random name for IvyDetector working folder.


cobya (approved):
Thanks for the contribution! If this doesn't fix the issue you're hitting, feel free to open a new PR 👍

PR #1106, opened 2024-05-14
Update RustCli Parsing to process pkgId, and introduce manual override

FernandoRojo:
Allows users to manually override rustCLI detector by setting an environment variable DisableRustCliScan=True

Updates cargo metadata output processing to adhere to changes introduced that make the output follow the PkgId Spec: https://doc.rust-lang.org/cargo/reference/pkgid-spec.html

Changes to the metadata output were introduced in 1.77: https://github.com/rust-lang/cargo/pull/12914

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
@@ -212,6 +212,16 @@ public class RustCliDetector : FileComponentDetector
         // 4. url#version => https://github.com/rust-lang/cargo#0.33.0
         // 5. url#name => https://github.com/rust-lang/crates.io-index#packageName
         // 6. url#name@version => https://github.com/rust-lang/cargo#crates-io@0.21.0
+
+        // First, try parsing using the old format in cases where a version of rust older than 1.77 is being used.
+        if (ParseDependencyCargoLock(dependency, out packageName, out version, out source))
+        {
+            if (!(string.IsNullOrEmpty(packageName) || string.IsNullOrEmpty(version)))
+            {
+                return true;
+            }
+        }
+
         var match = DependencyFormatRegexPkgId.Match(dependency);
         packageName = null;
         version = null;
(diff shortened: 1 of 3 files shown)


src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -187,9 +203,81 @@ private static bool ShouldFallbackFromError(string error)
         return true;
     }
 
-    private static bool ParseDependency(string dependency, out string packageName, out string version, out string source)
+    private static bool ParseDependencyMetadata(string dependency, out string packageName, out string version, out string source)
     {
-        var match = DependencyFormatRegex.Match(dependency);
+        // There are a few different formats for pkgids: https://doc.rust-lang.org/cargo/commands/cargo-pkgid.html#description
```
FernandoRojo, line 208: There must be a better way to do this, was pressed for time but if there are any suggestions here I'm all ears

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
```diff
@@ -187,9 +203,81 @@ private static bool ShouldFallbackFromError(string error)
...
+
+            // if there is a third group, then the second must be packageName, third is version.
+            if (thirdGroup.Success)
+            {
+                packageName = secondGroup.Value;
+                version = thirdGroup.Value;
+            }
+
+            // if there is no third group, but there is a second, the second group could be either the name or the version, check if the value starts with a number (not allowed)
+            else if (secondGroup.Success)
```
cobya, line 254: Can we try and get a test case covering these two below matches?

cobya:
@FernandoRojo to verify, the new locations in snapshot verification are expected correct?

src/Microsoft.ComponentDetection.Detectors/rust/RustCliDetector.cs
FernandoRojo: changed the logic to use the pkgIds directly, so this whole parsing method is no longer needed.

cobya (approved):

PR #1110, opened 2024-05-15
Add pnpm v6 support as an experimental detector

cobya:
After discussing with @CraigMacomber on #1028 we decided to take an experimental approach during rollout of PNPM lockfile v6 scanning. This PR is a modified version of the fork created with the experimental detector setup added along with some minor changes.

Original PR description:
> This addresses #503 by doing the following:
> 
> * Explicitly mark code specific to version 5 as "V5"
> * Add new code for V6.
> * Add version detection logic to select which code to run.
> * Add tests for new code.
> 
> The old code wasn't clear about what file formats it actually worked with, and a lot of its tests seems to be from versions earlier than 5 which it didn't really parse correctly (ex: the version it parses isn't present in those versions as it had a different name). Thus to preserve existing behavior I pointed cases which failed to detect a version through the old V5 code.
> 
> The included verification tests didn't cover a lot of cases (link and file references, aliased packages, shared shrink-wrap format, disambiguated deps with peer deps) so I also tested on the package lock from https://github.com/microsoft/FluidFramework which is 44k lines and has all the mentioned edge cases. If desired, I can add the v5 and v6 versions of that file to the verification tests directory.
> 
> This should be in a working state but I need to actually understand how the verification tests results are verifies before this should be merged.
> 
> This is my first time working in this codebase, as well as the first time in several years that I'm working in C#, so please be extra careful reviewing this: don't trust that I know what I'm doing.


src/Microsoft.ComponentDetection.Detectors/pnpm/Contracts/PnpmHasDependenciesV6.cs
```diff
@@ -0,0 +1,14 @@
+namespace Microsoft.ComponentDetection.Detectors.Pnpm;
+
+using System.Collections.Generic;
+
+#pragma warning disable SA1300 // Used for deserialization and the process is case sensitive
```
JamieMagee, line 5: nit: fix the casing on these and use [`JsonPropertyNameAttribute`][1] (`System.Text.Json`) or [`JsonPropertyAttribute`][2] (`Newtonsoft.Json`) to override the deserialization,

[1]: https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonpropertynameattribute?view=net-8.0
[2]: https://www.newtonsoft.com/json/help/html/JsonPropertyName.htm

src/Microsoft.ComponentDetection.Detectors/pnpm/Pnpm6ComponentDetector.cs
```diff
@@ -0,0 +1,159 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.ComponentDetection.Contracts;
+using Microsoft.ComponentDetection.Contracts.Internal;
+using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using Microsoft.Extensions.Logging;
+
+public class Pnpm6ComponentDetector : FileComponentDetector, IExperimentalDetector
```
JamieMagee, line 13: There is a lot of overlap between this and the v5 detector. What would you say to, in the long run, a `PnpmComponentDetectorFactory` which implemented `FileComponentDetector`, and an `IPnpmDetector` interface. The `PnpmComponentDetectorFactory` should instantiate the correct version of the pnpm detector and pass it the already deserialized lockfile.

src/Microsoft.ComponentDetection.Detectors/pnpm/Pnpm6ComponentDetector.cs
```diff
@@ -0,0 +1,159 @@
...
+{
+    public Pnpm6ComponentDetector(
+        IComponentStreamEnumerableFactory componentStreamEnumerableFactory,
+        IObservableDirectoryWalkerFactory walkerFactory,
+        ILogger<Pnpm6ComponentDetector> logger)
+    {
+        this.ComponentStreamEnumerableFactory = componentStreamEnumerableFactory;
+        this.Scanner = walkerFactory;
+        this.NeedsAutomaticRootDependencyCalculation = true;
```
JamieMagee, line 22: nit: move this out of the constructor `public override bool NeedsAutomaticRootDependencyCalculation => true;`

src/Microsoft.ComponentDetection.Detectors/pnpm/Contracts/PnpmHasDependenciesV6.cs
cobya: Now deserialized with YamlMember Alias.

src/Microsoft.ComponentDetection.Detectors/pnpm/Pnpm6ComponentDetector.cs
```diff
@@ -0,0 +1,159 @@
...
+        this.ProcessDependencyList(singleFileComponentRecorder, components, item.devDependencies);
+        this.ProcessDependencyList(singleFileComponentRecorder, components, item.optionalDependencies);
+    }
+
+    private void ProcessDependencyList(ISingleFileComponentRecorder singleFileComponentRecorder, Dictionary<string, (DetectedComponent C, Package P)> components, Dictionary<string, PnpmYamlV6Dependency> dependencies)
+    {
+        foreach (var (name, dep) in dependencies ?? Enumerable.Empty<KeyValuePair<string, PnpmYamlV6Dependency>>())
+        {
+            // Ignore "file:" and "link:" as these are local packages.
+            if (dep.version.StartsWith("link:") | dep.version.StartsWith("file:"))
```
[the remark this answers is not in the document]
cobya: Good catch!

JamieMagee (approved):

PR #1145, opened 2024-05-31
Merged pnpm6 experiment into pnpm detector

pauld-msft:
`Pnpm6` has gone through a week or so of experimentation after being introduced with https://github.com/microsoft/component-detection/pull/1110 and has been validated to work as expected. 

This pull request:
- removes the experiment
- merges the pnpm6 changes into pnpm detector, using the suggested format from [this comment](https://github.com/microsoft/component-detection/pull/1110#discussion_r1602071025)
- increments Pnpm detector version from 5 -> 6


src/Microsoft.ComponentDetection.Detectors/pnpm/PnpmComponentDetectorFactory.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                this.Logger.LogWarning("Unsupported lockfileVersion in pnpm yaml file {File}", file.Location);
+            }
+            else
+            {
+                detector.RecordDependencyGraphFromFile(fileContent, singleFileComponentRecorder);
+            }
+        }
+        catch (Exception e)
+        {
```
cobya, line 66: can you add a `FailedParsingFileRecord` telemetry object here? PipReportComponentDetector has an example

src/Microsoft.ComponentDetection.Detectors/pnpm/PnpmComponentDetectorFactory.cs
```diff
@@ -0,0 +1,99 @@
...
+            this.Logger.LogDebug("Skipping found file, it was detected as being within a {SkippedFolder} folder.", skippedFolder);
+        }
+
+        try
+        {
+            var fileContent = await new StreamReader(file.Stream).ReadToEndAsync();
+            var detector = this.GetPnpmComponentDetector(fileContent, file.Location);
+            if (detector == null)
+            {
```
cobya, line 57: let's create a new telemetry item equivalent to `PipReportVersionTelemetryRecord` so we can track which versions aren't supported yet. I'm open to standardizing the telemetry object here as well so we can start applying it across our parsing logic

cobya (approved):
Changes look good to me 👍

src/Microsoft.ComponentDetection.Detectors/pnpm/PnpmComponentDetectorFactory.cs
pauld-msft: created abstract VersionTelemetryRecord that both of these can inherit from

cobya (approved):

PR #1163, opened 2024-06-07
Sanitize Docker environment vars in logs

grvillic:
## Context
We spotted some ephemeral credentials in our telemetry due to syft being executed in a container and returning all/most of its environment variables.

## Solution
Sanitize confidential information placed in URLs within env vars passed to Syft.

### DIFF
--- a/src/Microsoft.ComponentDetection.Common/Utilities/StringUtilities.cs
+++ b/src/Microsoft.ComponentDetection.Common/Utilities/StringUtilities.cs
@@ -5,7 +5,8 @@ using System.Text.RegularExpressions;
 
 public static class StringUtilities
 {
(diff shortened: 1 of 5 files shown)


src/Microsoft.ComponentDetection.Common/Utilities/StringUtilities.cs
```diff
@@ -5,7 +5,8 @@ namespace Microsoft.ComponentDetection.Common;
 
 public static class StringUtilities
 {
-    private static readonly Regex SensitiveInfoRegex = new Regex(@"(?<=https://)(.+)(?=@)", RegexOptions.Compiled, TimeSpan.FromSeconds(5));
+    private static readonly Regex SensitiveInfoRegex = new Regex(@"(?<=https://)(.+)(?=@)", RegexOptions.Compiled | RegexOptions.IgnoreCase, TimeSpan.FromSeconds(5));
```
grvillic, line 8: No evidence of uppercase urls, but it doesn't hurt to include it here.

FernandoRojo (approved):

PR #1166, opened 2024-06-10
Add env var that allows customers to skip pip report

pauld-msft:
Pip report causes customers to have their runs timeout. To immediately unblock customers, this adds the ability to not run pipreport detection. Investigating root cause in meantime

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
@@ -15,6 +15,8 @@ using Microsoft.Extensions.Logging;
 
 public class PipReportComponentDetector : FileComponentDetector, IExperimentalDetector
 {
+    private const string DisablePipReportScanEnvVar = "DisablePipReportScan";
+
     /// <summary>
     /// The maximum version of the report specification that this detector can handle.
     /// </summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
```diff
@@ -81,6 +86,19 @@ protected override async Task OnFileFoundAsync(ProcessRequest processRequest, ID
         FileInfo reportFile = null;
         try
         {
+            if (this.IsPipReportManuallyDisabled())
+            {
+                this.Logger.LogWarning("PipReport: The pip report has been manually disabled, fallback strategy performed.");
```
FernandoRojo, line 91: Nit: Comment is a bit off, since there is no fallback strategy, maybe just state that only the standard pip results will be collected?

FernandoRojo (approved):

src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
pauld-msft: going to just mention that the pip report is skipped, so that in the future when we deprecate existing pip result collection to help with network isolation, we will not need to update the warning message

PR #1171, opened 2024-06-12
Enable parallelism by detector

pauld-msft:
Adds some configurable parallelism to a single detector, e.g. within PipReport. 

Added some simple pip reports that take 3-5 seconds to generate.

Running with the `MaxDetectionThreads` set to 1:
image

Running without the `MaxDetectionThreads` (default of 10):
image

Finally, with `MaxDetectionThreads` set to 2:
image

Scan log from the 2 parallel threads is also interesting since you can see that 2 pip reports are triggered at the same time and the next set is triggered once these finish:
[GovCompDisc_Log_20240612152920140_41928.log](https://github.com/user-attachments/files/15811613/GovCompDisc_Log_20240612152920140_41928.log)

### DIFF
--- a/src/Microsoft.ComponentDetection.Contracts/ScanRequest.cs
+++ b/src/Microsoft.ComponentDetection.Contracts/ScanRequest.cs
@@ -19,7 +19,7 @@ public class ScanRequest
     /// <param name="imagesToScan">Container images to scan.</param>
     /// <param name="componentRecorder">Detector component recorder.</param>
     /// <param name="maxThreads">Max number of threads to use for detection.</param>
-    public ScanRequest(DirectoryInfo sourceDirectory, ExcludeDirectoryPredicate directoryExclusionPredicate, ILogger logger, IDictionary<string, string> detectorArgs, IEnumerable<string> imagesToScan, IComponentRecorder componentRecorder, int maxThreads = 10)
+    public ScanRequest(DirectoryInfo sourceDirectory, ExcludeDirectoryPredicate directoryExclusionPredicate, ILogger logger, IDictionary<string, string> detectorArgs, IEnumerable<string> imagesToScan, IComponentRecorder componentRecorder, int maxThreads = 3)
     {
         this.SourceDirectory = sourceDirectory;
         this.DirectoryExclusionPredicate = directoryExclusionPredicate;
--- a/src/Microsoft.ComponentDetection.Contracts/ScanRequest.cs
+++ b/src/Microsoft.ComponentDetection.Contracts/ScanRequest.cs
@@ -18,13 +18,15 @@ public class ScanRequest
     /// <param name="detectorArgs">A dictionary of custom detector arguments supplied externally.</param>
     /// <param name="imagesToScan">Container images to scan.</param>
     /// <param name="componentRecorder">Detector component recorder.</param>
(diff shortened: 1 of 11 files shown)


src/Microsoft.ComponentDetection.Orchestrator/Services/DetectorProcessingService.cs
```diff
@@ -23,6 +23,8 @@ namespace Microsoft.ComponentDetection.Orchestrator.Services;
 
 public class DetectorProcessingService : IDetectorProcessingService
 {
+    private const int DefaultMaxDetectionThreads = 10;
```
cobya, line 26: We should probably be a bit more conservative with the default here - maybe 3-5

cobya: Since some of the detector processes are reaching out to APIs for artifact information, we don't want to have them get throttled for too many concurrent requests.

pauld-msft: makes sense, updated

src/Microsoft.ComponentDetection.Contracts/ScanRequest.cs
```diff
@@ -18,13 +18,15 @@ public class ScanRequest
     /// <param name="detectorArgs">A dictionary of custom detector arguments supplied externally.</param>
     /// <param name="imagesToScan">Container images to scan.</param>
     /// <param name="componentRecorder">Detector component recorder.</param>
-    public ScanRequest(DirectoryInfo sourceDirectory, ExcludeDirectoryPredicate directoryExclusionPredicate, ILogger logger, IDictionary<string, string> detectorArgs, IEnumerable<string> imagesToScan, IComponentRecorder componentRecorder)
+    /// <param name="maxThreads">Max number of threads to use for detection.</param>
+    public ScanRequest(DirectoryInfo sourceDirectory, ExcludeDirectoryPredicate directoryExclusionPredicate, ILogger logger, IDictionary<string, string> detectorArgs, IEnumerable<string> imagesToScan, IComponentRecorder componentRecorder, int maxThreads = 10)
```
cobya, line 22: Default max threads here is still 10

cobya (dismissed):

src/Microsoft.ComponentDetection.Contracts/ScanRequest.cs
cobya, line 22:
```suggestion
    public ScanRequest(DirectoryInfo sourceDirectory, ExcludeDirectoryPredicate directoryExclusionPredicate, ILogger logger, IDictionary<string, string> detectorArgs, IEnumerable<string> imagesToScan, IComponentRecorder componentRecorder, int maxThreads = 3)
```

cobya:
Looks like there are some parallelization issues with the GoWithReplace detector - can you take a look?

pauld-msft:
> Looks like there are some parallelization issues with the GoWithReplace detector - can you take a look?

Yep, seems like go w/ replace was using a class level list of components, which was re-used by the different threads. Changed this up to initialize the component list as the file found method is invoked, which should make it thread safe. Also did a quick spot check and could not find similar issues with the other detectors

FernandoRojo (approved):

PR #1176, opened 2024-06-14
Go With Replace Detector

amitla1:
Changing regular go detector to the experimental go with replace detector


cobya:
@amitla1do we still need this with #1181 being merged?

PR #1181, opened 2024-06-18
Go With Replace Detector no File Parsing

amitla1:
Remove the file parsing for the replace section of the file
The current go detector and the new go detector that has the replace module sometimes detect a different number of components and they should be the same number. The file parsing logic that was changed in the new go detector couldve impacted this therefore reverting that change and just implementing replace logic for CLI and not parsing a file

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/go/GoComponentWithReplaceDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/go/GoComponentWithReplaceDetector.cs
@@ -15,7 +15,6 @@ using Microsoft.ComponentDetection.Contracts;
 using Microsoft.ComponentDetection.Contracts.Internal;
 using Microsoft.ComponentDetection.Contracts.TypedComponent;
 using Microsoft.Extensions.Logging;
-using MoreLinq;
 using Newtonsoft.Json;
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.ComponentDetection.Detectors/go/GoComponentWithReplaceDetector.cs
```diff
@@ -277,90 +276,41 @@ private void AddGoComponent(IList<GoComponent> goComponents, string line, ISingl
...
-                this.AddGoComponent(goComponents, line, singleFileComponentRecorder);
-            }
-            else if (inReplaceBlock)
-            {
-                this.ReplaceGoComponents(goComponents, line, singleFileComponentRecorder);
+                // In go >= 1.17, direct dependencies are listed as "require x/y v1.2.3", and transitive dependencies
+                // are listed in the require () section
+                if (line.StartsWith("require "))
+                {
+                    this.TryRegisterDependencyFromModLine(line[8..], singleFileComponentRecorder);
```
RushabhBhansali, line 302: 8 is a [magic number](https://www.baeldung.com/cs/antipatterns-magic-numbers#magic-numbers). I suggest replacing it with `length("require "

RushabhBhansali (approved):

PR #1183, opened 2024-06-20
Pauldorsch/reconcile dependency graph logic

pauld-msft:
This PR has a few improvements:

- reconciles the dependency graph creation between the 2 pip detectors. They now use a shared method for determining valid parents. E.g. we were ignoring `extra` requirements for the legacy pip, but not for PipReport, but now we treat both the same.
- legacy detection was dropping packages that had non lowercase letters in their METADATA file path, so added case insensitive fallback logic
- handle null or empty conditional env vars gracefully by allowing the package through rather than erroring
- noticed some local parallelization issues, so only enabling for pip report for right now

The `test/Microsoft.ComponentDetection.VerificationTests/resources/pip/roots/requirements.txt` file added in this PR has 30 diffs in main at the moment, all in the AddedRootId section of the experiment results. Now, it has 0 diffs.


pauld-msft:
Verify snapshots are failing due to new validation test cases being added. However, if you see the checks that ran for `d70d94072e9a452b0cf5d294acfb76df025a2fa0` you can view the actual diffs, which just involve changing the dependency graph as expected

src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
```diff
@@ -75,6 +84,26 @@ protected override async Task<IObservable<ProcessRequest>> OnPrepareDetectionAsy
...
+            this.Logger.LogInformation($"No python found on system. Python detection will not run.");
+
+            return Enumerable.Empty<ProcessRequest>().ToObservable();
+        }
+        else
+        {
+            var pythonVersion = await this.pythonCommandService.GetPythonVersionAsync(pythonExePath);
+            this.pythonResolver.SetPythonEnvironmentVariable("python_version", pythonVersion);
+
+            var pythonPlatformString = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
```
cobya, line 99: How are we resolving MacOS?

src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
```diff
@@ -200,7 +229,7 @@ protected override async Task OnFileFoundAsync(ProcessRequest processRequest, ID
                 // futures; python_version <= \"2.7\"
                 // sphinx (!=1.8.0,!=3.1.0,!=3.1.1,>=1.6.5) ; extra == 'docs'
                 var dependencySpec = new PipDependencySpecification($"Requires-Dist: {dependency}", requiresDist: true);
-                if (dependencySpec.PackageIsUnsafe())
+                if (!dependencySpec.IsValidParentPackage(this.pythonResolver.GetPythonEnvironmentVariables()))
```
cobya, line 232: Can we avoid calling `this.pythonResolver.GetPythonEnvironmentVariables()` every iteration?

src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
```diff
@@ -55,6 +62,8 @@ public class PipReportComponentDetector : FileComponentDetector, IExperimentalDe
 
     public override int Version { get; } = 2;
```
cobya, line 63: Since we're changing the logic a bit for Pip and PipReport let's bump detector versions

src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
pauld-msft: updating to use actual python script for this: https://docs.python.org/3/library/sys.html#sys.platform

src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
```diff
@@ -198,11 +193,16 @@ protected override async Task OnFileFoundAsync(ProcessRequest processRequest, ID
         // graph ourselves using the requires_dist field.
         var dependenciesByPkg = new Dictionary<string, List<PipDependencySpecification>>(StringComparer.OrdinalIgnoreCase);
         var nodeReferences = new Dictionary<string, PipReportGraphNode>(StringComparer.OrdinalIgnoreCase);
+        var pythonEnvVars = this.pythonResolver.GetPythonEnvironmentVariables();
 
         foreach (var package in report.InstallItems)
         {
             // Normalize the package name to ensure consistency between the package name and the graph nodes.
             var normalizedPkgName = PipReportUtilities.NormalizePackageNameFormat(package.Metadata.Name);
+            if (PipDependencySpecification.PackagesToIgnore.Contains(normalizedPkgName))
```
cobya, line 202: no need to make this public - you can call the existing wrapper `PipDependencySpecification.PackageIsUnsafe`

cobya (approved):

src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
pauld-msft: we haven't created a `PipDependencySpecification` though, and we don't until we are iterating through the dependencies we already added to the node references list. I could create this object from the info we have here, but this would be slower since there is some regex / other logic that happens within the [constructor](https://github.com/microsoft/component-detection/blob/2284e06a29901ca60ba6d4bcf8b694ed459c2a63/src/Microsoft.ComponentDetection.Detectors/pip/Contracts/PipDependencySpecification.cs#L60-L116). Going to leave as-is for now, but happy to switch this up later on if you feel strongly :)

PR #1190, opened 2024-07-02
PipReport option to fallback and parse roots from source

pauld-msft:
PipReportOverrideBehavior that allows us to skip pip report or fallback to scanning source files directly without reaching out to remote package repos for the full dependency graph.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/pip/Contracts/PipDependencySpecification.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/pip/Contracts/PipDependencySpecification.cs
@@ -160,7 +160,7 @@ public class PipDependencySpecification
                 continue; // If the variable isn't in the environment, we can't evaluate it.
             }
 
-            if (string.Equals(conditionalVar, "python_version", System.StringComparison.OrdinalIgnoreCase))
+            if (string.Equals(conditionalVar, "python_version", StringComparison.OrdinalIgnoreCase))
             {
                 var pythonVersion = PythonVersion.Create(conditionalValue);
                 if (pythonVersion.Valid)
@@ -173,10 +173,10 @@ public class PipDependencySpecification
                     conditionMet = pythonEnvironmentVariables[conditionalVar] == conditionalValue;
                 }
             }
-            else if (string.Equals(conditionalVar, "sys_platform", System.StringComparison.OrdinalIgnoreCase))
+            else if (string.Equals(conditionalVar, "sys_platform", StringComparison.OrdinalIgnoreCase))
             {
                 // if the platform is not windows or linux (empty string in env var), allow the package to be added. Otherwise, ensure it matches the python condition
-                conditionMet = string.Equals(pythonEnvironmentVariables[conditionalVar], conditionalValue, System.StringComparison.OrdinalIgnoreCase);
+                conditionMet = string.Equals(pythonEnvironmentVariables[conditionalVar], conditionalValue, StringComparison.OrdinalIgnoreCase);
             }
             else
             {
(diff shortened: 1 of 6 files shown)


src/Microsoft.ComponentDetection.Detectors/pip/Contracts/PipDependencySpecification.cs
```diff
@@ -197,6 +197,33 @@ public bool PackageConditionsMet(Dictionary<string, string> pythonEnvironmentVar
         return conditionsMet;
     }
 
+    /// <summary>
+    /// Iterates through the package versions that are explicitly stated, and returns
+    /// the highest version that adheres to the version requirements.
+    /// </summary>
+    /// <example>
+    /// DependencySpecifiers: (&gt;=1.2.3, !=1.2.4, &lt;2.0.0)
```
pauld-msft, line 205: Thoughts on whether this should just detect for `==` and `===`, instead of grabbing package versions that match with `>=` and `<=` as well?

src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
```diff
@@ -308,6 +339,54 @@ protected override async Task OnFileFoundAsync(ProcessRequest processRequest, ID
...
+            .Where(x => !x.PackageIsUnsafe())
+            .Where(x => x.PackageConditionsMet(this.pythonResolver.GetPythonEnvironmentVariables()))
+            .ToList();
+
+        listedPackage.Select(x => (x.Name, Version: x.GetHighestExplicitPackageVersion()))
+            .Where(x => !string.IsNullOrEmpty(x.Version))
+            .Select(x => new PipComponent(x.Name, x.Version))
+            .Select(x => new DetectedComponent(x))
+            .ToList()
+            .ForEach(pipComponent => recorder.RegisterUsage(pipComponent, isExplicitReferencedDependency: true));
```
pauld-msft, line 366: Note: this is just registering the usage of these components, not running the root resolution since that logic uses metadata from pypi. 

Not sure if there is anything else we can glean from the files themselves to record in the `ISingleFileComponentRecorder`?

src/Microsoft.ComponentDetection.Detectors/pip/Contracts/PipDependencySpecification.cs
grvillic: Mind elaborating a bit more on the difference here? I would expect it the package version range to be more precise than just detecting explicit references, but I may be missing more context here.

pauld-msft: Sure, so if the requirements have an explicit `==` or `===`, then we know that the pip installation would pull that package. If it has `>=` by contrast, it is likely that the actual pip installation will pull the latest version from the feed, which may not be the version stated in the requirements file. For example:

take `>=1.2.3, !=1.2.4, <2.0.0` from this doc. We can't reach out to the feed itself to determine any versions, so we only have these 3 versions to go from. The only matching one is `1.2.3`, BUT the feed may have a `1.3.0` for all we know. 

Basically I think it boils down to whether we want to include versions that _may_ be resolved by a pip install, or only the ones we _know_ will be.

grvillic: Ah I follow now. I think it makes sense to follow the matching version, which would be 1.2.3 in your example. 

If customer wants to be explicit we can drop verbose logs that package was added due to naive detection strategy and explicit version range consumption policy. If they truly aren't needing it anymore, they can update the manifest accordingly.

grvillic (approved):

PR #1201, opened 2024-07-12
PipReport back to experimental, add pre-generated PipReport parsing

pauld-msft:
also add 
- some new env vars to control pip report behavior
- introduce fast deps by default
- add ability to consume pre-generated report files


src/Microsoft.ComponentDetection.Detectors/pip/PipCommandService.cs
```diff
@@ -127,6 +129,11 @@ public async Task<(PipInstallationReport Report, FileInfo ReportFile)> GenerateI
             pipReportCommand += $" --index-url {this.environmentService.GetEnvironmentVariable("PIP_INDEX_URL")}";
         }
 
+        if (!this.environmentService.DoesEnvironmentVariableExist(PipReportDisableFastDepsEnvVar) || !this.environmentService.IsEnvironmentVariableValueTrue(PipReportDisableFastDepsEnvVar))
```
cobya, line 132:
```suggestion
        if (!this.environmentService.DoesEnvironmentVariableExist(PipReportDisableFastDepsEnvVar) || this.environmentService.IsEnvironmentVariableValueTrue(PipReportDisableFastDepsEnvVar))
```
Wouldn't the current logic be disabling if set to false?

src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
```diff
@@ -150,44 +158,103 @@ protected override async Task OnFileFoundAsync(ProcessRequest processRequest, ID
...
 
-                using var versionRecord = new InvalidParseVersionTelemetryRecord
+                foreach (var existingReport in preGeneratedReportFiles)
                 {
-                    DetectorId = this.Id,
-                    FilePath = file.Location,
-                    Version = report.Version,
-                    MaxVersion = MaxReportVersion.ToString(),
-                };
+                    this.Logger.LogInformation("PipReport: Using pre-generated pip report {File}.", existingReport.FullName);
```
cobya, line 190: nit: can we add a note to this log to say which file it was going to scan instead?

src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
```diff
@@ -150,44 +158,103 @@ protected override async Task OnFileFoundAsync(ProcessRequest processRequest, ID
...
-                    report.Version,
-                    MaxReportVersion);
+                preGeneratedReportFiles = PipReportPreGeneratedFilePatterns
+                    .SelectMany(pattern => fileParentDirectoryInfo.GetFiles(pattern))
+                    .Where(file => File.Exists(file.FullName))
+                    .ToList();
+            }
+
+            List<PipInstallationReport> reports = new();
+            if (preGeneratedReportFiles is not null && preGeneratedReportFiles.Any())
```
cobya, line 184: How does this handle setup.py and requirements.txt in the same folder? Does it parse the files twice?

cobya (approved):
Changes look good to me, just one comment on the new parsing logic.

PR #1203, opened 2024-07-17
Promote VCPKG detector to enabled by default

grvillic:
## Summary
We have been running VCPKG detector for more than a year and after looking at its telemetry we have seen promising results and detection based on customer feedback.

Promoting to PROD detector.


pauld-msft (approved):
LGTM as long as detector version doesn't need to be updated

PR #1205, opened 2024-07-19
Pauldorsch/bugfix invalid pipreport files

pauld-msft:
Add validation to pre-generated pip reports to confirm that they contain the correct set of Requested dependencies.

### Sample problem
```
pip install -r requirements.txt --no-input --no-compile --report component-detection-pip-report.json
pip install pyOpenSSL==23.0.0 --no-input --no-compile --report component-detection-pip-report.json
```
When a customer has multiple reports generated with same name in same directory, it will be overridden. We do not want to use the report from the above commands instead of running detection on `requirements.txt`.

### Solution
Ensure that the list of packages explicitly found in the `requirements.txt` or `setup.py` files is a subset of the `Requested` packages in the report.

### Other improvements
Add telemetry to tell us when a pre-generated report was used vs. generating the report


src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
```diff
@@ -442,6 +464,27 @@ protected override async Task OnFileFoundAsync(ProcessRequest processRequest, ID
...
+    /// in the pip file. This prevents invalid reports from being used to create the dependency graph.
+    /// </summary>
+    private async Task<bool> IsValidPreGeneratedReportAsync(PipInstallationReport report, string pythonExePath, string filePath)
+    {
+        var initialPackages = await this.pythonCommandService.ParseFileAsync(filePath, pythonExePath);
+        var listedPackage = initialPackages.Where(tuple => tuple.PackageString != null)
+            .Select(tuple => tuple.PackageString)
+            .Where(x => !string.IsNullOrWhiteSpace(x))
+            .Select(x => new PipDependencySpecification(x))
+            .Where(x => !x.PackageIsUnsafe())
```
cobya, line 478: We should probably move this filtering logic into a util class, we are doing it in a few places now

cobya (approved):

PR #1213, opened 2024-07-30
Add logs to MvnCLI and use dictionaries to improve perf on large repos

grvillic:
## Context
MvnCLI can hang when scanning directories with significant amount of pom.xmls. At the moment it is hard to tell if this is bug in the detector itself or Mvn CLI APIs.

## Solution
We are adding logs to be able to improve troubleshooting the detector, and we are also introducing a build variable called `MvnCLIFileLevelTimeoutSeconds` that will allow users restrict max time execution per each processed file. Screenshot below on how the logs when the build variable sets a very low threshold, and the tool cancels the command's execution.

image

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
@@ -73,7 +73,7 @@ public class MavenCommandService : IMavenCommandService
                 processRequest.SingleFileComponentRecorder.RegisterPackageParseFailure(pomFile.Location);
             }
 
-            if (cliFileTimeout.IsCancellationRequested)
+            if (timeoutSeconds != -1 && cliFileTimeout.IsCancellationRequested)
             {
                 this.logger.LogWarning("{DetectorPrefix}: There was a timeout in {PomFileLocation} file. Increase it using {TimeoutVar} environment variable.", DetectorLogPrefix, pomFile.Location, MvnCLIFileLevelTimeoutSecondsEnvVar);
             }
--- a/src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
@@ -73,9 +73,9 @@ public class MavenCommandService : IMavenCommandService
                 processRequest.SingleFileComponentRecorder.RegisterPackageParseFailure(pomFile.Location);
             }
 
-            if (timeoutSeconds != -1)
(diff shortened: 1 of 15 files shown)


src/Microsoft.ComponentDetection.Detectors/maven/MvnCliComponentDetector.cs
```diff
@@ -33,19 +36,24 @@ public class MvnCliComponentDetector : FileComponentDetector
 
     public override string Id => "MvnCli";
 
-    public override IList<string> SearchPatterns => new List<string> { "pom.xml" };
+    public override IList<string> SearchPatterns => new List<string> { MavenManifest };
 
     public override IEnumerable<ComponentType> SupportedComponentTypes => new[] { ComponentType.Maven };
 
-    public override int Version => 3;
+    public override int Version => 4;
```
grvillic, line 43: No actual dependency graph changes but bumped it to identify when the fix was deployed, in case there is any unexpected regression.

pauld-msft (approved):

src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
```diff
@@ -36,22 +42,45 @@ public async Task<bool> MavenCLIExistsAsync()
...
+            var errorMessage = string.IsNullOrWhiteSpace(result.StdErr) ? result.StdOut : result.StdErr;
+            var isErrorMessagePopulated = !string.IsNullOrWhiteSpace(errorMessage);
+
+            if (isErrorMessagePopulated)
+            {
+                this.logger.LogError("Mvn output: {MvnStdErr}", errorMessage);
+                processRequest.SingleFileComponentRecorder.RegisterPackageParseFailure(pomFile.Location);
+            }
+
+            if (timeoutSeconds != -1)
```
pauld-msft, line 76: This just indicates that the timeout was set and exist code isn't 0. I think we could potentially log this when a non timeout error occurs while the timeout was still set. 

So we may want to additionally check that `cancellationToken.IsCancellationRequested` and `result.ExitCode == -1` (from my testing this appears to be the exit code when we kill the process)

pauld-msft (approved):

src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
grvillic: Technically the timeouts can only happen if that is set, but looking at `cancellationToken.IsCancellationRequested` seems like a good way to ensure we print only when we cancel the execution.

pauld-msft (approved):

PR #1224, opened 2024-08-12
Add support to persist pip reports

pauld-msft:
The primary use case here is when the same build stage has both Component Detection and SBOM (where CD runs again). This will allow whichever comes second to be able to just use the previous runs' reports, rather than regenerating them.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/pip/PipCommandService.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/pip/PipCommandService.cs
@@ -131,7 +131,7 @@ public class PipCommandService : IPipCommandService
         var workingDir = new DirectoryInfo(this.pathUtilityService.GetParentDirectory(formattedPath));
 
         CommandLineExecutionResult command;
-        var reportName = Path.GetRandomFileName();
+        var reportName = Path.GetFileNameWithoutExtension(Path.GetRandomFileName()) + ".component-detection-pip-report.json";
         var reportFile = new FileInfo(Path.Combine(workingDir.FullName, reportName));
 
         string pipReportCommand;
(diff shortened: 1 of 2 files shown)


src/Microsoft.ComponentDetection.Detectors/pip/PipCommandService.cs
```diff
@@ -131,7 +131,7 @@ private async Task<bool> CanCommandBeLocatedAsync(string pipPath)
         var workingDir = new DirectoryInfo(this.pathUtilityService.GetParentDirectory(formattedPath));
 
         CommandLineExecutionResult command;
-        var reportName = Path.GetRandomFileName();
+        var reportName = Path.GetFileNameWithoutExtension(Path.GetRandomFileName()) + ".component-detection-pip-report.json";
```
grvillic, line 134: Do we need to change the API here, why not just `$"{Path.GetRandomFileName().component-detection-pip-report}"`, since we are not using the `.json` extension piece we don't need to worry about it, right?

src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
```diff
@@ -300,11 +301,14 @@ protected override async Task OnFileFoundAsync(ProcessRequest processRequest, ID
         finally
         {
             // Clean up the report output JSON file so it isn't left on the machine.
-            foreach (var reportFile in reportFiles)
+            if (!this.envVarService.IsEnvironmentVariableValueTrue(PipReportPersistReportsEnvVar) && reportFiles.Any())
```
grvillic, line 304: nit: if reportFiles is empty, the line below will skip it, namely `reportFiles.Any()` seems a bit redundant.

src/Microsoft.ComponentDetection.Detectors/pip/PipCommandService.cs
pauld-msft: we do use the `json` portion when searching for pre-generated files: https://github.com/microsoft/component-detection/blob/main/src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs#L26 

this update allows us to match  `*.component-detection-pip-report.json`

grvillic (approved):

PR #1232, opened 2024-08-20
Pauldorsch/fix invalid version bug

pauld-msft:
Fixes a bug seen when using the default env vars. The fallback behavior throws an ArgumentException for invalid versions, e.g. `"info", "notrealversion", ...` and we don't handle it gracefully. This catches all exceptions that might be thrown, but also handles the invalid exceptions as we actually expect it to.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/pip/Contracts/PipDependencySpecification.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/pip/Contracts/PipDependencySpecification.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Text.RegularExpressions;
+using Microsoft.Extensions.Logging;
 
 /// <summary>
 /// Represents a package and a list of dependency specifications that the package must be.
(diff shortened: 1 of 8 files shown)


src/Microsoft.ComponentDetection.Detectors/pip/Contracts/PipDependencySpecification.cs
```diff
@@ -166,7 +166,14 @@ public bool PackageConditionsMet(Dictionary<string, string> pythonEnvironmentVar
...
                 {
                     var conditionalSpec = $"{conditionalOperator}{conditionalValue}";
-                    conditionMet = PythonVersionUtilities.VersionValidForSpec(pythonEnvironmentVariables[conditionalVar], new List<string> { conditionalSpec });
+                    try
+                    {
+                        conditionMet = PythonVersionUtilities.VersionValidForSpec(pythonEnvironmentVariables[conditionalVar], new List<string> { conditionalSpec });
+                    }
+                    catch (ArgumentException)
+                    {
+                        conditionMet = false;
```
cobya, line 175: can we use a LogDebug statement here for attempting to diagnose issues in the future?

src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
```diff
@@ -294,8 +294,15 @@ protected override async Task OnFileFoundAsync(ProcessRequest processRequest, ID
             // if pipreport fails, try to at least list the dependencies that are found in the source files
             if (this.GetPipReportOverrideBehavior() != PipReportOverrideBehavior.SourceCodeScan && !this.PipReportSkipFallbackOnFailure())
             {
-                this.Logger.LogInformation("PipReport: Trying to Manually compile dependency list for '{File}' without reaching out to a remote feed.", file.Location);
-                await this.RegisterExplicitComponentsInFileAsync(singleFileComponentRecorder, file.Location, pythonExePath);
+                try
+                {
+                    this.Logger.LogInformation("PipReport: Trying to Manually compile dependency list for '{File}' without reaching out to a remote feed.", file.Location);
```
cobya, line 299: we should make a note here and in the docs that if users see this message, python detection essentially isn't happening

cobya (approved):

src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
pauld-msft: made the note here, and it is already in the pip report TSG doc. Our wrapper makes it so that we do not fallback to this logic, so internal users should not hit this

Issue #771, opened 2023-09-06
Switch to .NET 8 RC


melotic:

melotic:
.NET 8.0.100-rc.1 was just released yesterday, and is a go-live release: https://dotnet.microsoft.com/en-us/download/dotnet/8.0

cobya:
.NET release 8.0.0 was released on November 14th, 2023. See (8.0.0 release notes)[https://github.com/dotnet/core/blob/main/release-notes/8.0/8.0.0/8.0.0.md] for more details.

cobya:
There is an ask to support .NET 8 for ARM64 for OSX in the sbom-tool via https://github.com/microsoft/dropvalidator/issues/848

PR #1249, opened 2024-09-16
Update Component Detection to .NET v8

cobya:
Update Component Detection to .NET v8, the latest LTS branch of .NET Core. See #771 for more information.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/npm/Contracts/PackageLockV1.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/npm/Contracts/PackageLockV1.cs
@@ -1,4 +1,4 @@
-﻿#nullable enable
+#nullable enable
 namespace Microsoft.ComponentDetection.Detectors.Npm.Contracts;
 
 using System.Collections.Generic;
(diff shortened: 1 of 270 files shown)


cobya:
Dropping the SDK version back to 8.0.303 to avoid https://github.com/NuGet/Home/issues/13729

src/Microsoft.ComponentDetection.Detectors/npm/Contracts/PackageLockV1.cs
```diff
@@ -1,4 +1,4 @@
-﻿#nullable enable
+#nullable enable
```
pauld-msft, line 1: Not super relevant to this pull request, but is there a reason that we don't have nullable configured at our project level with `<Nullable>enable</Nullable>`?

test/Microsoft.ComponentDetection.Common.Tests/BaseDetectionTelemetryRecordTests.cs
```diff
@@ -73,7 +68,7 @@ public void SerializableProperties()
                 if (!serializableTypes.Contains(property.PropertyType) &&
                     Attribute.GetCustomAttribute(property.PropertyType, typeof(DataContractAttribute)) == null)
```
pauld-msft, line 69: dont need this second check anymore, doing this in the assertion below

pauld-msft (approved):

src/Microsoft.ComponentDetection.Detectors/npm/Contracts/PackageLockV1.cs
cobya: We probably could - I'm not aware of any reasoning behind why we wouldn't allow it, but agree that this is out of scope for this PR

PR #1250, opened 2024-09-19
add MSBuild binary log (.binlog) component detector

brettfo:
Adds a new NuGet component detector to scan MSBuild binary log (`*.binlog`) files.

There was one major reason to add this: the current NuGet component detector that scans `project.assets.json` isn't always 100% accurate.  This can happen when a given NuGet package has a transitive dependency on a `System.*` package, but the installed SDK has a newer version of that file that is replaced at build time.

One common example: a project has `<PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="6.0.0" />`.  That package has a transitive dependency on `System.Text.Json/6.0.0`.  No matter the SDK used to restore and build the project, the `project.assets.json` file will report that the package `System.Text.Json/6.0.0` was used and implies that it'll be present at runtime.  However, if the `6.0.424` SDK or newer was used to build, the `System.Text.Json.dll` file from the `6.0.0` package is removed and replaced with a newer `System.Text.Json.dll` from the SDK's runtime package.

To avoid reporting a false positive for a package that was replaced at build time (and therefore won't be present at runtime) the MSBuild binary log scanner was added.

If a binary log is generated with the `/bl` flag, every MSBuild event will be recorded.  The new detector in this PR scans each event for the relevant `AddItem` and `RemoveItem` elements (corresponding to things like `<SomeItemGroup Include="..." ... />` and `<SomeItemGroup Remove="..." ... />` and uses that to build a 100% accurate dependency set.

Most of the content of this PR is to support the tests to make them as reliable as possible through 2 different means:

1. Generate a fresh `.binlog` file for each test by running the appropriate `dotnet build ...` command.  This ensures there are no large binary test assets added to this repo; they're simply generated as needed.
2. Mock all necessary NuGet packages to prevent network access.  To accomplish this, 3 common NuGet packages are faked for the tests: `Microsoft.NETCore.App.Ref`, `Microsoft.AspNetCore.App.Ref`, and `Microsoft.WindowsDesktop.App.Ref`.

### Future work

To fully prevent reporting false positives, it would be a good idea to merge the binary log detector and the `project.assets.json` detector so that if a binary log is present and covers a given `.csproj`, _only_ the binary log detector is used and fall back to the original detector otherwise.  Without this work, the binary log detector will properly **not** report `System.Text.Json/6.0.0`, but the `project.assets.json` detector would, so the false positive will still appear.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -21,9 +21,12 @@
     <PackageVersion Include="MinVer" Version="5.0.0" />
     <PackageVersion Include="Moq" Version="4.18.4" />
     <PackageVersion Include="morelinq" Version="4.2.0" />
+    <PackageVersion Include="MSBuild.StructuredLogger" Version="2.2.317" />
     <PackageVersion Include="MSTest.TestFramework" Version="3.5.1" />
     <PackageVersion Include="MSTest.Analyzers" Version="3.5.1" />
     <PackageVersion Include="MSTest.TestAdapter" Version="3.5.1" />
+    <PackageVersion Include="Microsoft.Build.Framework" Version="17.5.0" />
+    <PackageVersion Include="Microsoft.Build.Locator" Version="1.6.1" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.10.0" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageVersion Include="Newtonsoft.Json.Schema" Version="3.0.16" />
--- a/src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
@@ -1,0 +1,212 @@
+﻿namespace Microsoft.ComponentDetection.Detectors.NuGet;
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using Microsoft.Build.Locator;
+using Microsoft.Build.Logging.StructuredLogger;
+using Microsoft.ComponentDetection.Contracts;
+using Microsoft.ComponentDetection.Contracts.Internal;
+using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using Microsoft.Extensions.Logging;
+
+using Task = System.Threading.Tasks.Task;
+
+public class NuGetMSBuildBinaryLogComponentDetector : FileComponentDetector
+{
+    private static readonly HashSet<string> TopLevelPackageItemNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+    {
+        "PackageReference",
+    };
+
+    // the items listed below represent collection names that NuGet will resolve a package into, along with the metadata value names to get the package name and version
+    private static readonly Dictionary<string, (string NameMetadata, string VersionMetadata)> ResolvedPackageItemNames = new Dictionary<string, (string, string)>(StringComparer.OrdinalIgnoreCase)
+    {
+        ["NativeCopyLocalItems"] = ("NuGetPackageId", "NuGetPackageVersion"),
+        ["ResourceCopyLocalItems"] = ("NuGetPackageId", "NuGetPackageVersion"),
+        ["RuntimeCopyLocalItems"] = ("NuGetPackageId", "NuGetPackageVersion"),
+        ["ResolvedAnalyzers"] = ("NuGetPackageId", "NuGetPackageVersion"),
+        ["_PackageDependenciesDesignTime"] = ("Name", "Version"),
+    };
+
+    private static bool isMSBuildRegistered;
+
+    public NuGetMSBuildBinaryLogComponentDetector(
+        IObservableDirectoryWalkerFactory walkerFactory,
+        ILogger<NuGetMSBuildBinaryLogComponentDetector> logger)
+    {
+        this.Scanner = walkerFactory;
+        this.Logger = logger;
+    }
+
+    public override string Id { get; } = "NuGetMSBuildBinaryLog";
+
+    public override IEnumerable<string> Categories => new[] { Enum.GetName(typeof(DetectorClass), DetectorClass.NuGet) };
+
+    public override IList<string> SearchPatterns { get; } = new List<string> { "*.binlog" };
+
+    public override IEnumerable<ComponentType> SupportedComponentTypes { get; } = new[] { ComponentType.NuGet };
+
+    public override int Version { get; } = 1;
+
+    private static void ProcessResolvedPackageReference(Dictionary<string, HashSet<string>> topLevelDependencies, Dictionary<string, Dictionary<string, string>> projectResolvedDependencies, NamedNode node)
+    {
+        var doRemoveOperation = node is RemoveItem;
+        var doAddOperation = node is AddItem;
+        if (TopLevelPackageItemNames.Contains(node.Name))
+        {
+            var projectEvaluation = node.GetNearestParent<ProjectEvaluation>();
+            if (projectEvaluation is not null)
+            {
+                foreach (var child in node.Children.OfType<Item>())
+                {
+                    var packageName = child.Name;
+                    if (!topLevelDependencies.TryGetValue(projectEvaluation.ProjectFile, out var topLevel))
+                    {
+                        topLevel = new(StringComparer.OrdinalIgnoreCase);
+                        topLevelDependencies[projectEvaluation.ProjectFile] = topLevel;
+                    }
+
+                    if (doRemoveOperation)
+                    {
+                        topLevel.Remove(packageName);
+                    }
+
+                    if (doAddOperation)
+                    {
+                        topLevel.Add(packageName);
+                    }
+                }
+            }
+        }
+        else if (ResolvedPackageItemNames.TryGetValue(node.Name, out var metadataNames))
+        {
+            var nameMetadata = metadataNames.NameMetadata;
+            var versionMetadata = metadataNames.VersionMetadata;
+            var originalProject = node.GetNearestParent<Project>();
+            if (originalProject is not null)
+            {
+                foreach (var child in node.Children.OfType<Item>())
+                {
+                    var packageName = GetChildMetadataValue(child, nameMetadata);
+                    var packageVersion = GetChildMetadataValue(child, versionMetadata);
+                    if (packageName is not null && packageVersion is not null)
+                    {
+                        var project = originalProject;
+                        while (project is not null)
+                        {
+                            if (!projectResolvedDependencies.TryGetValue(project.ProjectFile, out var projectDependencies))
+                            {
+                                projectDependencies = new(StringComparer.OrdinalIgnoreCase);
+                                projectResolvedDependencies[project.ProjectFile] = projectDependencies;
+                            }
+
+                            if (doRemoveOperation)
+                            {
+                                projectDependencies.Remove(packageName);
+                            }
+
+                            if (doAddOperation)
+                            {
+                                projectDependencies[packageName] = packageVersion;
+                            }
+
+                            project = project.GetNearestParent<Project>();
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    private static string GetChildMetadataValue(TreeNode node, string metadataItemName)
+    {
+        var metadata = node.Children.OfType<Metadata>();
+        var metadataValue = metadata.FirstOrDefault(m => m.Name.Equals(metadataItemName, StringComparison.OrdinalIgnoreCase))?.Value;
+        return metadataValue;
+    }
+
+    protected override Task OnFileFoundAsync(ProcessRequest processRequest, IDictionary<string, string> detectorArgs, CancellationToken cancellationToken = default)
+    {
+        try
+        {
+            if (!isMSBuildRegistered)
+            {
+                // this must happen once per process, and never again
+                var defaultInstance = MSBuildLocator.QueryVisualStudioInstances().First();
+                MSBuildLocator.RegisterInstance(defaultInstance);
+                isMSBuildRegistered = true;
+            }
+
+            var singleFileComponentRecorder = this.ComponentRecorder.CreateSingleFileComponentRecorder(processRequest.ComponentStream.Location);
+            var buildRoot = BinaryLog.ReadBuild(processRequest.ComponentStream.Stream);
+            this.RecordLockfileVersion(buildRoot.FileFormatVersion);
+            this.ProcessBinLog(buildRoot, singleFileComponentRecorder);
+        }
+        catch (Exception e)
+        {
+            // If something went wrong, just ignore the package
+            this.Logger.LogError(e, "Failed to process MSBuild binary log {BinLogFile}", processRequest.ComponentStream.Location);
+        }
+
+        return Task.CompletedTask;
+    }
+
+    protected override Task OnDetectionFinishedAsync()
+    {
+        return Task.CompletedTask;
+    }
+
+    private void ProcessBinLog(Build buildRoot, ISingleFileComponentRecorder componentRecorder)
+    {
+        // maps a project path to a set of resolved dependencies
+        var projectTopLevelDependencies = new Dictionary<string, HashSet<string>>(StringComparer.OrdinalIgnoreCase);
+        var projectResolvedDependencies = new Dictionary<string, Dictionary<string, string>>(StringComparer.OrdinalIgnoreCase);
+        buildRoot.VisitAllChildren<BaseNode>(node =>
+        {
+            switch (node)
+            {
+                case NamedNode namedNode when namedNode is AddItem or RemoveItem:
+                    ProcessResolvedPackageReference(projectTopLevelDependencies, projectResolvedDependencies, namedNode);
+                    break;
+                default:
+                    break;
+            }
+        });
+
+        // dependencies were resolved per project, we need to re-arrange them to be per package/version
+        var projectsPerPackage = new Dictionary<string, HashSet<string>>(StringComparer.OrdinalIgnoreCase);
+        foreach (var projectPath in projectResolvedDependencies.Keys)
+        {
+            var projectDependencies = projectResolvedDependencies[projectPath];
+            foreach (var (packageName, packageVersion) in projectDependencies)
+            {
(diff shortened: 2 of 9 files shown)


src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
```diff
@@ -0,0 +1,212 @@
...
+            {
+                // this must happen once per process, and never again
+                var defaultInstance = MSBuildLocator.QueryVisualStudioInstances().First();
+                MSBuildLocator.RegisterInstance(defaultInstance);
+                isMSBuildRegistered = true;
+            }
+
+            var singleFileComponentRecorder = this.ComponentRecorder.CreateSingleFileComponentRecorder(processRequest.ComponentStream.Location);
+            var buildRoot = BinaryLog.ReadBuild(processRequest.ComponentStream.Stream);
+            this.RecordLockfileVersion(buildRoot.FileFormatVersion);
```
KirillOsenkov, line 144: did you mean RecordLogFileVersion?

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
```diff
@@ -0,0 +1,212 @@
...
+        var metadata = node.Children.OfType<Metadata>();
+        var metadataValue = metadata.FirstOrDefault(m => m.Name.Equals(metadataItemName, StringComparison.OrdinalIgnoreCase))?.Value;
+        return metadataValue;
+    }
+
+    protected override Task OnFileFoundAsync(ProcessRequest processRequest, IDictionary<string, string> detectorArgs, CancellationToken cancellationToken = default)
+    {
+        try
+        {
+            if (!isMSBuildRegistered)
```
KirillOsenkov, line 134: can OnFileFoundAsync get called concurrently? If so, you might want a lock here

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
```diff
@@ -0,0 +1,212 @@
...
+            var projectEvaluation = node.GetNearestParent<ProjectEvaluation>();
+            if (projectEvaluation is not null)
+            {
+                foreach (var child in node.Children.OfType<Item>())
+                {
+                    var packageName = child.Name;
+                    if (!topLevelDependencies.TryGetValue(projectEvaluation.ProjectFile, out var topLevel))
+                    {
+                        topLevel = new(StringComparer.OrdinalIgnoreCase);
+                        topLevelDependencies[projectEvaluation.ProjectFile] = topLevel;
```
KirillOsenkov, line 68: one project file may be evaluated and built several times, you'll often see several evaluations per .csproj. Should we pick the best evaluation somehow or is it fine to do for each evaluation? Restore does an eval, then the build does another eval, and each target framework is a separate eval. We probably want a union of all evaluations?

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
```diff
@@ -0,0 +1,212 @@
...
+                    }
+
+                    if (doRemoveOperation)
```
KirillOsenkov, line 71: a bit worried that you might be getting additem/removeitem from different evaluations in an interleaved way, and since they key by the project path this might get confused.

On the other hand it shouldn't happen because the binlog is a tree, and you're walking the tree linearly, so in theory each evaluation should be processed sequentially one after another.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
```diff
@@ -0,0 +1,212 @@
...
+                            {
+                                projectDependencies.Remove(packageName);
+                            }
+
+                            if (doAddOperation)
+                            {
+                                projectDependencies[packageName] = packageVersion;
+                            }
+
+                            project = project.GetNearestParent<Project>();
```
KirillOsenkov, line 115: hmm, why are you walking up the project chain? I don't think these items flow to the calling project?

KirillOsenkov:
As an alternative, you could also see how `binlogtool listnuget msbuild.binlog output.txt` does it. It opens all embedded project.assets.json files, reads the contents into a model, and walks all of them to collect a flat list of all packages and versions used by the build.

I think your approach is fine too (?)

@jeffkl as an expert FYI, in case he wants to take a quick look

KirillOsenkov:
actually we could diff the results of binlogtool listnuget with these results, and see if there are any discrepancies. Would be like an oracle!

brettfo:
> As an alternative, you could also see how `binlogtool listnuget msbuild.binlog output.txt` does it. It opens all embedded project.assets.json files, reads the contents into a model, and walks all of them to collect a flat list of all packages and versions used by the build.
> 
> I think your approach is fine too (?)
> 
> @jeffkl as an expert FYI, in case he wants to take a quick look

I explicitly don't want the contents of `project.assets.json` because they don't necessarily reflect what ends up in the output.  There are more details in the original description, but in some cases the SDK will replace an assembly that NuGet has already resolved, so according to NuGet, a certain package ended up in the build, but in reality it's not there and shouldn't get reported.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
brettfo: This method comes from the parent class and there's no `Log` variant, only `Lock`, but I figure this would be good enough; it's meant to gather telemetry on what versions of files are encountered, and while the `.binlog` isn't a lock file, it does have the same effect of reporting exactly what ended up in the output directory.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
brettfo: Good catch, I'll push a new commit shortly.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
brettfo: I'll experiment a bit to see if I can find inaccurate results, but generally we _really_ only care about the item groups that NuGet populates like `@(RuntimeCopyLocalItems)` and I think that only happens during the `Restore` evaluation.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
brettfo: I found I needed this to track transitive dependencies.  E.g., if `library.csproj` has `<PackageReference Include="Some.Package" />` and `unitTests.csproj` has `<ProjectReference Include="..\library\library.csproj" />` this will add `Some.Package` first to `library.csproj` and then crawl up the chain to `unitTests.csproj` so that the dependency is properly reported for both projects.  The only oddity with this (and I may have to special case it) is the `.sln` file also appears in a `Project` node, but I _do_ seem to be getting the proper hierarchy.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
brettfo: What might be an indication that the traversal is bad?  If I try to process `RemoveItem` but the item isn't already present?  Or is that something that MSBuild won't allow?

KirillOsenkov: yes I guess, but as I said, maybe I'm just being paranoid

KirillOsenkov:
I see, makes sense! 👍🏻

brettfo:
For those interested, I ran the package `Microsoft.Extensions.Configuration.Json/6.0.0` through both the `binlogtool` suggegsted above and this PR.  The results are as follows:

`binlogtool`:

```
Microsoft.Extensions.Configuration/6.0.0
Microsoft.Extensions.Configuration.Abstractions/6.0.0
Microsoft.Extensions.Configuration.FileExtensions/6.0.0
Microsoft.Extensions.Configuration.Json/6.0.0
Microsoft.Extensions.FileProviders.Abstractions/6.0.0
Microsoft.Extensions.FileProviders.Physical/6.0.0
Microsoft.Extensions.FileSystemGlobbing/6.0.0
Microsoft.Extensions.Primitives/6.0.0
System.Runtime.CompilerServices.Unsafe/6.0.0
System.Text.Encodings.Web/6.0.0
System.Text.Json/6.0.0

```

This PR:

```
Microsoft.Extensions.Configuration/6.0.0
Microsoft.Extensions.Configuration.Abstractions/6.0.0
Microsoft.Extensions.Configuration.FileExtensions/6.0.0
Microsoft.Extensions.Configuration.Json/6.0.0
Microsoft.Extensions.FileProviders.Abstractions/6.0.0
Microsoft.Extensions.FileProviders.Physical/6.0.0
Microsoft.Extensions.FileSystemGlobbing/6.0.0
Microsoft.Extensions.Primitives/6.0.0
```

The difference being that the SDK (`6.0.425` on my machine) removed the `System.*` packages.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
```diff
@@ -0,0 +1,224 @@
...
+        return metadataValue;
+    }
+
+    protected override Task OnFileFoundAsync(ProcessRequest processRequest, IDictionary<string, string> detectorArgs, CancellationToken cancellationToken = default)
+    {
+        try
+        {
+            lock (MSBuildRegistrationGate)
+            {
+                if (!isMSBuildRegistered)
```
rainersigwald, line 138: nit: you could use `MSBuildLocator.IsRegistered` and not have to carry this around yourself.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
```diff
@@ -0,0 +1,224 @@
...
+                    var packageName = child.Name;
+                    if (!topLevelDependencies.TryGetValue(projectEvaluation.ProjectFile, out var topLevel))
+                    {
+                        topLevel = new(StringComparer.OrdinalIgnoreCase);
+                        topLevelDependencies[projectEvaluation.ProjectFile] = topLevel;
+                    }
+
+                    if (doRemoveOperation)
+                    {
+                        topLevel.Remove(packageName);
```
rainersigwald, line 75: Specifically I think this is going to cause problems if one TF of a project references a thing and the other gets it removed by framework unification.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
rainersigwald: I don't think I understand this. `unitTests.csproj` should have `Some.Package` in its assets file and pull assets out of it in its `ResolvePackageAssets`. Can you share a log or project setup where this is necessary?

Directory.Packages.props
```diff
@@ -21,9 +21,12 @@
     <PackageVersion Include="MinVer" Version="5.0.0" />
     <PackageVersion Include="Moq" Version="4.18.4" />
     <PackageVersion Include="morelinq" Version="4.2.0" />
+    <PackageVersion Include="MSBuild.StructuredLogger" Version="2.2.317" />
     <PackageVersion Include="MSTest.TestFramework" Version="3.5.1" />
     <PackageVersion Include="MSTest.Analyzers" Version="3.5.1" />
     <PackageVersion Include="MSTest.TestAdapter" Version="3.5.1" />
+    <PackageVersion Include="Microsoft.Build.Framework" Version="17.5.0" />
+    <PackageVersion Include="Microsoft.Build.Locator" Version="1.6.1" />
```
rainersigwald, line 29: I have a concern here. Are you running as a standalone .NET 6 application currently? If so, you won't be able to load MSBuild from new SDKs. Have you tried on a machine that has _only_ .NET 8 installed?

brettfo: This seems to work in the unit tests which should only have .NET6 installed (it's installing [here](https://github.com/microsoft/component-detection/actions/runs/10950179321/workflow?pr=1250#L29) and if I understand that action correctly, it uses `global.json` to determine what to install, so no .NET 8 on the CI machine)

rainersigwald: Right, the problem will be when you run the tool on a different machine which has only .NET 8 installed.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
brettfo: Do you have an example of this?  I'd like to add a test to make sure we don't remove anything that shouldn't be removed.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
brettfo: I just pushed a commit that covers this scenario; check the unit test for some notes, but the end result is that we're scanning the `.binlog` correctly and after a `dotnet build /bl` we'd pick up and report both `Newtonsoft.Json/13.0.3` and `System.Text.Json/8.0.4` and ultimately, that's the end goal: map a `.csproj` (or rather anything that's not `.sln`) to a set of package names and versions that came from it, regardless of the TFM that it came from.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
rainersigwald: I'm saying that the targets that read the assets file should produce `Some.Package` in the transitive case, so walking the project graph here doesn't make sense to me.

brettfo: > ...that read the assets file...

That's the problem, the assets file could be wrong, so I need to crawl it manually.  The PR description explains a scenario where the assets file isn't correct.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
rainersigwald: Try this:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFrameworks>net472;net8.0</TargetFrameworks>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="System.Text.Json" Version="6.0.0" />
  </ItemGroup>

</Project>
```

This project DOES use STJ 6.0.0, but ONLY for the net472 TF; it should be removed by conflict resolution against the net8.0 framework in that TF.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
```diff
@@ -0,0 +1,353 @@
...
+        "PackageReference",
+    };
+
+    // the items listed below represent collection names that NuGet will resolve a package into, along with the metadata value names to get the package name and version
+    private static readonly Dictionary<string, (string NameMetadata, string VersionMetadata)> ResolvedPackageItemNames = new Dictionary<string, (string, string)>(StringComparer.OrdinalIgnoreCase)
+    {
+        // regular restore operations
+        ["NativeCopyLocalItems"] = ("NuGetPackageId", "NuGetPackageVersion"),
+        ["ResourceCopyLocalItems"] = ("NuGetPackageId", "NuGetPackageVersion"),
+        ["RuntimeCopyLocalItems"] = ("NuGetPackageId", "NuGetPackageVersion"),
```
ericstj, line 30: Will you read the value of these late enough in the build to understand that they haven't been reduced by ConflictResolution? 
https://github.com/dotnet/sdk/blob/c3a8f72c3a5491c693ff8e49e7406136a12c3040/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ConflictResolution.targets#L51
https://github.com/dotnet/sdk/blob/c3a8f72c3a5491c693ff8e49e7406136a12c3040/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L301-L306

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetMSBuildBinaryLogComponentDetector.cs
ericstj: The assets file will still be a superset of what the project might use.  You get the "real usage" by looking at the outcome of package resolution / conflict resolution.   

If you walk the project graph you end up trying to replay parts of the build.  We shouldn't do that when we can just observe what it did.

Also - the assets file isn't "Incorrect" here - it is correct from NuGet's perspective.  It's just that the build has more policy that it applies to decide if it will actually use a package's contents.  If that package contributes assets which are "older" than some other contribution it will be dropped.  The most common case is framework, as you mentioned -- it's what we designed the feature for in the SDK.  Technically it could be any conflict though - when any two packages try to provide the same file the build will compare them to decide who's copy wins.

ericstj:
https://github.com/microsoft/component-detection/pull/1710 should replace this.  @brettfo @rainersigwald @KirillOsenkov -- could use your 👀 on that.

PR #1254, opened 2024-09-30
Adding Support for Relative Paths Go Replace Detector

amitla1:
The Current Go Replace detector does not support replacement paths that are relative paths. It only supports replacement given a path and a version. This was causing errors in Go CLI detection. This new upgraded Go Replacement Detector version supports relative paths, if that path contains a go.mod file. There were also tests as well as telemetry and log statements added for this.


src/Microsoft.ComponentDetection.Detectors/go/GoComponentWithReplaceDetector.cs
```diff
@@ -467,6 +498,11 @@ private void RecordBuildDependencies(string goListOutput, ISingleFileComponentRe
         }
     }
 
+    private bool ReplacePathExists(GoBuildModule dependency)
```
grvillic, line 501: Oh when I mentioned refactoring I meant more of the code block in them, not just the if statement. Also you can use Null coalesce logic which is equivalent and less verbose `dependency.Replace?.Path != null`

amitla1: I would prefer not to move it to another function since in one the goComponent variable gets set and in the other block it does not

grvillic (approved):

PR #1255, opened 2024-09-30
Add ARM64 Runtimes

cobya:
Since .NET 5.0, ARM64 execution has been supported for the Windows ecoystem and support was later added to macOS. Component Detection has not been publishing these binaries, so add them to the CI pipeline.


.github/workflows/release.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rid: [win-x64, linux-x64, osx-x64]
+        rid: [win-x64, linux-x64, osx-x64, win-arm64, osx-arm64]
```
JamieMagee, line 17: Could you add `linux-arm64` in while you're here?

pauld-msft (approved):

PR #1259, opened 2024-10-02
Add support for cleaning up files created by tool

pauld-msft:
### Problem
The python PipReport detector creates some files during its run of `pip install --report --dry-run`. There aren't options in the pip install cli that can restrict or clean up these files.

### Solution
Introduce a new property on the FileComponentDetector that can be overridden in the inheriting classes to specify files/directories that the detector may create and that should be cleaned up after the detector has run. This will only remove new files that were added during detection run, and will leave existing files that match the specified patterns.


src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs
```diff
@@ -81,6 +83,8 @@ private enum PipReportOverrideBehavior
 
     protected override bool EnableParallelism { get; set; } = true;
 
+    protected override IList<string> CleanupPatterns => ["*.egg", "*.egg-info", "*.pyc", "*.pyo", "*.pyd", "__pycache__"];
```
pauld-msft, line 86: Tried to keep the additions to the detectors to a minimum so that it is quite easy in the future to add to other detectors if needed

src/Microsoft.ComponentDetection.Orchestrator/Commands/ScanSettings.cs
```diff
@@ -80,6 +80,10 @@ public class ScanSettings : BaseSettings
     [Description("Max number of parallel threads used for a single detection process, ex: PipReport, Npm, Nuget.")]
     public int? MaxDetectionThreads { get; set; }
 
+    [CommandOption("--CleanupCreatedFiles")]
+    [Description("Whether or not to cleanup files that are created during detection, based on the rules provided in each detector.")]
+    public bool? CleanupCreatedFiles { get; set; }
```
RushabhBhansali, line 85: nit: mentioning default value = true in the description might be helpful while running this from command line.

src/Microsoft.ComponentDetection.Contracts/FileComponentDetectorWithCleanup.cs
```diff
@@ -0,0 +1,154 @@
...
+            // Ensure that only one cleanup process is running at a time, helping to prevent conflicts
+            await this.cleanupSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+
+            try
+            {
+                // Clean up any new files or directories created during the scan that match the clean up patterns.
+                // If the cleanupCreatedFiles flag is set to false, this will be a dry run and will just log the files that it would clean.
+                var dryRun = !cleanupCreatedFiles;
+                var dryRunStr = dryRun ? "[DRYRUN] " : string.Empty;
+                var (newFiles, newDirs) = this.DirectoryUtilityService.GetFilesAndDirectories(fileParentDirectory, this.CleanupPatterns, DefaultCleanDepth);
```
RushabhBhansali, line 82: nit: `newFile` and `newDir` names give an impression of these being the newly created file. can we change it to something different for e.g. `latestFiles` `latestDirs`

RushabhBhansali (commented):
provided comments

src/Microsoft.ComponentDetection.Common/DirectoryUtilityService.cs
```diff
@@ -0,0 +1,51 @@
...
+    public IEnumerable<string> EnumerateFiles(string path, string searchPattern, SearchOption searchOption) =>
+        Directory.EnumerateFiles(path, searchPattern, searchOption);
+
+    /// <inheritdoc />
+    public bool Exists(string directoryPath) => Directory.Exists(directoryPath);
+
+    /// <inheritdoc />
+    public (HashSet<string> Files, HashSet<string> Directories) GetFilesAndDirectories(string root, IList<string> patterns, int depth)
+    {
+        var fileList = new List<string>();
```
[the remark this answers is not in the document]
pauld-msft: I included a depth here to avoid the infinite loop that can be introduced when AllDirectories is used: https://learn.microsoft.com/en-us/dotnet/api/system.io.searchoption?view=net-8.0#remarks 

It also reduces the amount of computation that happens for each of the detectors, and so far I haven't seen a directory created more than 2 levels below the `setup.py` files

.github/workflows/snapshot-verify.yml
```diff
@@ -52,6 +54,11 @@ jobs:
       - name: Setup .NET Core
         uses: actions/setup-dotnet@6bd8b7f7774af54e05809fcc5431931b3eb1ddee # v4.0.1
 
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
```
pauld-msft, line 60: Allows pipreport to run in verification

RushabhBhansali (approved):

PR #1266, opened 2024-10-08
Add TargetFramework awareness to NuGet detector

ericstj:
This PR does 2 things.

1. Adds `TargetFramework` to NuGet package references.  This can be useful when querying component data to understand if components are used in a place where a vulnerability applies.
2. Adds _framework package_ handling.  The .NET SDK will do [conflict resolution](https://github.com/dotnet/sdk/tree/main/src/Tasks/Common/ConflictResolution) and drop assets from packages that overlap with the framework.  NuGet is planning to do the same https://github.com/NuGet/Home/issues/7344 but until then, it's beneficial to have component detection duplicate some of this logic.  When a package is identified as overlapping with the framework we'll treat it as a Development Dependency so that it might be auto-dismissed.
- .NETFramework projects do not get this - .NETFramework does not participate in conflict resolution by default.  Even when enabled framework assemblies can be bypassed using bindingRedirects, or avoiding references to them.  Due this fragility it's not safe to apply framework package rules to .NETFramework.
- packages.config usage is also excluded since it precludes SDK conflict resolution (and is also only used on .NETFramework projects).

A couple notes:
1. I reused _Development Dependency_ rather than plumbing a new concept.
2. I only mapped data for the `Microsoft.NETCore.App` - the default shared framework.  We could consider doing the same for `Microsoft.ASPNETCore.App` and `Microsoft.WindowsDesktop.App` but we'd need to plumb the reference information out of the assets file - currently that's not read and I'm not aware of a supported NuGet API for reading it (though it is present under `project/frameworks/<framework>/frameworkReferences/<name>`
3. I've tested this manually on a simple project and a more complex repository.  It's doing what I expect and I see the `targetFramework` information added to the `component` in the ScanManifest.  I noticed that development dependencies are labeled as such if all references to the component are development dependencies.  If any reference is not a development dependency then the component is not labeled as a development dependency.
4. I need to add automated tests for this.
5. I added the tool `findLastPackage` that I used to generate the sources for the framework lists - but this is just for reference.  I don't intend to check this tool in.
6. Maybe we could even do more with CG here to mark libaries as development dependencies when all their runtime assets are excluded.  This technique is common for plugins and excluding them in this way would be beneficial for folks even when not in a framework dependency case.  I'll add this.


src/Microsoft.ComponentDetection.Detectors/nuget/NuGetPackagesConfigDetector.cs
```diff
@@ -53,17 +53,27 @@ protected override Task OnFileFoundAsync(ProcessRequest processRequest, IDiction
...
+                    /* TODO: Is this really the same concept?
+                       Docs for NuGet say packages.config development dependencies are just not persisted as dependencies in the package.
+                       That is not same as excluding from the output directory / runtime. */
                     package.IsDevelopmentDependency);
+
+                // get the actual component in case it already exists
+                var libraryComponent = singleFileComponentRecorder.GetComponent(package.PackageIdentity.Id);
+
+                // Add framework information to the actual component
+                ((NuGetComponent)libraryComponent.Component).TargetFrameworks.Add(package.TargetFramework.GetShortFolderName());
```
grvillic, line 76: Are sure `package.TargetFramework` is always included? Might be worth adding a `string.IsNullOrWhiteSpace(package.TargetFramework?.GetShortFolderName())` and not add it to TargetFramework list if it isn't valid.

src/Microsoft.ComponentDetection.Detectors/nuget/FrameworkPackages/FrameworkPackages.cs
```diff
@@ -0,0 +1,116 @@
...
+
+    public static FrameworkPackages GetFrameworkPackages(NuGetFramework framework)
+    {
+        if (FrameworkPackagesByFramework.TryGetValue(framework, out var frameworkPackages))
+        {
+            return frameworkPackages;
+        }
+
+        // if we didn't predefine the package overrides, load them from the targeting pack
+        // we might just leave this out since in future frameworks we'll have this functionality built into NuGet.
```
grvillic, line 54: Any link to public documentation about this coming out later on? You share offline with me if it is a private wiki.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetProjectModelProjectCentricComponentDetector.cs
```diff
@@ -264,19 +103,27 @@ private void NavigateAndRegister(
...
         }
 
+        var isFrameworkComponent = frameworkPackages.IsAFrameworkComponent(library.Name, library.Version);
+        var isDevelopmentDependency = IsADevelopmentDependency(library);
+
         visited ??= [];
 
         var libraryComponent = new DetectedComponent(new NuGetComponent(library.Name, library.Version.ToNormalizedString()));
-        singleFileComponentRecorder.RegisterUsage(libraryComponent, explicitlyReferencedComponentIds.Contains(libraryComponent.Component.Id), parentComponentId);
+        singleFileComponentRecorder.RegisterUsage(libraryComponent, explicitlyReferencedComponentIds.Contains(libraryComponent.Component.Id), parentComponentId, isDevelopmentDependency: isFrameworkComponent || isDevelopmentDependency);
```
grvillic, line 120: `isDevelopmentDependency` is used to flag dependencies that are never shipped with binaries to production, is that the case for all the framework components? If someone introduced the dependency directly in their project it shouldn't be treated as Development dependency right?

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetProjectModelProjectCentricComponentDetector.cs
```diff
@@ -264,19 +103,27 @@ private void NavigateAndRegister(
...
+
+        // get the actual component in case it already exists
+        libraryComponent = singleFileComponentRecorder.GetComponent(libraryComponent.Component.Id);
+
+        // Add framework information to the actual component
+        ((NuGetComponent)libraryComponent.Component).TargetFrameworks.Add(target.TargetFramework.GetShortFolderName());
```
grvillic, line 126: ditto about being certain that TargetFramework is populated everywhere

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetProjectModelProjectCentricComponentDetector.cs
```diff
@@ -293,30 +140,15 @@ private void NavigateAndRegister(
...
-            {
-                // Also track shallow children if this is a top level library so we have a rough count of how many things have been ommitted + root relationships
-                foreach (var item in dependencies)
-                {
-                    this.frameworkComponentsThatWereOmmittedWithCount.AddOrUpdate(item.Id, 1, (name, existing) => existing + 1);
-                }
-            }
-        }
+        // a placeholder item is an empty file that doesn't exist with name _._ meant to indicate an empty folder in a nuget package, but also used by NuGet when a package's assets are excluded.
+        bool IsAPlaceholderItem(LockFileItem item) => Path.GetFileName(item.Path).Equals("_._", StringComparison.OrdinalIgnoreCase);
```
grvillic, line 148: Any links to documentation about this `_._`, I have never seen this beofre.

src/Microsoft.ComponentDetection.Detectors/nuget/FrameworkPackages/FrameworkPackages.cs
ericstj: Here's the design doc NuGet is working on: https://github.com/NuGet/Home/pull/13634

src/Microsoft.ComponentDetection.Detectors/nuget/FrameworkPackages/FrameworkPackages.net9.0.cs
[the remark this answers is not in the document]
ericstj: Frameworks already have this list but it wasn't used as the source of truth for frameworks up until this point.  I do fallback to loading from the PackageOverrides.txt which is the source of truth moving forward.  That said, it will be moot once NuGet honors this list to avoid even downloading the packages.

I'll add some more links.  As I mentioned in the tool readme, we compute these for now.  Think of it as capturing the result of what the SDK does already given immutable inputs, so its safe to store as a stand-in for reading the output of the SDK (which the build log analyzer can do).

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetPackagesConfigDetector.cs
ericstj: Yeah, I can add that.   Agreed that we can't trust what packages.config exposes - it's user defined.  If we made this read the project it could do better, but I'm not inclined to do much for packages.config since it's not going to benefit from the auto-dismissal.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetProjectModelProjectCentricComponentDetector.cs
ericstj: > is that the case for all the framework components

Yes, that's the case.  This is meant to represent the cases where the .NET SDK will drop the package assets in favor of the framework assets.

> If someone introduced the dependency directly in their project it shouldn't be treated as Development dependency right?

No, this would still be dropped by the SDK.  For example if I have a `net8.0` project and directly reference `System.Text.Json` with version <= `8.0.0` it will not be included in my app.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetProjectModelProjectCentricComponentDetector.cs
ericstj: I think we can trust it here - this isn't a user file - but I will double check if I can somehow break it.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetProjectModelProjectCentricComponentDetector.cs
```diff
@@ -264,19 +103,27 @@ private void NavigateAndRegister(
...
         {
             return;
         }
 
+        var isFrameworkComponent = frameworkPackages.IsAFrameworkComponent(library.Name, library.Version);
+        var isDevelopmentDependency = IsADevelopmentDependency(library);
+
         visited ??= [];
 
         var libraryComponent = new DetectedComponent(new NuGetComponent(library.Name, library.Version.ToNormalizedString()));
```
grvillic, line 119: If we happen to be dropping the framework version due to dependency resolution, should this version be higher? Namely folks using `System.Text.Json 6.0.0` but actually running a dotnet core app with `System.Text.Json 7.0.4`, shouldn't the framework version win here, or which one ends up getting used by the project?

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetProjectModelProjectCentricComponentDetector.cs
ericstj: Here's an old discussion https://github.com/dotnet/aspnetcore/issues/744#issuecomment-123411563

SDK does the same: https://github.com/dotnet/sdk/blob/73bde768ff9058723021da4a9b28f93ff85f8952/src/Tasks/Microsoft.NET.Build.Tasks/NuGetUtils.NuGet.cs#L11

There used to be public docs on this, when creating a nuspec you'd use this file to create an empty directory, so that it could represent the package would install but wouldn't provide any assets.  It was important when representing a assembly that was inboxed in a later framework, or that targets weren't needed on a framework, etc.  NuGet reused it when implementation `ExcludeAssets` so that it would look the same as if a package included an empty folder.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetProjectModelProjectCentricComponentDetector.cs
ericstj: When the framework wins it means the version isn't shipping with the app (for framework dependent app) or that the framework's runtime pack is winning (for self-contained).  In the first case there is no need for an alert - nothing for the app to do.  In the latter case, the runtime package will be seen by CG and the CVE will apply to the framework runtime package.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetProjectModelProjectCentricComponentDetector.cs
ericstj: For reference here's the place it's set https://github.com/NuGet/NuGet.Client/blob/8b88f9e28677f39bd871f4edded2ea4f68b01ac7/src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs#L404

I think we can count on it always being set, nonetheless I added a fallback.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetProjectModelProjectCentricComponentDetector.cs
ericstj: I updated the public docs to mention this: https://github.com/NuGet/docs.microsoft.com-nuget/pull/3348/files

I also changed this to use the public API: `PackagingCoreConstants.EmptyFolder`

ericstj:
Closing this one as it's out of date.  Will open up a new PR that promotes the experiment to the main detector.

PR #1272, opened 2024-10-10
Move Go Replace Detector to Prod

amitla1:
Move Go replace detector from experimental to the regular go detector #1052


src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -284,6 +287,7 @@ private async Task ParseGoModFileAsync(
         GoGraphTelemetryRecord goGraphTelemetryRecord)
     {
         using var reader = new StreamReader(file.Stream);
+        var startString = "require ";
```
grvillic, line 290: nit: let's make this a constant in the class

grvillic (approved):

grvillic (approved):

PR #1273, opened 2024-10-14
Handle Go Replace Exceptions

amitla1:
When trying to use the replace module if the replace module path goes to a private repository it can cause an error in this case use the original path and log the warning to users that way they know as they may need to action on their end to use replace modules


src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -476,11 +476,20 @@ private void RecordBuildDependencies(string goListOutput, ISingleFileComponentRe
...
+                    var dependencyReplacementName = $"{dependency.Replace.Path} {dependency.Replace.Version}";
+                    goComponent = new GoComponent(dependency.Replace.Path, dependency.Replace.Version);
+                    this.Logger.LogInformation("go Module {GoModule} being replaced with module {GoModuleReplacement}", dependencyName, dependencyReplacementName);
+                    record.GoModPathAndVersion = dependencyName;
+                    record.GoModReplacement = dependencyReplacementName;
+                }
+                catch (Exception ex)
+                {
+                    record.ExceptionMessage = ex.Message;
+                    this.Logger.LogWarning("tried to use replace module but got this error {ErrorMessage} using original module instead", ex.Message);
```
grvillic, line 490: Let's be explicit here on what is the original module and version we used, the replace version that failed, and the exception message.

grvillic (approved):

PR #1285, opened 2024-10-28
Add experimental NuGet detector for framework and dev dependencies

ericstj:
Replaces #1266 as an experimental detector.

This PR does 3 things.

1. Adds `TargetFramework` to NuGet package references.  This can be useful when querying component data to understand if components are used in a place where a vulnerability applies.
2. Adds _framework package_ handling.  The .NET SDK will do [conflict resolution](https://github.com/dotnet/sdk/tree/main/src/Tasks/Common/ConflictResolution) and drop assets from packages that overlap with the framework.  NuGet is planning to do the same https://github.com/NuGet/Home/issues/7344 but until then, it's beneficial to have component detection duplicate some of this logic.  When a package is identified as overlapping with the framework we'll treat it as a Development Dependency so that it might be auto-dismissed.
  - .NETFramework projects do not get this - .NETFramework does not participate in conflict resolution by default.  Even when enabled framework assemblies can be bypassed using bindingRedirects, or avoiding references to them.  Due this fragility it's not safe to apply framework package rules to .NETFramework.
  - packages.config usage is also excluded since it precludes SDK conflict resolution (and is also only used on .NETFramework projects).
3. Recognizes `ExcludeAssets="Runtime"` usage as a Development Dependencies, also any packages which don't contribute to "runtime" will be developement dependencies.

I reused _Development Dependency_ rather than plumbing a new concept. I only mapped data for the `Microsoft.NETCore.App` - the default shared framework.  We could consider doing the same for `Microsoft.ASPNETCore.App` and `Microsoft.WindowsDesktop.App` but we'd need to plumb the reference information out of the assets file - currently that's not read and I'm not aware of a supported NuGet API for reading it (though it is present under `project/frameworks/<framework>/frameworkReferences/<name>` .NET Core 1.x has no data since it was packages itself.  I have a fallback for future frameworks to read the data from the targeting packs.


src/Microsoft.ComponentDetection.Detectors/nuget/NuGetPackageReferenceFrameworkAwareDetector.cs
```diff
@@ -0,0 +1,231 @@
...
+using global::NuGet.Packaging.Core;
+using global::NuGet.ProjectModel;
+using global::NuGet.Versioning;
+using Microsoft.ComponentDetection.Contracts;
+using Microsoft.ComponentDetection.Contracts.Internal;
+using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+public class NuGetPackageReferenceFrameworkAwareDetector : FileComponentDetector
```
grvillic, line 19: We need to add the experimental interface here to safely validate it beforehand:

```suggestion
public class NuGetPackageReferenceFrameworkAwareDetector : FileComponentDetector, IExperimentalDetector
```

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetPackageReferenceFrameworkAwareDetector.cs
```diff
@@ -0,0 +1,231 @@
...
+{
+    public const string OmittedFrameworkComponentsTelemetryKey = "OmittedFrameworkComponents";
+
+    public const string ProjectDependencyType = "project";
+
+    private readonly ConcurrentDictionary<string, int> frameworkComponentsThatWereOmmittedWithCount = new ConcurrentDictionary<string, int>();
```
grvillic, line 25: Are we using this dictionary at all in this new detector?

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetProjectModelProjectCentricComponentDetector.cs
```diff
@@ -278,6 +278,13 @@ private void NavigateAndRegister(
         var libraryComponent = new DetectedComponent(new NuGetComponent(library.Name, library.Version.ToNormalizedString()));
         singleFileComponentRecorder.RegisterUsage(libraryComponent, explicitlyReferencedComponentIds.Contains(libraryComponent.Component.Id), parentComponentId);
 
+        // get the actual component in case it already exists
+        libraryComponent = singleFileComponentRecorder.GetComponent(libraryComponent.Component.Id);
+
+        // Add framework information to the actual component
+        var targetFramework = target.TargetFramework?.GetShortFolderName() ?? "unknown";
```
grvillic, line 285: If we don't know the framework (aka `unknown`, empty, or whitespace), let's not add anything to the `TargetFrameworks` set.

src/Microsoft.ComponentDetection.Orchestrator/Experiments/Configs/NuGetTargetFrameworkExperiment.cs
```diff
@@ -0,0 +1,23 @@
+namespace Microsoft.ComponentDetection.Orchestrator.Experiments.Configs;
+
+using Microsoft.ComponentDetection.Contracts;
+using Microsoft.ComponentDetection.Detectors.NuGet;
+using Microsoft.ComponentDetection.Detectors.Pip;
+
+/// <summary>
+/// Validating the <see cref="SimplePipComponentDetector"/>.
```
grvillic, line 8: nit: `cref` is pointing to the Pip detector

grvillic (changes requested):

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetPackageReferenceFrameworkAwareDetector.cs
ericstj: Nope, I can remove it.  Let me see if there is more here to remove.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetProjectModelProjectCentricComponentDetector.cs
ericstj: This was intentional.  The addition of `TargetFrameworks` to the `NuGetComponent` is for diagnostics.  It's helpful to understand when a component was detected without a framework and be able to distinguish that from a case where a component was reported by a different detector (either old one, or `NuGetComponentDetector`)

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetPackageReferenceFrameworkAwareDetector.cs
```diff
@@ -0,0 +1,231 @@
...
+                    .ToHashSet();
+
+            // Since we report projects as the location, we ignore the passed in single file recorder.
+            var singleFileComponentRecorder = this.ComponentRecorder.CreateSingleFileComponentRecorder(lockFile.PackageSpec.RestoreMetadata.ProjectPath);
+            foreach (var target in lockFile.Targets)
+            {
+                var frameworkPackages = FrameworkPackages.GetFrameworkPackages(target.TargetFramework);
+
+                // This call to GetTargetLibrary is not guarded, because if this can't be resolved then something is fundamentally broken (e.g. an explicit dependency reference not being in the list of libraries)
+                foreach (var library in explicitReferencedDependencies.Select(x => target.GetTargetLibrary(x.Name)).Where(x => x != null))
```
ericstj, line 79: There's an existing bug here.  The detector is treating the union of top-level dependencies for all frameworks as top level when evaluating a single target framework.  That's why this `x != null` check is here - it shouldn't be required if the top level dependencies were walked per-framework.

It still ends up walking the dependency graph for that specific target framework so I don't think any dependencies are missed, but it will over-report top-level dependencies.

In the interest of minimizing churn with the existing detector I'll leave this behavior alone.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetPackageReferenceFrameworkAwareDetector.cs
```diff
@@ -0,0 +1,231 @@
...
+        catch (Exception e)
+        {
+            // If something went wrong, just ignore the package
+            this.Logger.LogError(e, "Failed to process NuGet lockfile {NuGetLockFile}", processRequest.ComponentStream.Location);
+        }
+
+        return Task.CompletedTask;
+    }
+
+    protected override Task OnDetectionFinishedAsync()
```
ericstj, line 94: This is not required.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetProjectModelProjectCentricComponentDetector.cs
ericstj: Actually it's never going to be the case here so I'm fine to just leave it out.  That's what I did for packages.config.

ericstj:
@grvillic - ready for another pass.  Thank you for your review.

grvillic (approved):

PR #1296, opened 2024-10-31
Record NuGet PackageDownloads

ericstj:
Read PackageDownloads from the assets file and record them as top-level development dependencies.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/nuget/NuGetPackageReferenceFrameworkAwareDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/nuget/NuGetPackageReferenceFrameworkAwareDetector.cs
@@ -6,6 +6,7 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using global::NuGet.Frameworks;
 using global::NuGet.Packaging.Core;
 using global::NuGet.ProjectModel;
 using global::NuGet.Versioning;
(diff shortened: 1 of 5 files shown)


ericstj:
@grvillic -- thoughts on this one?

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetPackageReferenceFrameworkAwareDetector.cs
```diff
@@ -141,16 +144,35 @@ private void NavigateAndRegister(
...
     }
 
-    private List<(string Name, Version Version, VersionRange VersionRange)> GetTopLevelLibraries(LockFile lockFile)
+    private void RegisterPackageDownloads(ISingleFileComponentRecorder singleFileComponentRecorder, LockFile lockFile)
+    {
+        foreach (var framework in lockFile.PackageSpec.TargetFrameworks)
+        {
+            foreach (var packageDownload in framework.DownloadDependencies)
+            {
+                var libraryComponent = new DetectedComponent(new NuGetComponent(packageDownload.Name, packageDownload.VersionRange.MinVersion.ToNormalizedString()));
```
grvillic, line 153: I don't know if this is even possible, but let's skip Download dependencies if package download name or version range are null or empty. We can log an informational message when that is the case. Just safeguarding any invalid manifest files that are missing metadata or are corrupted.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetPackageReferenceFrameworkAwareDetector.cs
```diff
@@ -141,16 +144,35 @@ private void NavigateAndRegister(
...
+                    targetFramework: framework.FrameworkName?.GetShortFolderName());
+            }
+        }
+    }
+
+    private List<(string Name, Version Version, VersionRange VersionRange, bool IsPackageDownload)> GetTopLevelLibraries(LockFile lockFile)
     {
         // First, populate libraries from the TargetFrameworks section -- This is the base level authoritative list of nuget packages a project has dependencies on.
-        var toBeFilled = new List<(string Name, Version Version, VersionRange VersionRange)>();
+        var toBeFilled = new List<(string Name, Version Version, VersionRange VersionRange, bool IsPackageDownload)>();
```
grvillic, line 169: Are we using this new property `IsPackageDownload`? Doesn't seem like it.

grvillic (changes requested):
Minor comments

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetPackageReferenceFrameworkAwareDetector.cs
ericstj: I don't ever expect these to be null or have null values so we can just skip them.  https://github.com/NuGet/NuGet.Client/blob/38729596a6549df3cd1eed37b93d51df1bf62bff/src/NuGet.Core/NuGet.ProjectModel/JsonPackageSpecReader.Utf8JsonStreamReader.cs#L746

grvillic (approved):

PR #1298, opened 2024-11-05
Add .NETCoreApp 1.1 runtime packages to FrameworkPackages

ericstj:
These packages are referenced when building RID-specific / self-contained applications and referencing 1.x era .NET Core packages.


ericstj:
I think I might have some duplicates I can remove.  Will test locally and reduce.

PR #1300, opened 2024-11-12
Detect nuget framework packages for additional frameworks

ericstj:
This adds support for additional framework references - Microsoft.AspNetCore.App and Microsoft.WindowsDesktop.App.  It is a follow up to https://github.com/microsoft/component-detection/pull/1285

While doing this I also adjusted the data used to generate the package information.  I now consider the platformmanifest rather than reference assemblies.  This is more accurate since conflict resolution uses the platform manifest for runtime assemblies.  I also improved the reduction algorithm to remove duplicates.

I added tests for these scenarios as well as for a "missing framework" scenario.  The latter made me realize we might write to the cache of package information and I modified it to use ConcurrentDictionaries.

@grvillic


src/Microsoft.ComponentDetection.Detectors/nuget/FrameworkPackages/FrameworkPackages.net8.0.cs
```diff
@@ -9,18 +9,182 @@ internal partial class FrameworkPackages
 {
     internal static class NETCoreApp80
     {
-        internal static FrameworkPackages Instance { get; } = new(NuGetFramework.Parse("net8.0"), NETCoreApp70.Instance)
+        internal static FrameworkPackages Instance { get; } = new(NuGetFramework.Parse("net8.0"), "Microsoft.NETCore.App", NETCoreApp70.Instance)
```
grvillic, line 12: Let's create a shared static constant for these values: "Microsoft.NETCore.App", "Microsoft.AspNetCore.App", and "Microsoft.WindowsDesktop.App" that we can consume across all files.

grvillic (approved):

PR #648, opened 2023-07-10
fix(experiments): don't overwrite file locations

melotic:
Fixes experiments overwriting file locations and corrupting them with bad relative paths.

PR #1306, opened 2024-11-16
Fix detector component filepath custom overrides

grvillic:
## Context
A detector does not need to specify location where a component was found. Nevertheless, custom wrapper plugins can report additional locations if desired and that shouldn't be blocked. 

In PR https://github.com/microsoft/component-detection/pull/648 a regression occurred while trying to address an experiments framework bug. We don't want to unsuspectingly record locations for experimental detectors before producing the final graph. We added a check to prevent experiments from overriding the graph, but this was partially incomplete, because we are clearing the file paths at the top of the method.

## Solution
Use the same flag to verify if the component file paths should be updated and only clear the list when generating final graph.


src/Microsoft.ComponentDetection.Orchestrator/Services/GraphTranslation/DefaultGraphTranslationService.cs
```diff
@@ -103,7 +103,11 @@ private IEnumerable<DetectedComponent> GatherSetOfDetectedComponentsUnmerged(IEn
                     // clone custom locations and make them relative to root.
                     var declaredRawFilePaths = component.FilePaths ?? [];
                     var componentCustomLocations = JsonConvert.DeserializeObject<HashSet<string>>(JsonConvert.SerializeObject(declaredRawFilePaths));
-                    component.FilePaths?.Clear();
```
grvillic, line 106: the gist of the fix was here, we are clearing paths which impact the final graph, since we are reinserting them when `updateLocations` is false.

PR #1307, opened 2024-11-18
Update syft image and doc - use docker commands

pauld-msft:
Syft image update to newest version.

Updated docs as well since Skopeo has issue with authenticating with ACR. Also, it requires WSL to be setup as there is no Windows support.
Docker tagging and push works for this scenario, so we can use that instead: https://github.com/containers/skopeo/issues/1534


grvillic (approved):

grvillic (changes requested):
Let's bump the linux detector version too

grvillic (approved):

PR #1316, opened 2024-11-28
Add Swift Package Manager component detection support

raporpe:
Component detection **does not** currently **support detection of Swift Package Manager**. This package manager is the new default for swift-related projects now that CocoaPods (supported by Component detection) [is on maintenance mode](https://blog.cocoapods.org/CocoaPods-Support-Plans/).

This new detector adds basic support to SwiftPM by parsing the JSON file `Package.resolved` that is generated when dependencies are resolved. This file is usually present in repositories and after building a Swift Package and its contents are enough to detect both direct and transitive dependencies.

This detector does not parse the `Package.swift` file (hard to parse and does not include transitive dependencies) and does not differentiate between build dependencies, transitive dependencies, and direct dependencies (it is not possible by just parsing `Package.resolved`).

This detector will register two kind of components: a `SwiftPMComponent` and a `GitComponent`. I have decided to also register a Git component because SwiftPM does not rely in a repository infrastructure such as NPM, it directly downloads other Swift Packages from the git repositories. I need feedback in this regard, since I do not know what is the best practice (maybe having two registered components for the same dependency is not a good idea?)

I had to invest time in troubleshooting the verification tests because I was not passing them. The issue is that the reference snapshot has less packages because the environments differ for `PipReport` since PR #1259. I have added the same environment vars to the snapshot-publish pipeline although maybe @pauld-msft could take a look at this PR to verify that my modification is correct.

### DIFF
--- a/.github/workflows/snapshot-publish.yml
+++ b/.github/workflows/snapshot-publish.yml
@@ -2,6 +2,8 @@ name: Publish snapshot of test scan
 
 env:
   CD_DETECTOR_EXPERIMENTS: 1
+  PipReportSkipFallbackOnFailure: "true"
+  PIP_INDEX_URL: "https://pypi.python.org/simple"
 
 on:
   push:
@@ -54,7 +56,7 @@ jobs:
         run:
           dotnet run scan --Verbosity Verbose --SourceDirectory ${{ github.workspace }}/test/Microsoft.ComponentDetection.VerificationTests/resources --Output ${{ github.workspace }}/output
           --DockerImagesToScan "docker.io/library/debian@sha256:9b0e3056b8cd8630271825665a0613cc27829d6a24906dc0122b3b4834312f7d,mcr.microsoft.com/cbl-mariner/base/core@sha256:c1bc83a3d385eccbb2f7f7da43a726c697e22a996f693a407c35ac7b4387cd59,docker.io/library/alpine@sha256:1304f174557314a7ed9eddb4eab12fed12cb0cd9809e4c28f29af86979a3c870"
-          --DetectorArgs DockerReference=EnableIfDefaultOff,SPDX22SBOM=EnableIfDefaultOff,SimplePip=EnableIfDefaultOff
+          --DetectorArgs DockerReference=EnableIfDefaultOff,SPDX22SBOM=EnableIfDefaultOff
 
       - name: Upload output folder
         uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # v4.4.3
--- a/src/Microsoft.ComponentDetection.Detectors/swiftpm/SwiftPMResolvedComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/swiftpm/SwiftPMResolvedComponentDetector.cs
@@ -1,0 +1,105 @@
+namespace Microsoft.ComponentDetection.Detectors.SwiftPM;
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.ComponentDetection.Contracts;
+using Microsoft.ComponentDetection.Contracts.Internal;
+using Microsoft.ComponentDetection.Contracts.TypedComponent;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+/// <summary>
+/// Detects SwiftPM components.
+/// </summary>
+public class SwiftPMResolvedComponentDetector : FileComponentDetector, IDefaultOffComponentDetector
+{
+    public SwiftPMResolvedComponentDetector(
+        IComponentStreamEnumerableFactory componentStreamEnumerableFactory,
+        IObservableDirectoryWalkerFactory walkerFactory,
+        ILogger<SwiftPMResolvedComponentDetector> logger)
+    {
+        this.ComponentStreamEnumerableFactory = componentStreamEnumerableFactory;
+        this.Scanner = walkerFactory;
+        this.Logger = logger;
+    }
+
+    public override string Id { get; } = "SwiftPM";
+
+    public override IEnumerable<string> Categories => [Enum.GetName(DetectorClass.SwiftPM)];
+
+    public override IList<string> SearchPatterns { get; } = ["Package.resolved"];
+
+    public override IEnumerable<ComponentType> SupportedComponentTypes => [ComponentType.SwiftPM];
+
+    public override int Version => 2;
+
+    protected override Task OnFileFoundAsync(
+        ProcessRequest processRequest,
+        IDictionary<string, string> detectorArgs,
+        CancellationToken cancellationToken = default)
+    {
+        try
+        {
+            this.ProcessPackageResolvedFile(processRequest.SingleFileComponentRecorder, processRequest.ComponentStream);
+        }
+        catch (Exception exception)
+        {
+            this.Logger.LogError(exception, "SwiftPMComponentDetector: Error processing Package.resolved file: {Location}", processRequest.ComponentStream.Location);
+        }
+
+        return Task.CompletedTask;
+    }
+
+    private void ProcessPackageResolvedFile(ISingleFileComponentRecorder singleFileComponentRecorder, IComponentStream componentStream)
+    {
+        var parsedResolvedFile = this.ReadAndParseResolvedFile(componentStream.Stream);
+
+        foreach (var package in parsedResolvedFile.Pins)
+        {
+            // We are only interested in packages coming from remote sources such as git
+            // The Package Kind is not an enum because the SwiftPM contract does not specify the possible values.
+            var targetSwiftPackageKind = "remoteSourceControl";
+            if (package.Kind == targetSwiftPackageKind)
+            {
+                // The version of the package is not always available.
+                var version = package.State.Version ?? package.State.Branch ?? package.State.Revision;
+
+                var detectedSwiftPMComponent = new SwiftPMComponent(
+                    name: package.Identity,
+                    version: version,
+                    packageUrl: package.Location,
+                    hash: package.State.Revision);
+                var newDetectedSwiftComponent = new DetectedComponent(component: detectedSwiftPMComponent, detector: this);
+                singleFileComponentRecorder.RegisterUsage(newDetectedSwiftComponent);
+
+                // We also register a Git component for the same package so that the git URL is registered.
+                // SwiftPM directly downloads the package from the git URL.
+                var detectedGitComponent = new GitComponent(
+                    repositoryUrl: new Uri(package.Location),
+                    commitHash: package.State.Revision,
+                    tag: version);
+                var newDetectedGitComponent = new DetectedComponent(component: detectedGitComponent, detector: this);
+                singleFileComponentRecorder.RegisterUsage(newDetectedGitComponent);
+            }
+        }
+    }
+
+    /// <summary>
+    /// Reads the stream of the package resolved file and parses it.
+    /// </summary>
+    /// <param name="stream">The stream of the file to parse.</param>
+    /// <returns>The parsed object.</returns>
+    private SwiftPMResolvedFile ReadAndParseResolvedFile(Stream stream)
+    {
+        string resolvedFile;
+        using (var reader = new StreamReader(stream))
+        {
+            resolvedFile = reader.ReadToEnd();
+        }
+
+        return JsonConvert.DeserializeObject<SwiftPMResolvedFile>(resolvedFile);
+    }
+}
--- a/docs/detectors/swiftpm.md
+++ b/docs/detectors/swiftpm.md
@@ -1,0 +1,24 @@
+# Go Detection
+
+## Requirements
+
+Swift Package Manager detection requires the following file to be present in the scan directory:
+
+-   `Package.resolved` file
+
+## Detection strategy
+
+The detector `SwiftPMResolvedComponentDetector` only parses the `Package.resolved` file to get the dependencies.
+This file contains a json representation of the resolved dependencies of the project with the transitive dependencies.
+The version, the url and commit hash of the dependencies are stored in this file. 
+
+[This is the only reference in the Apple documentation to the `Package.resolved` file.][1]
+
+
+## Known limitations
+
+Right now the detector does not support parsing `Package.swift` file to get the dependencies. 
+It only supports parsing `Package.resolved` file. 
+Some projects only commit the `Package.swift`, which is why it is planned to support parsing `Package.swift` in the future.
+
+[1]: https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#package-dependency
--- a/src/Microsoft.ComponentDetection.Contracts/DetectorClass.cs
+++ b/src/Microsoft.ComponentDetection.Contracts/DetectorClass.cs
@@ -44,4 +44,7 @@ public enum DetectorClass
 
     /// <summary>Indicates a detector applies to Docker references.</summary>
     DockerReference,
+
+    /// <summary> Indicates a detector applies to SwiftPM packages.</summary>
+    SwiftPM,
 }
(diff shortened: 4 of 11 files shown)


.github/workflows/snapshot-publish.yml
```diff
@@ -2,6 +2,8 @@ name: Publish snapshot of test scan
 
 env:
   CD_DETECTOR_EXPERIMENTS: 1
+  PipReportSkipFallbackOnFailure: "true"
+  PIP_INDEX_URL: "https://pypi.python.org/simple"
```
raporpe, line 6: The snapshot reference for the verification tests needs to be updated with the same environment as the verification pipeline or otherwise it will fail.

raporpe:
> @raporpe please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree company="Microsoft"

.github/workflows/snapshot-publish.yml
```diff
@@ -54,7 +56,7 @@ jobs:
         run:
           dotnet run scan --Verbosity Verbose --SourceDirectory ${{ github.workspace }}/test/Microsoft.ComponentDetection.VerificationTests/resources --Output ${{ github.workspace }}/output
           --DockerImagesToScan "docker.io/library/debian@sha256:9b0e3056b8cd8630271825665a0613cc27829d6a24906dc0122b3b4834312f7d,mcr.microsoft.com/cbl-mariner/base/core@sha256:c1bc83a3d385eccbb2f7f7da43a726c697e22a996f693a407c35ac7b4387cd59,docker.io/library/alpine@sha256:1304f174557314a7ed9eddb4eab12fed12cb0cd9809e4c28f29af86979a3c870"
-          --DetectorArgs DockerReference=EnableIfDefaultOff,SPDX22SBOM=EnableIfDefaultOff,SimplePip=EnableIfDefaultOff
+          --DetectorArgs DockerReference=EnableIfDefaultOff,SPDX22SBOM=EnableIfDefaultOff
```
raporpe, line 59: I removed this line so that both snapshot generation and verification are aligned. `SimplePip` was disabled from the verification pipeline in PR #1259

src/Microsoft.ComponentDetection.Detectors/swiftpm/SwiftPMResolvedComponentDetector.cs
```diff
@@ -0,0 +1,105 @@
...
+        IComponentStreamEnumerableFactory componentStreamEnumerableFactory,
+        IObservableDirectoryWalkerFactory walkerFactory,
+        ILogger<SwiftPMResolvedComponentDetector> logger)
+    {
+        this.ComponentStreamEnumerableFactory = componentStreamEnumerableFactory;
+        this.Scanner = walkerFactory;
+        this.Logger = logger;
+    }
+
+    public override string Id { get; } = "SwiftPM";
```
grvillic, line 29: The `PM` suffix seems superfluous

raporpe: I added the PM suffix because the Swift Package Manager community usually adds it to differentiate the Swift language from the package manager. This is the start of the apple documentation:

image

Maybe I could name it just `swift`. Looking at the codebase ,`go` is the name for go packages so it makes sense to just leave it swift.

src/Microsoft.ComponentDetection.Detectors/swiftpm/SwiftResolvedComponentDetector.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    public override string Id { get; } = "Swift";
+
+    public override IEnumerable<string> Categories => [Enum.GetName(DetectorClass.Swift)];
+
+    public override IList<string> SearchPatterns { get; } = ["Package.resolved"];
+
+    public override IEnumerable<ComponentType> SupportedComponentTypes => [ComponentType.Swift];
+
+    public override int Version => 2;
```
grvillic, line 37: since this detector is just being introduced its version can be `

src/Microsoft.ComponentDetection.Detectors/swiftpm/SwiftResolvedComponentDetector.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    private void ProcessPackageResolvedFile(ISingleFileComponentRecorder singleFileComponentRecorder, IComponentStream componentStream)
+    {
+        var parsedResolvedFile = this.ReadAndParseResolvedFile(componentStream.Stream);
+
+        foreach (var package in parsedResolvedFile.Pins)
+        {
+            // We are only interested in packages coming from remote sources such as git
+            // The Package Kind is not an enum because the Swift Package Manager contract does not specify the possible values.
+            var targetSwiftPackageKind = "remoteSourceControl";
```
grvillic, line 64: let's make this a private constant of this class.

src/Microsoft.ComponentDetection.Detectors/swiftpm/SwiftResolvedComponentDetector.cs
```diff
@@ -0,0 +1,105 @@
...
+                    version: version,
+                    packageUrl: package.Location,
+                    hash: package.State.Revision);
+                var newDetectedSwiftComponent = new DetectedComponent(component: detectedSwiftComponent, detector: this);
+                singleFileComponentRecorder.RegisterUsage(newDetectedSwiftComponent);
+
+                // We also register a Git component for the same package so that the git URL is registered.
+                // Swift Package Manager directly downloads the package from the git URL.
+                var detectedGitComponent = new GitComponent(
+                    repositoryUrl: new Uri(package.Location),
```
grvillic, line 81: is this location guaranteed to be a valid Git Url? I ask because if this is missing or an unexpected location it will effectively break the parsing of all Swift components in that scan.

src/Microsoft.ComponentDetection.Detectors/swiftpm/SwiftResolvedComponentDetector.cs
RushabhBhansali: version can be `1`

raporpe: Fixing, thank you for catching it

src/Microsoft.ComponentDetection.Detectors/swiftpm/SwiftResolvedComponentDetector.cs
raporpe: Changing, thanks

src/Microsoft.ComponentDetection.Detectors/swiftpm/SwiftResolvedComponentDetector.cs
raporpe: When developing the detectector I tried to look into the SwiftPM documentation and found not format guarantee that this is a git url. Swift expects a repository when the kind is "remoteSourceControl". Anyway, I'm going to take a look into the SwiftPM source code to see what it's actually doing

raporpe: Ok I found the documentation that I needed: 
Screenshot 2025-01-27 at 18 23 12

 and 

Screenshot 2025-01-27 at 18 19 45

and in the source code they are verifying that the URL is a valid git repository because SwiftPM depends on git tags to be able to download the specified version.

This is the source code where they are doing it:

Screenshot 2025-01-27 at 18 18 26
Screenshot 2025-01-27 at 18 18 43

So yes, I would say it's very very probable that this is almost always going to be a valid git url for most cases.

These are all the available "Kinds" of packages and I'm covering one of the most common, which is the git url. I think that SwiftPM registries are not that common.

Screenshot 2025-01-27 at 17 53 26

raporpe: As you told me in teams, I will make sure to enclose each for loop iteration with a try so that if a single dependency fails the rest of dependencies will be registered 👍🏼

src/Microsoft.ComponentDetection.Contracts/TypedComponent/SwiftComponent.cs
```diff
@@ -0,0 +1,50 @@
...
+    public override ComponentType Type => ComponentType.Swift;
+
+    public override string Id => $"{this.Name} {this.Version} - {this.Type}";
+
+    // The type is swiftpm
+    public PackageURL PackageURL => new PackageURL(
+        type: "swift",
+        @namespace: new Uri(this.packageUrl).Host,
+        name: this.Name,
+        version: this.hash, // Hash has priority over version when creating a PackageURL
```
grvillic, line 44: have you confirmed in PURL spec hash is what is used as opposed to version? These are not meant to be arbitrary. We should be aligned with the spec has defined for Swift, I don't recall seeing hashes for purls in the past. See sample [Swift PURL](https://github.com/package-url/packageurl-swift) and [purl-spec](https://github.com/package-url/purl-spec?tab=readme-ov-file#known-implementations)

raporpe: When I was implementing this part I remember that Purl did not define precisely what the "namespace" should be. After thinking for a while, I came up with these two options:
1. Since this is a git repository that points to a url, the "namespace" should be the hostname. Simple and straightforward.
2. However, if we are pointing to *github.com/apple/swift* then the namespace should include "apple" and also "github.com". This makes more sense than option 1 because otherwise we would be ommiting "apple" and two users could have the package "swift":  *github.com/apple/swift* and *github.com/microsoft/swift*.

In the end I chose option 1 because Github is a subset of git repositories and I we have to generalize this. The following repository could exist (in theory): applegit.com/swift and the option 2 would consider that *applegit.com/swift* is the namespace when in this case the namespace is only the hostname *applegit.com*.

I reviewed the links that you sent me and the Swift implementation does not actually explain what should we consider a namespace, it is only a general Purl parser. So I decided to look into some database that uses Purl and tried to look for a swift package: https://osv.dev/vulnerability/GHSA-w8xv-rwgf-4fwh

What I found is that for the swift package at *github.com/apple/swift-asn1*, the corresponding Purl is `pkg:swift/github.com/apple/swift-asn1`, where namespace=github.com/apple.
This opens an interesting question, since not all git repositories will have the "path structure" of github.

> Should we treat the host as the namespace and make it "easy and simple" or try to read the first path directly within the hostname?

---

Regarding the version, I agree that it is better to pass the "Version" property. However, this "Version" might contain a hash since you can configure the swift package to point to a branch, tag or commit hash. You can see here (https://github.com/microsoft/component-detection/pull/1316/files#diff-ca41ae04f5f5aabdf0e6e7e13e30415d106bcefcead18b96188fa54cbdb905b1R71) that we are prioritising the version/branch before the hash.

src/Microsoft.ComponentDetection.Detectors/swiftpm/SwiftResolvedComponentDetector.cs
```diff
@@ -0,0 +1,113 @@
...
+                    var newDetectedSwiftComponent = new DetectedComponent(component: detectedSwiftComponent, detector: this);
+                    singleFileComponentRecorder.RegisterUsage(newDetectedSwiftComponent);
+
+                    // We also register a Git component for the same package so that the git URL is registered.
+                    // Swift Package Manager directly downloads the package from the git URL.
+                    var detectedGitComponent = new GitComponent(
+                        repositoryUrl: new Uri(package.Location),
+                        commitHash: package.State.Revision,
+                        tag: version);
+                    var newDetectedGitComponent = new DetectedComponent(component: detectedGitComponent, detector: this);
```
grvillic, line 87:
```suggestion
                    var newDetectedGitComponent = new DetectedComponent(component: detectedGitComponent);
```
I don't think we need to set the detector; it will automatically be captured by the graph. I believe this was just a legacy param.

src/Microsoft.ComponentDetection.Contracts/TypedComponent/SwiftComponent.cs
grvillic: > Should we treat the host as the namespace and make it "easy and simple" or try to read the first path directly within the hostname?

If it isn't documented, I would stick with "easy and simple" approach, so let's treat host as namespace. I would also recommend raising an issue in purl spec to ask for clarification and input from folks. We likely won't be the first ones asking this question.

> that we are prioritizing the version/branch before the hash

The PURL is not touching the version at all, only the hash. In order for this claim to be true, it should be `version: this.Version` if I understand correctly. That way it is version when available, otherwise it will be hash in the PURL.

See [swift type](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#swift).

image

src/Microsoft.ComponentDetection.Detectors/swiftpm/SwiftResolvedComponentDetector.cs
raporpe: Oh, yes, it seems to be a legacy param (the rest of the project is not using it) although it's not mentioned in the documentation. Removing...

src/Microsoft.ComponentDetection.Contracts/TypedComponent/SwiftComponent.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            { "repository_url", this.packageUrl.AbsoluteUri },
+        },
+        subpath: null);
+
+    private string GetNamespaceFromPackageUrl()
+    {
+        // In the case of github.com, the namespace should contain the user/organization
+        // See https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#swift
+        if (this.packageUrl.Host.Contains("github.com"))
```
grvillic, line 59: Should we make this case insensitive?

src/Microsoft.ComponentDetection.Contracts/TypedComponent/SwiftComponent.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            return this.packageUrl.Host + "/" + this.packageUrl.Segments[1].Trim('/');
```
grvillic, line 61: We can use string interpolation here

raporpe: yes maybe it's more readable

src/Microsoft.ComponentDetection.Contracts/TypedComponent/SwiftComponent.cs
raporpe: fixed

raporpe: `.Host` performs some normalizations regarding uppercase but it's not mentioned in the documentation so performing the string normalization by ourselves could still be valuable. See the test I added [here](https://github.com/microsoft/component-detection/pull/1316/commits/39705f65f61d4d8cf58cf7f3855d217d7185bcd5).

grvillic (approved):

PR #1320, opened 2024-12-04
Revert to treating PackageDownload as Development Dependency

ericstj:
While we do know that some PackageDownloads are used at runtime and have active alerts against them - the user action to fix those isn't centered around updating the dependency, it's instead about updating the SDK. As such, we feel that treating them as non-development dependencies now may result in user confusion and lot lead to the right result.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/nuget/NuGetPackageReferenceFrameworkAwareDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/nuget/NuGetPackageReferenceFrameworkAwareDetector.cs
@@ -192,13 +192,12 @@ public class NuGetPackageReferenceFrameworkAwareDetector : FileComponentDetector
 
                 var libraryComponent = new DetectedComponent(new NuGetComponent(packageDownload.Name, packageDownload.VersionRange.MinVersion.ToNormalizedString()));
 
-                // Conservatively assume that PackageDownloads are not develeopment dependencies even though usage will not effect any runtime behavior.
-                // Most often they are used for some runtime deployment -- runtime packs, host packs, AOT infrastructure, etc, so opt in treating them as non-development-dependencies.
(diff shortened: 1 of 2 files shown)


grvillic (approved):

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetPackageReferenceFrameworkAwareDetector.cs
```diff
@@ -192,13 +192,12 @@ private void RegisterPackageDownloads(ISingleFileComponentRecorder singleFileCom
...
 
-                // Conservatively assume that PackageDownloads are not develeopment dependencies even though usage will not effect any runtime behavior.
-                // Most often they are used for some runtime deployment -- runtime packs, host packs, AOT infrastructure, etc, so opt in treating them as non-development-dependencies.
+                // PackageDownload is always a development dependency since it's usage does not make it part of the application
                 singleFileComponentRecorder.RegisterUsage(
                     libraryComponent,
                     isExplicitReferencedDependency: true,
                     parentComponentId: null,
-                    isDevelopmentDependency: false,
+                    isDevelopmentDependency: true,
```
grvillic, line 200: Do we have any tests for this? I was expecting something to break with this type of change, we should probably at one that validates PackageDownloads always go as dev dependencies.

grvillic (changes requested):
Adding some tests to avoid regressions will help here.

src/Microsoft.ComponentDetection.Detectors/nuget/NuGetPackageReferenceFrameworkAwareDetector.cs
ericstj: We test that they appear but not whether they should be devDependencies or not.  I've added a test.

grvillic (approved):

PR #1328, opened 2024-12-18
Handle unauthorized files / dirs for cleanup

pauld-msft:
Issue reported by customer. Agents that do not even have read access to some parts of the file system can run into issues when trying to determine the files and directories that need to be cleaned up.

Fix: error handle the unauthorized exception, and skip cleanup for those cases


src/Microsoft.ComponentDetection.Contracts/FileComponentDetectorWithCleanup.cs
```diff
@@ -56,80 +56,78 @@ protected async Task WithCleanupAsync(
         // determining the files that exist as there will be no subsequent cleanup process.
         if (this.FileUtilityService == null
             || this.DirectoryUtilityService == null
-            || !this.TryGetCleanupFileDirectory(processRequest, out var fileParentDirectory))
+            || !this.TryGetCleanupFileDirectory(processRequest, out var fileParentDirectory)
+            || !cleanupCreatedFiles)
```
RushabhBhansali, line 60: can we add a unit test : return early if the `cleanupCreatedFiles` is false.

src/Microsoft.ComponentDetection.Contracts/FileComponentDetectorWithCleanup.cs
```diff
@@ -151,4 +149,19 @@ private bool TryGetCleanupFileDirectory(ProcessRequest processRequest, out strin
...
+    private (bool Success, HashSet<string> Files, HashSet<string> Directories) TryGetFilesAndDirectories(string root, IList<string> patterns, int depth)
+    {
+        try
+        {
+            var (files, directories) = this.DirectoryUtilityService.GetFilesAndDirectories(root, patterns, depth);
+            return (true, files, directories);
+        }
+        catch (UnauthorizedAccessException e)
+        {
+            // ignore files and directories that we don't have access to
```
RushabhBhansali, line 162: can we mock this exception and make sure that this is covered in a unit test ?

src/Microsoft.ComponentDetection.Contracts/FileComponentDetectorWithCleanup.cs
pauld-msft: That flag was covered to ensure that its not cleaning up if that flag is set to false, added an extra check to ensure that we don't even try to get files/directories though

src/Microsoft.ComponentDetection.Contracts/FileComponentDetectorWithCleanup.cs
pauld-msft: done

RushabhBhansali (approved):

PR #1343, opened 2025-01-20
Align environment between snapshot publish and verify

raporpe:
The snapshot publish pipeline is missing two env variables that change the behavior of pip and might generate different results because of this.

*snapshot-verify.yml*
https://github.com/microsoft/component-detection/blob/main/.github/workflows/snapshot-verify.yml#L5

*snapshot-publish.yml*
https://github.com/microsoft/component-detection/blob/main/.github/workflows/snapshot-publish.yml#L4

### DIFF
--- a/.github/workflows/snapshot-publish.yml
+++ b/.github/workflows/snapshot-publish.yml
@@ -2,6 +2,8 @@ name: Publish snapshot of test scan
 
 env:
   CD_DETECTOR_EXPERIMENTS: 1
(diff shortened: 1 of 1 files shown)


raporpe:
After merging this PR, please do not forget to rerun the snapshot publish pipeline so that the changes of this PR are reflected when running the verification pipeline

raporpe:
@FernandoRojo you modified the snapshot publish pipeline at #1294, could you give me some feedback?

PR #1348, opened 2025-01-22
Register VS Code built-in packages as skipped

rzhao271:
This PR modifies the existing NPM component detector to skip components corresponding to VS Code built-in extensions in order to reduce false positives while scanning the VS Code repository.

Example component that should not be flagged: https://github.com/microsoft/vscode/blob/main/extensions/handlebars/package.json

#### New Detector Checklist

- [x] I have gone through the docs for creating a new detector [here](https://github.com/microsoft/component-detection/blob/main/docs/creating-a-new-detector.md)
- [ ] My new detector implements `IDefaultOffComponentDetector`
- [ ] I have created a PR to the [verification repo](https://github.com/microsoft/componentdetection-verification) with components that my new detector can find
- [ ] (If necessary) I have updated the [Feature Overview](https://github.com/microsoft/component-detection/blob/main/README.md#feature-overview) table in the README

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/npm/NpmComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/npm/NpmComponentDetector.cs
@@ -79,10 +79,18 @@ public class NpmComponentDetector : FileComponentDetector
         var name = packageJToken["name"].ToString();
         var version = packageJToken["version"].ToString();
         var authorToken = packageJToken["author"];
+        var enginesToken = packageJToken["engines"];
 
         if (!SemanticVersion.TryParse(version, out _))
         {
             this.Logger.LogWarning("Unable to parse version {NpmPackageVersion} for package {NpmPackageName} found at path {NpmPackageLocation}. This may indicate an invalid npm package component and it will not be registered.", version, name, filePath);
+            singleFileComponentRecorder.RegisterPackageParseFailure($"{name} - {version}");
+            return false;
+        }
+
+        if (enginesToken != null && enginesToken["vscode"] != null)
+        {
+            this.Logger.LogInformation("{NpmPackageName} found at path {NpmPackageLocation} represents a built-in VS Code extension. This package will not be registered.", name, filePath);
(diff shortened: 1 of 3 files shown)


src/Microsoft.ComponentDetection.Detectors/npm/NpmComponentDetector.cs
```diff
@@ -87,6 +88,13 @@ protected virtual bool ProcessIndividualPackageJTokens(string filePath, ISingleF
             return false;
         }
 
+        if (enginesToken != null && enginesToken["vscode"] != null)
+        {
+            this.Logger.LogInformation("{NpmPackageName} found at path {NpmPackageLocation} represents a built-in VS Code extension. This package will not be registered.", name, filePath);
+            singleFileComponentRecorder.RegisterPackageParseFailure($"{name} - {version}");
```
grvillic, line 94: Why is this treated as a failure? Seems like we will only log and return false, but no need to record it as a failure.

rzhao271: I wanted to figure out how to mark it as a skipped component, and saw https://github.com/rzhao271/component-detection/blob/fffd04ca4130f4e206e08c1f1507e11cd4e3e2cc/src/Microsoft.ComponentDetection.Common/DependencyGraph/ComponentRecorder.cs#L204. I'll remove the line.

grvillic (approved):

src/Microsoft.ComponentDetection.Detectors/npm/NpmComponentDetector.cs
```diff
@@ -87,6 +88,12 @@ protected virtual bool ProcessIndividualPackageJTokens(string filePath, ISingleF
             return false;
         }
```
grvillic, line 90: Since you are changing the graph generated by the detector you need to bump the version of the detector see line 43, it should be bumped to 3.

`public override int Version { get; } = 2;`

grvillic (approved):

PR #1359, opened 2025-02-13
Users/jcfiorenzano23/go 117 experiment

jcfiorenzano:
This PR introduces a new experimental detector for go. The main difference of this detector with respect to the usual one is that only relies on the information coming in the go.mod file to parse the components and go.mod after 1.17 introduces all types of dependencies, not only direct ones. This is already supported by the default go detector but only if fallback is enabled as the default behavior is to parse the result coming from the go list command.

This detector also adds the characteristic that also uses the go graph command to create the graph of dependencies between the components. This feature is only available in the default detector when parsing the information coming from the CLI.

To simplify the sharing of code between both detectors, the PR also refactored the default detector and extract the different parsers to its own file:
- GoModParser
- GoSumParser
- GoCLIParser


src/Microsoft.ComponentDetection.Detectors/go/Go117ComponentDetector.cs
```diff
@@ -0,0 +1,204 @@
...
+
+    private async Task<bool> ShouldRunGoGraphAsync()
+    {
+        var goVersion = await this.GetGoVersionAsync();
+        if (goVersion == null)
+        {
+            return false;
+        }
+
+        return goVersion >= new Version(1, 11);
```
grvillic, line 158: We should also respect the [DisableGoCliScan](https://github.com/microsoft/component-detection/blob/main/docs/detectors/go.md#fallback-detection-strategy) variable here, if customer explicitly asked for no CLI, then we should not generate the graph (edges) between the nodes regardless of Go CLI Version. They will see flat list of components and may be fine for performance reasons they just won't be able to determine root module that brought in the dependency.

grvillic (approved):

PR #1360, opened 2025-02-24
Updated categories of rust detectors to match doc spec

SystemGlitch101:


SystemGlitch101:
## Summary

When running component detection with the `--DetectorCategories Cargo` option, it does not actually run the detector. This fixes the issue by changing the category to match the `DetectorClass.Cargo` enum, instead of being a hard-coded "Rust".

SystemGlitch101:

FernandoRojo (approved):

PR #1368, opened 2025-03-13
Fix link version parsing in pnpm9

FernandoRojo:


src/Microsoft.ComponentDetection.Detectors/pnpm/Pnpm9Detector.cs
```diff
@@ -73,11 +73,12 @@ private void ProcessDependencyList(ISingleFileComponentRecorder singleFileCompon
         {
             var pnpmDependencyPath = this.pnpmParsingUtilities.ReconstructPnpmDependencyPath(name, dep.Version);
             var (component, package) = components[pnpmDependencyPath];
+            var (_, packageVersion) = this.pnpmParsingUtilities.ExtractNameAndVersionFromPnpmPackagePath(pnpmDependencyPath);
```
grvillic, line 76: we should bump the Pnpm Detector version

FernandoRojo: done

grvillic (approved):

PR #1369, opened 2025-03-13
Update manifest.schema.json to include DotNet

grvillic:
## Context
We recently introduced a new detector but we missed updating the manifest schema due to noisy failures with Pip tests, which will be fixed asap.


ericstj (approved):
Thanks for adding this, sorry I missed it.

FernandoRojo (approved):

PR #1371, opened 2025-03-15
Move null guard statement up in NuGetComponentDetector.IsValidPath to avoid uncessary throw of ArgumentNullException in FileInfo constructor

WesHaze:
Moves guard statement up in the NuGetComponentDetector to avoid unnecessary throw in the FileInfo constructor


WesHaze:
> Thanks for submitting this!

No worries, thank you guys too!  Do I need to take further action or will it get merged eventually, presumably when your colleague gets around to it and tests are run? I do not see a merge button.

PR #1374, opened 2025-03-20
Remove incomplete ELF Syft results for Mariner 2.0

pauld-msft:
Mariner 2.0 images do not contain a specific enough version in its ELF notes. https://github.com/microsoft/azurelinux/pull/10405 fixes it for azure linux 3.0, but the change is unable to be ported to 2.0. The packages we skip here are detected with the release version elsewhere in the image, so we are safe to remove the incomplete duplicates.


grvillic:
Let's bump the detector version since we are changing the dependency graph captured by Linux.

grvillic (approved):

PR #1379, opened 2025-03-28
Fix pnpm 9 link dependency bug in detector

jasonpaulos:
I noticed that some pnpm dependencies were being incorrectly marked as root dependencies from a v9 lockfile.

After debugging, I realized the problem was that local dependencies referenced with `workspace:`/`link:` do not appear in the `snapshots` section of the lockfile. As a result, when the detector tries to look them up in the `components` dictionary, an exception is thrown and the rest of the file stops being processed. Because of this, the detector was unable to finish building the dependency graph and incorrectly assumed many dependencies were root dependencies because their ancestor information was never populated.

In order to expose this case in the existing unit test for link dependencies, I added another dependency after `SampleLinkDependency`. Prior to this change, the detector would throw when processing `SampleLinkDependency` and then be unable to mark `sampleDevDependency` as an ancestor of `sampleIndirectDevDependency`.


jasonpaulos:
@microsoft-github-policy-service agree company="Microsoft"

JamieMagee:
The verification test failures all appear to be related to Python dependencies. I am assuming it's just an out-of-date cache for those tests.

JamieMagee (approved):

pauld-msft (approved):

PR #1401, opened 2025-05-06
Improve detection graph traversal for large result sets

pauld-msft:
When a large number of results are returned from detection, there was a massive slowdown in `GatherSetOfDetectedComponentsUnmerged`. Investigation proved that the bottleneck was during the `componentRecorder.GetComponent` call that was invoked to convert a string to a `TypedComponent` when adding the roots and ancestors of each component to the result set:
```
foreach recorderDetectorPair:
  foreach component:             <-- thousands of these
    foreach relevantDependencyGraph:     <-- thousands of these
      foreach calculateAncestors():
        componentRecorder.GetComponent(ancestor)  <-- slow call invoked for each ancestor
      foreach calculateRoots():
        componentRecorder.GetComponent(root)  <-- slow call invoked for each root
```
This change makes it so that we iterate over all of the components in each dependency graph, and run `componentRecorder.GetComponent` a single time on each component in the dependency graph. Then, we use that `TypedComponent` as the return type for the calculate ancestor/roots call.
```
foreach recorderDetectorPair:
  foreach dependencyGraph:
    foreach component in dependencyGraph:
      component.TypedComponent = componentRecorder.GetComponent()  <-- slow call invoked single time per graph node
  foreach component:
    foreach relevantDependencyGraph:
      calculateAncestorsAsTypedComponents():
      calculateRootsAsTypedComponents():
```

Overall, for a large repository with 3600 Nuget dependency graphs and 2200 Nuget components, this change reduces the total CD runtime from ~28 minutes to ~8 minutes, with 0 change in the resulting manifest file.


src/Microsoft.ComponentDetection.Common/DependencyGraph/DependencyGraph.cs
```diff
@@ -141,6 +142,30 @@ public ICollection<string> GetAncestors(string componentId)
...
+        return this.GetAncestors(componentId)
+            .Select(a => this.componentNodes[a].TypedComponent ?? toTypedComponent(a))
+            .ToHashSet(new ComponentComparer());
+    }
+
+    public HashSet<TypedComponent> GetRootsAsTypedComponents(string componentId, Func<string, TypedComponent> toTypedComponent)
+    {
+        ArgumentNullException.ThrowIfNull(componentId);
+        return this.GetExplicitReferencedDependencyIds(componentId)
+            .Select(r => this.componentNodes[r].TypedComponent ?? toTypedComponent(r))
```
grvillic, line 157: Are we certain `this.componentNodes[r]` will exist, should we check first before `this.componentNodes[r].TypedComponent`?

pauld-msft: the `GetAncestors` and `GetExplicitReferencedDependencyIds` calls do confirm that all of the nodes that are returned are in the componentNodes list, but it definitely doesn't hurt to not make that assumption in case things change in the future. I'll add a quick check

grvillic (approved):

pauld-msft:
> > This change reduces the total CD runtime from ~28 minutes to ~8 minutes, with 0 change in the resulting manifest file.
> 
> What about the before/after for CG runtime itself?

This repo is open source, and details for implementations that consume these binaries are kept out of here. The time savings should still apply for any libraries that are running these component detection binaries.

narasamdya:
> > > This change reduces the total CD runtime from ~28 minutes to ~8 minutes, with 0 change in the resulting manifest file.
> > 
> > 
> > What about the before/after for CG runtime itself?
> 
> This repo is open source, and details for implementations that consume these binaries are kept out of here. The time savings should still apply for any libraries that are running these component detection binaries.

Basically, I was just asking the runtime spent by the CG itself, before and after your changes. It's because you mention the whole CD runtimes, but I'm curious about the CG portion itself.

Issue #1413, opened 2025-05-23
Invalid and incomplete PackageUrl in ConanComponent


RodneyRichardson:
ConanComponent.PackageUrl incorrectly specifies String.Empty for namespace and subpath (instead of null), resulting in something like `pkg:conan//openssl@3.4.1#` instead of `pkg:conan/openssl@3.4.1`. Compare with cargo fix in #1410.

It also does not support all specification options from https://github.com/package-url/purl-spec/blob/main/PURL-TYPES.rst#conan, for example: 

```
pkg:conan/openssl.org/openssl@3.0.3?user=bincrafters&channel=stable
pkg:conan/openssl.org/openssl@3.0.3?arch=x86_64&build_type=Debug&compiler=Visual%20Studio&compiler.runtime=MDd&compiler.version=16&os=Windows&shared=True&rrev=93a82349c31917d2d674d22065c7a9ef9f380c8e&prev=b429db8a0e324114c25ec387bfd8281f330d7c5c
```

This might be practical to fix at the same time as adding Conan 2.0 support (#1228).

JamieMagee:
This is resolved by https://github.com/microsoft/component-detection/pull/1730

PR #1414, opened 2025-05-23
Fix ConanComponent.PackageURL

RodneyRichardson:
Pass `null` instead of `String.Empty` to generate a valid PackageUrl.

Partially addresses #1413


JamieMagee:
No longer required since https://github.com/microsoft/component-detection/pull/1730

PR #1426, opened 2025-06-09
Replace existing Go detector with Go117 detector

AMaini503:
<h2> Changes </h2>

- `GoComponentDetector.cs`
   - Added a try/catch around go cli execution. This was fixed in response to a failing UT.
   - Changed detector version from 8 to 9
   - Rest of the code is copy/paste from `Go117ComponentDetector.cs`
 - `GoComponentDetectorTests.cs`
    - UTs testing `go mod graph` were mocking go list output. This was changed to mock `go.mod` parsing.
    - Copied over all `Go117` unit tests
 - `GoModParser.cs`: This parser wasn't handling replace directives like `go list` would have.
     - Add support to collect all replace directives. Packages that resolve to local paths are ignored. Packages that resolve to other packages are substituted before registration.


src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -14,31 +15,13 @@ namespace Microsoft.ComponentDetection.Detectors.Go;
 using Microsoft.ComponentDetection.Contracts.TypedComponent;
 using Microsoft.Extensions.Logging;
 
-public class GoComponentDetector : FileComponentDetector
+public class GoComponentDetector : FileComponentDetector, IExperimentalDetector
```
grvillic: remove the Experimental Detector interface from here, this detector is already live, it is just being improved with the updates you are deploying.

grvillic: In this PR let's also update the [Go docs](https://github.com/microsoft/component-detection/blob/main/docs/detectors/go.md)

AMaini503: Updated.

AMaini503: Removed all references to Go117

grvillic (approved):

PR #1432, opened 2025-06-11
User/aamaini/go sort requests

AMaini503:
Previous Go detector used to ignore child directories of already processed directories because `go list` in parents already accounted for children.
To keep parity with this behavior, Go Detector preprocesses go.mod, go.sum files in order of increasing depth and keeps track of already processed directories.

- `OnPrepareDetectionAsync` sorts all process requests in order of depth, fullname making it fully deterministic and stable sort.
- `OnFileFoundAsync` now tracks processed project roots if go.mod was successfully parsed and go version in it was >= 1.17
- Add UTs to verify detector skips child nested directories.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
@@ -50,12 +50,12 @@ public class GoComponentDetector : FileComponentDetector
 
     public override int Version => 9;
 
-    protected override Task<IObservable<ProcessRequest>> OnPrepareDetectionAsync(
+    protected async override Task<IObservable<ProcessRequest>> OnPrepareDetectionAsync(
         IObservable<ProcessRequest> processRequests,
         IDictionary<string, string> detectorArgs,
         CancellationToken cancellationToken = default)
     {
-        var goModProcessRequests = processRequests.Where(processRequest =>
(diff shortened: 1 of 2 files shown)


src/Microsoft.ComponentDetection.Detectors/go/GoComponentDetector.cs
```diff
@@ -108,7 +114,20 @@ protected override async Task OnFileFoundAsync(ProcessRequest processRequest, ID
...
+                if (wasModParsedSuccessfully &&
+                    !string.IsNullOrEmpty(record.GoModVersion) &&
+                    System.Version.TryParse(record.GoModVersion, out var goVersion) &&
+                    goVersion >= new Version(1, 17))
+                {
+                    this.projectRoots.Add(projectRootDirectory.FullName);
+                }
+                else
+                {
+                    this.Logger.LogWarning("Not adding {Root} to processed roots: {ParseSuccess} {GoModVersion}", projectRootDirectory.FullName, wasModParsedSuccessfully, record.GoModVersion);
```
grvillic: This can be noisy, I think we can have it as verbose not as a warning.

AMaini503: Fixed.

grvillic (approved):

PR #1443, opened 2025-06-27
Take a dependency on a Rust library?

nathanhammond:
`Cargo.lock` parsing is tedious and needs constant maintenance. It would be nice to use a pre-existing parser to do it. This is a possible solution for loading data from `Cargo.lock` without having to implement a parser.

Mega-detector layers:

1. Lockfile parsing. Can't take runtime dependencies.
2. `cargo metadata`. Requires `cargo` to exist.
3. SBOM parser. Requires a `cargo build` to have been completed.

This code is for Layer 1.

**Current status:** library embedding tech demo. Data reaches C# but isn't consumed by a detector.


grvillic:
Thanks for the contribution, but we will not be introducing custom Rust parsers. All customers should migrate to generating SBOMs if they want accurate detection, otherwise preexisting detectors would do their best effort, with the downside of overreporting.

PR #1459, opened 2025-08-13
Update YamlDotNet from version 15.1.6 -> 16.3.0

dorb-legit:


dorb-legit:
@microsoft-github-policy-service agree company="Legit Security"

Issue #464, opened 2023-03-07
Introduce nullable reference types


JamieMagee:
[Nullable reference types][1] were introduced in C# 8 (.NET Core 3.0) and are enabled by default in new projects created targeting .NET 6. They help to prevent `NullReferenceException`s with 3 main features:

- Improved static flow analysis that determines if a variable may be null before dereferencing it
- Attributes that annotate APIs so that the flow analysis determines null-state
- Variable annotations that developers use to explicitly declare the intended null-state for a variable

We should use the [nullable reference migration guide][2] to plan our migration.

NOTE: Adding nullable reference types is not considered a semver breaking change.

[1]: https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references
[2]: https://learn.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies

PR #1471, opened 2025-10-14
enable nullable by default 

cataggar:
This enabled nullable by default and then adds `#nullable disable` anywhere it is needed in the current code.  This will help ensure new code uses nullable and allows existing code to move to it incrementally. This follows the official guidance for moving to nullable.

Incremental step for 
- #464

Following step 2 in https://learn.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies :

> Nullable enable as the default: Set this default when you're actively developing new features. You want all new code to benefit from nullable reference types and nullable static analysis. Using this default means you must add a #nullable disable to the top of each file. You'll remove these preprocessor directives as you address the warnings in each file.

34359803e21670c02eaa5bb3225d58b7ad6b9694 by Copilot in 
- https://github.com/cataggar/component-detection/pull/7

e9ff160c02ae3ced79cf2e0fea7c1090bbd27565 by Copilot locally.

### DIFF
--- a/src/Microsoft.ComponentDetection.Common/Telemetry/Records/RustDetectionTelemetryRecord.cs
+++ b/src/Microsoft.ComponentDetection.Common/Telemetry/Records/RustDetectionTelemetryRecord.cs
@@ -1,3 +1,4 @@
+#nullable disable
 namespace Microsoft.ComponentDetection.Common.Telemetry.Records;
 
 public class RustDetectionTelemetryRecord : BaseDetectionTelemetryRecord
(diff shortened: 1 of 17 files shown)


src/Microsoft.ComponentDetection.Common/DockerReference/DigestUtility.cs
```diff
@@ -1,4 +1,5 @@
-namespace Microsoft.ComponentDetection.Common;
+#nullable disable
+namespace Microsoft.ComponentDetection.Common.DockerReference;
```
cataggar: Why change the namespace Copilot?

src/Microsoft.ComponentDetection.Common/Utilities/StringUtilities.cs
```diff
@@ -1,4 +1,5 @@
-namespace Microsoft.ComponentDetection.Common;
+#nullable disable
+namespace Microsoft.ComponentDetection.Common.Utilities;
```
cataggar: Bad Copilot. another random namespace change.

cataggar:
The CI build is currently failing due to:
- #1473

JamieMagee:
@cataggar if you can rebase this branch, I can review and merge this.

JamieMagee (approved):

PR #1474, opened 2025-10-20
Unify Rust detection logic with SBOM mode, ownership mapping, and skip optimizations

AMaini503:
# Unified Rust Detector – SBOM, Fallback, and Ownership Enhancements

## Summary
This PR introduces a **unified Rust component detection model** that consolidates SBOM, CLI, and Lock-based detection strategies into a single consistent implementation.  
It also refactors supporting classes and introduces key performance and accuracy improvements.

### 1. Unified Detection Model
- When `*.cargo-sbom.json` files are found, detection runs purely in **SBOM mode** (most accurate).
- If no SBOMs are found, the detector operates in **fallback mode**:
  - Prefer `cargo metadata` on `Cargo.toml`.
  - If CLI is disabled or fails, fall back to equivalent `Cargo.lock` processing.
- Ensures a single consistent pipeline while retaining accuracy and resilience.

### 2. Parser Refactoring
- Extracted parsing logic for SBOM, CLI, and Lock files into separate interfaces for improved testability and maintainability.

### 3. `RustContextMetadataBuilder` Enhancements
- SBOMs don’t natively indicate which `Cargo.toml` introduced each dependency.  
- This class now runs `cargo metadata` to map packages to their corresponding TOMLs.
- Enables ownership attribution to user-owned source files.

### 4. Skip Optimizations
- When `cargo metadata` is executed on a workspace root TOML, it inherently includes all member packages.
- The detector now skips redundant member TOML processing, improving scan efficiency and avoiding duplicate work.

### 5. Unit Test Coverage
- Added comprehensive tests for:
  - All Rust parsers (SBOM, CLI, Lock)
  - All Rust detectors
  - `RustContextMetadataBuilder`
- Verified behavioral parity with legacy detectors.

## Validation

Manual validation confirms the unified detector produces identical results to the legacy detectors in both modes:

| Test Repo | Old SBOM | New SBOM (SBOM Mode) | Old CLI | New SBOM (Fallback Mode) |
|------------|-----------|----------------------|----------|---------------------------|
| Repo1 | 151 | 151 | 404 | 404 |
| Repo2 | 489 | 489 | 590 | 590 |

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/rust/RustSbomDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/rust/RustSbomDetector.cs
@@ -21,6 +21,9 @@ using Tomlyn.Model;
 /// </summary>
 public class RustSbomDetector : FileComponentDetector, IExperimentalDetector
 {
+    private const string CargoTomlFileName = "Cargo.toml";
+    private const string CargoLockFileName = "Cargo.lock";
+
     private static readonly TomlModelOptions TomlOptions = new TomlModelOptions
     {
         IgnoreMissingProperties = true,
@@ -118,10 +121,10 @@ public class RustSbomDetector : FileComponentDetector, IExperimentalDetector
     public override IEnumerable<ComponentType> SupportedComponentTypes => [ComponentType.Cargo];
 
     /// <inheritdoc />
-    public override int Version => 1;
+    public override int Version => 2;
 
     /// <inheritdoc />
-    public override IList<string> SearchPatterns { get; } = ["Cargo.toml", "Cargo.lock", "*.cargo-sbom.json"];
+    public override IList<string> SearchPatterns { get; } = [CargoTomlFileName, CargoLockFileName, "*.cargo-sbom.json"];
 
     /// <inheritdoc />
     protected override async Task<IObservable<ProcessRequest>> OnPrepareDetectionAsync(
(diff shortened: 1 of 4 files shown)


src/Microsoft.ComponentDetection.Detectors/rust/RustSbomDetector.cs
```diff
@@ -3,124 +3,708 @@ namespace Microsoft.ComponentDetection.Detectors.Rust;
...
+            // FALLBACK mode: Select Cargo.toml and Cargo.lock files
+            // Order: TOML before LOCK, then depth ascending, then path ascending
+            filteredRequests = allRequests
+                .Where(r =>
+                {
+                    var fileName = Path.GetFileName(r.ComponentStream.Location);
+                    return fileName.Equals("Cargo.toml", this.pathComparison) ||
+                           fileName.Equals("Cargo.lock", this.pathComparison);
+                })
+                .OrderBy(r => Path.GetFileName(r.ComponentStream.Location).Equals("Cargo.lock", this.pathComparison) ? 1 : 0) // TOML before LOCK
```
grvillic: nit: should be create const string for these file names so we can share between its references

AMaini503: Fixed.

grvillic (approved):

PR #1475, opened 2025-10-21
Handle multiple trailing directory separators

ericstj:
We were faced with paths coming from project.assets.json with multiple trailing directory separators:
```
##[warning]Project output path /__w/1/s/artifacts/obj// specified by /mnt/vss/_work/1/s/artifacts/obj/project.assets.json does not exist.
```
When normalizing the path we'd remove one of the trailing slashes, but not both.  Later on:
```
##[debug]Attempting to rebase /__w/1/s/artifacts/obj/ to /mnt/vss/_work/1/s/artifacts/obj using relative artifacts/obj
##[warning]Project path /__w/1/s/eng/icu.proj specified by /mnt/vss/_work/1/s/artifacts/obj/project.assets.json does not exist
```
We couldn't rebase the path because of the extra trailing slash, so we don't find the real path to the project relative to the scan directory.

As a result we'd report the component as coming from some path outside the scan directory --
```
--- Component: ---
9.0.305 net45 unknown - DotNet
--- Found at: ---
/../../../../../__w/1/s/eng/icu.proj
```

Since its outside the scan directory, we couldn't locate the project to probe for the .NET SDK, honoring `global.json`.  As a result we report the machine-wide SDK version that's preinstalled on the machine running component detection.

After the fix this extra project will correctly be mapped and the correct SDK will be reported.
```
--- Component: ---
10.0.100-rc.1.25420.111 net45 unknown - DotNet
--- Found at: ---
/eng/icu.proj
```

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/dotnet/DotNetComponentDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/dotnet/DotNetComponentDetector.cs
@@ -58,9 +58,22 @@ public class DotNetComponentDetector : FileComponentDetector
 
     public override IEnumerable<string> Categories => ["DotNet"];
 
+    private static string TrimAllEndingDirectorySeparators(string path)
+    {
+        string last;
+
+        do
+        {
+            last = path;
+            path = Path.TrimEndingDirectorySeparator(last);
+        }
+        while (!ReferenceEquals(last, path));
+
+        return path;
+    }
+
     [return: NotNullIfNotNull(nameof(path))]
-    private string? NormalizeDirectory(string? path) => string.IsNullOrEmpty(path) ? path :
-        this.pathUtilityService.NormalizePath(path).TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+    private string? NormalizeDirectory(string? path) => string.IsNullOrEmpty(path) ? path : TrimAllEndingDirectorySeparators(this.pathUtilityService.NormalizePath(path));
 
     /// <summary>
     /// Given a path under sourceDirectory, and the same path in another filesystem,


ericstj:
Now I remember why I used that method :-D  It's to avoid trimming the root.  I'll adjust the fix.

src/Microsoft.ComponentDetection.Detectors/dotnet/DotNetComponentDetector.cs
```diff
@@ -58,8 +58,22 @@ public DotNetComponentDetector(
...
+
+        do
+        {
+            last = path;
+            path = Path.TrimEndingDirectorySeparator(last);
+        }
+        while (!ReferenceEquals(last, path));
+
+        return path;
+    }
```
akoeplinger, line 73: I wonder if it'd be easier to just do `path.TrimEnd(Path.DirectorySeparatorChar).TrimEnd(Path.AltDirectorySeparatorChar)`

ericstj: No, we cannot do that because it would trim the root separator for paths like `/` and `C:\`.  We'd need to bake in the platform-specific root path logic or call `GetPathRoot` while still handling cases where a root path had extra separators (`C:\\`).  There's enough complexity in those approaches that this felt cleaner.  It would really be nice if we had the API for this...

akoeplinger: yeah. I wonder why `Path.TrimEndingDirectorySeparator` wasn't implemented that way in the first place...

ericstj: I suspect it wasn't dealing with user inputs, but platform inputs with a known single extra separator.

Issue #1476, opened 2025-10-23
locationsFoundAt specifies URL-encoded paths


lxop:
As per the title, e.g.:
```json
{
  "dependencyGraphs": {
    "C:\\src\\my project\\my project.csproj": {
      ...
    }
  },
  "componentsFound": [
    {
      "locationsFoundAt": [
        "/my%20project/my%20project.csproj"
      ],
      ...
    }
  ]
}
```

This behaviour means that paths including spaces are mangled. This affects downstream uses like https://github.com/advanced-security/component-detection-dependency-submission-action - because the paths are different between "dependencyGraphs" and "componentsFound", it fails to link them, and I get a bunch of warnings like
```txt
 Warning: No dependency graph entry found for manifest location: my%20project/my%20project.csproj
```

I have identified where this URL-encoding is being applied, and have a fix that I will push soon.

PR #1477, opened 2025-10-23
Unescape componentsFound.locationsFoundAt

lxop:
The use of the Uri class was mangling paths that include spaces.

Fixes: #1476


lxop:

PR #1478, opened 2025-10-27
Cleanup redundant rust detectors, add doc for rust detection

AMaini503:
- `RustCli`, `RustCrate` detection has been merged into `RustSbom` detector. Detectors, experiements related to redundant detectors is removed.
- Add doc on rust detection implemented by `RustSbomDetector`.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/rust/RustSbomDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/rust/RustSbomDetector.cs
@@ -9,6 +9,7 @@ using System.Reactive.Threading.Tasks;
 using System.Threading;
 using System.Threading.Tasks;
 using global::DotNet.Globbing;
+using Microsoft.ComponentDetection.Common.Telemetry.Records;
 using Microsoft.ComponentDetection.Contracts;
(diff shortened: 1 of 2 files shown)


src/Microsoft.ComponentDetection.Detectors/rust/RustSbomDetector.cs
pauld-msft, line 161: Not related to this pull request at all, but just for a general FYI, we won't collect this telemetry, instead its written to the console. If you want to be able to measure the impact of this feature / gather telemetry on how many detections are using SBOM vs. Cli fallback vs. static file fallback (which can include the duration data of the detector execution), consider adding a [new telemetry point such as seen in the PipReport detector](https://github.com/microsoft/component-detection/blob/e743edd33276bfc7f3798ede7fe442929baa201f/src/Microsoft.ComponentDetection.Detectors/pip/PipReportComponentDetector.cs#L167-L170), that logs on disposal due to its [base class](https://github.com/microsoft/component-detection/blob/2fd5d4fa4dee02262b772a55ef8ecbba812ff2e1/src/Microsoft.ComponentDetection.Common/Telemetry/Records/BaseDetectionTelemetryRecord.cs#L41-L42).

PR #1523, opened 2025-10-30
Migrate some tests from `Newtonsoft.Json` to `System.Text.Json`

JamieMagee:
This pull request contains the following changes:
- Migration of some tests from `Newtonsoft.Json.JsonConvert` to `System.Text.Json.JsonSerializer`
- Migration of `PipInstallationReport` from `Newtonsoft.Json` to `System.Text.Json`
  - No intermediate step was possible as it used `JObject` and `JArray` properties directly which cannot be handled by `System.Text.Json`
- Made parameterless constructors in the `TypedComponent` hierarchy `public` instead of `private` or `internal`
  - See [Non-public, parameterless constructors not used for deserialization][1]

Part of #231 

[1]: https://learn.microsoft.com/en-us/dotnet/core/compatibility/serialization/5.0/non-public-parameterless-constructors-not-used-for-deserialization

### DIFF
--- a/src/Microsoft.ComponentDetection.Contracts/TypedComponent/CargoComponent.cs
+++ b/src/Microsoft.ComponentDetection.Contracts/TypedComponent/CargoComponent.cs
@@ -7,7 +7,7 @@ using PackageUrl;
 
 public class CargoComponent : TypedComponent
 {
-    private CargoComponent()
+    public CargoComponent()
     {
         // reserved for deserialization
     }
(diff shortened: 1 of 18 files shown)


test/Microsoft.ComponentDetection.Detectors.Tests/PyPiClientTests.cs
```diff
@@ -124,7 +124,7 @@ public async Task GetReleases_DifferentEntries_CallsGetAsync_OnceAsync()
         await action.Should().NotThrowAsync();
         await action.Should().NotThrowAsync();
 
-        // Verify the API call was performed only once
```
ryanbrandenburg, line 127: Haha, classic.

PR #1530, opened 2025-11-14
Pass `CancellationToken` from commands

JamieMagee:
One of the new features in `Spectre.Console.Cli` was support for top-level `CancellationToken` support. This means, that we will be able to pipe cancellation support the entire way through Component Detection. This is a first trial run, passing the `CancellationToken` through to `ScanExecutionService` and `DetectorProcessingService`.

Additionally, I added XML documentation and migrated to primary constructors.

### DIFF
--- a/src/Microsoft.ComponentDetection.Orchestrator/Commands/ScanCommand.cs
+++ b/src/Microsoft.ComponentDetection.Orchestrator/Commands/ScanCommand.cs
@@ -44,7 +44,7 @@ public sealed class ScanCommand : AsyncCommand<ScanSettings>
         CancellationToken cancellationToken)
     {
         this.fileWritingService.Init(settings.Output);
-        var result = await this.scanExecutionService.ExecuteScanAsync(settings);
+        var result = await this.scanExecutionService.ExecuteScanAsync(settings, cancellationToken);
         this.WriteComponentManifest(settings, result);
         return 0;
     }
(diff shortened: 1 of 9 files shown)


src/Microsoft.ComponentDetection.Orchestrator/Commands/ScanCommand.cs
```diff
@@ -53,11 +53,12 @@ public override async Task<int> ExecuteAsync(
     /// Method to provide a way to execute the scan command and obtain the ScanResult object.
     /// </summary>
     /// <param name="settings">ScanSettings object specifying the parameters for the scan execution.</param>
+    /// <param name="cancellationToken">CancellationToken to monitor for cancellation requests.</param>
     /// <returns>A ScanResult object.</returns>
-    public async Task<ScanResult> ExecuteScanCommandAsync(ScanSettings settings)
+    public async Task<ScanResult> ExecuteScanCommandAsync(ScanSettings settings, CancellationToken cancellationToken = default)
```
ryanbrandenburg, line 58: I assume `cancellationToken` is optional because we haven't updated everywhere that uses this method yet?

src/Microsoft.ComponentDetection.Orchestrator/Services/DetectorProcessingService.cs
```diff
@@ -41,10 +41,12 @@ public DetectorProcessingService(
         this.logger = logger;
     }
 
+    /// <inheritdoc/>
     public async Task<DetectorProcessingResult> ProcessDetectorsAsync(
         ScanSettings settings,
         IEnumerable<IComponentDetector> detectors,
-        DetectorRestrictions detectorRestrictions)
+        DetectorRestrictions detectorRestrictions,
+        CancellationToken cancellationToken = default)
```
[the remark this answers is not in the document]
ryanbrandenburg: Copilot is right, seems this is unused so far?

src/Microsoft.ComponentDetection.Orchestrator/Commands/ScanCommand.cs
JamieMagee: Yes, this is an initial implementation. I plan to continue work on this, and plumb `CancellationToken` throughout the code.

PR #1547, opened 2025-11-19
Support packaging as a dotnet tool

JamieMagee:
See https://learn.microsoft.com/en-us/dotnet/core/tools/global-tools-how-to-create


ryanbrandenburg (approved):

grvillic (changes requested):
This impacts the generated .nupkg package with `dotnet pack`. Do not merge until we confirm downstream implications with generated artifacts.

PR #1551, opened 2025-11-21
Create LinuxApplicationLayerDetector

JamieMagee:
This change:
- Removes `pip` and `npm` component detection capabilities from `LinuxContainerDetector`
- Creates `LinuxApplicationLayerDetector` which subclasses `LinuxContainerDetector` and adds support back for `pip` and `npm` component detection
- Creates `LinuxApplicationLayerExperiment` to compare `LinuxContainerDetector` and `LinuxApplicationLayerDetector`

Please note that I also reformatted `LinuxContainerDetector`, so you may want to review with whitespace changes disabled.

### DIFF
--- a/test/Microsoft.ComponentDetection.Detectors.Tests/LinuxContainerDetectorTests.cs
+++ b/test/Microsoft.ComponentDetection.Detectors.Tests/LinuxContainerDetectorTests.cs
@@ -43,18 +43,39 @@ public class LinuxContainerDetectorTests
     public LinuxContainerDetectorTests()
     {
         this.mockDockerService = new Mock<IDockerService>();
-        this.mockDockerService.Setup(service => service.CanRunLinuxContainersAsync(It.IsAny<CancellationToken>()))
+        this.mockDockerService.Setup(service =>
+                service.CanRunLinuxContainersAsync(It.IsAny<CancellationToken>())
+            )
             .ReturnsAsync(true);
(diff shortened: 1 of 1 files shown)


src/Microsoft.ComponentDetection.Orchestrator/Experiments/Configs/LinuxApplicationLayerExperiment.cs
```diff
@@ -0,0 +1,25 @@
...
+/// packages in addition to system packages from Linux containers.
+/// </summary>
+public class LinuxApplicationLayerExperiment : IExperimentConfiguration
+{
+    /// <inheritdoc />
+    public string Name => "LinuxApplicationLayer";
+
+    /// <inheritdoc />
+    public bool IsInControlGroup(IComponentDetector componentDetector) =>
+        componentDetector is LinuxContainerDetector and not LinuxApplicationLayerDetector;
```
grvillic: This filter works great to compare delta between new linux detector and old linux detector, we might want to check against the other native detectors too (e.g. NPM)

JamieMagee: I added `NpmComponentDetector`, `NpmLockfileDetectorBase`, and `PipReportComponentDetector` to the control group.

JamieMagee:
Snapshot test differences are expected as we're effectively removing npm and pip components from the Linux detector.

PR #1557, opened 2025-11-23
Fix Analyser Warnings - SA1101, SA1201, & SA1202

AdamPlaskitt:
Enables the following StyleCop.Analysers as warnings and fixes existing issues:
- SA1101 - Local calls should be prefixed with `this`
- SA1201 - Elements should be in the correct order
- SA1202 - Elements should be ordered by access

Related to #202


AdamPlaskitt:
Reduced the scope of SA1202 changes to within `src/Microsoft.ComponentDetection.Contracts`

PR #1570, opened 2025-11-25
Migrate vcpkg and Go detectors from `Newtonsoft.Json` to `System.TextJson`

JamieMagee:
- `VcpkgComponentDetector`: Replace `JsonConvert.DeserializeObject` with `JsonSerializer.Deserialize`
- `GoCLIParser`: Replace `JsonTextReader`/`JsonSerializer` with `Utf8JsonReader` for parsing multi-object JSON output from `go list`
  - .NET 9 added `AllowMultipleValues`, but unfortunately we are still on .NET 8
- Add explicit `[JsonPropertyName]` attributes to vcpkg contract classes using SPDX 2.2.1 schema field names

Part of #231

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/go/Parsers/GoCLIParser.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/go/Parsers/GoCLIParser.cs
@@ -96,12 +96,6 @@ public class GoCLIParser : IGoParser
                 // Move past the consumed bytes plus any whitespace
                 var consumed = (int)reader.BytesConsumed;
                 remaining = remaining[consumed..];
-
-                // Skip whitespace between JSON objects
-                while (!remaining.IsEmpty && char.IsWhiteSpace((char)remaining[0]))
-                {
-                    remaining = remaining[1..];
-                }
             }
             catch (JsonException ex)
             {


src/Microsoft.ComponentDetection.Detectors/go/Parsers/GoCLIParser.cs
```diff
@@ -74,19 +74,37 @@ public async Task<bool> ParseAsync(
...
         using var record = new GoReplaceTelemetryRecord();
 
-        while (reader.Read())
+        // Go CLI outputs multiple JSON objects (not an array), so we need to parse them one by one
+        var remaining = goListOutput.AsMemory();
+        while (!remaining.IsEmpty)
         {
-            var serializer = new JsonSerializer();
-            var buildModule = serializer.Deserialize<GoBuildModule>(reader);
+            var reader = new Utf8JsonReader(System.Text.Encoding.UTF8.GetBytes(remaining.ToString()));
```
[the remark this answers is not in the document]
ryanbrandenburg: I think I agree with Copilot that I expected this outside of the while loop. Does it being inside have a particular meaning/effect? Maybe leave a comment about that if so.

I think if you move `reader` outside the loop you might also change the loop to pay attention to `reader` instead of `remaining`.

src/Microsoft.ComponentDetection.Detectors/go/Parsers/GoCLIParser.cs
```diff
@@ -74,19 +74,37 @@ public async Task<bool> ParseAsync(
...
+                    goBuildModules.Add(buildModule);
+                }
+
+                // Move past the consumed bytes plus any whitespace
+                var consumed = (int)reader.BytesConsumed;
+                remaining = remaining[consumed..];
 
-            goBuildModules.Add(buildModule);
+                // Skip whitespace between JSON objects
+                while (!remaining.IsEmpty && char.IsWhiteSpace(remaining.Span[0]))
```
ryanbrandenburg: Does STJ error if you have starting whitespace? I would have thought that it would silently ignore it.

src/Microsoft.ComponentDetection.Detectors/vcpkg/Contracts/Annotation.cs
```diff
@@ -2,14 +2,19 @@
 namespace Microsoft.ComponentDetection.Detectors.Vcpkg.Contracts;
 
 using System;
+using System.Text.Json.Serialization;
 
 public class Annotation
 {
+    [JsonPropertyName("annotationDate")]
```
ryanbrandenburg, line 9: I'm wondering why this didn't have annotations before?

src/Microsoft.ComponentDetection.Detectors/go/Parsers/GoCLIParser.cs
```diff
@@ -74,19 +74,37 @@ public async Task<bool> ParseAsync(
     private void RecordBuildDependencies(string goListOutput, ISingleFileComponentRecorder singleFileComponentRecorder, string projectRootDirectoryFullName)
```
ryanbrandenburg, line 76: If we didn't have explicit tests of this class before I think now might be a good time to add a couple.

src/Microsoft.ComponentDetection.Detectors/vcpkg/Contracts/Package.cs
```diff
@@ -1,21 +1,31 @@
 #nullable disable
 namespace Microsoft.ComponentDetection.Detectors.Vcpkg.Contracts;
 
+using System.Text.Json.Serialization;
+
 public class Package
 {
+    [JsonPropertyName("SPDXID")]
```
ryanbrandenburg, line 8: Just wanted to call out that this property doesn't seem to follow the same casing as the rest of it.

src/Microsoft.ComponentDetection.Detectors/vcpkg/Contracts/Annotation.cs
JamieMagee: I believe we relied on the fact that Newtonsoft will deserialize case insensitively by default, whereas `System.Text.Json` is a lot more strict.

src/Microsoft.ComponentDetection.Detectors/go/Parsers/GoCLIParser.cs
JamieMagee: It doesn't seem to mind spaces, but it doesn't like newlines.

src/Microsoft.ComponentDetection.Detectors/go/Parsers/GoCLIParser.cs
JamieMagee: Good idea. I'll add some now.

src/Microsoft.ComponentDetection.Detectors/go/Parsers/GoCLIParser.cs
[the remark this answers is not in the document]
JamieMagee: Yep, you're right. I implemented this, got it working, and then never went back to optimize this correctly. There's no specific reason why it shouldn't be created once, outside the loop.

src/Microsoft.ComponentDetection.Detectors/go/Parsers/GoCLIParser.cs
```diff
@@ -74,19 +74,37 @@ public async Task<bool> ParseAsync(
...
+            {
+                var buildModule = JsonSerializer.Deserialize<GoBuildModule>(ref reader);
+                if (buildModule != null)
+                {
+                    goBuildModules.Add(buildModule);
+                }
+
+                // Move past the consumed bytes plus any whitespace
+                var consumed = (int)reader.BytesConsumed;
+                remaining = remaining[consumed..];
```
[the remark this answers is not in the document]
JamieMagee: I don't believe that go allows module names to contain multi-byte unicode characters. Or even if it does, registries (i.e. GitHub, GitLab) don't allow them.

src/Microsoft.ComponentDetection.Detectors/vcpkg/Contracts/Package.cs
JamieMagee: Thanks for noticing, but in this instance it's on purpose. See https://raw.githubusercontent.com/spdx/spdx-spec/v2.2.1/schemas/spdx-schema.json

```json
"properties": {
  "SPDXID": {
    "type": "string",
    "description": "Uniquely identify any element in an SPDX document which may be referenced by other elements."
  }
}
```

src/Microsoft.ComponentDetection.Detectors/go/Parsers/GoCLIParser.cs
JamieMagee: Done!

test/Microsoft.ComponentDetection.Detectors.Tests/GoCLIParserTests.cs
```diff
@@ -0,0 +1,463 @@
...
+    /// <summary>
+    /// Creates a mock IComponentStream with the specified location.
+    /// </summary>
+    private static IComponentStream CreateComponentStream(string location)
+    {
+        var mock = new Mock<IComponentStream>();
+        mock.Setup(s => s.Location).Returns(location);
+        mock.Setup(s => s.Stream).Returns(new MemoryStream(Encoding.UTF8.GetBytes("module test")));
+        return mock.Object;
+    }
```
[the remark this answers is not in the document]
JamieMagee: I just removed the stream entirely, as `GoCLIParser` doesn't actually use it. It only uses `Location`

src/Microsoft.ComponentDetection.Detectors/go/Parsers/GoCLIParser.cs
```diff
@@ -66,27 +67,47 @@ public async Task<bool> ParseAsync(
...
 
-            goBuildModules.Add(buildModule);
+                // Skip whitespace between JSON objects
+                while (!remaining.IsEmpty && char.IsWhiteSpace((char)remaining[0]))
```
ryanbrandenburg: I ran a fiddle and am fairly confident that STJ will handle the whitespace without this `while` loop.

JamieMagee: Excellent! I just removed that.

PR #1592, opened 2025-12-09
Downgrade 9.x packages

sebasgomez238:
A PR originally created by renovate: https://github.com/microsoft/component-detection/pull/1511 bumped a set of packages from 8.x to version 9.0.10.

The dotnet-sdk bump was then removed from the PR and merged.

This is leading to downstream incompatibility for projects targeting net8.0 trying to use the version required by the component-detection libraries.

Similar issue can be seen here: https://github.com/dotnet/runtime/issues/112634

I'm proposing to revert these changes until component-detection targets net9.0 or higher


Directory.Packages.props
```diff
@@ -47,7 +47,7 @@
     <PackageVersion Include="System.Reactive" Version="6.1.0" />
     <PackageVersion Include="System.Runtime.Loader" Version="4.3.0" />
     <PackageVersion Include="System.Text.Json" Version="9.0.10" />
```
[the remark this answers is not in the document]
sebasgomez238: I only reverted the packages necessary to get back to a working state.

grvillic (approved):

Directory.Packages.props
grvillic, line 56: Should we revert Asn1 to `6.0.1`? The major update wasn't necessary here either.

sebasgomez238: This one isn't causing any issues so I think its okay to leave as is.

PR #1595, opened 2025-12-11
LinuxApplicationLayerExperiment adjustments.

sebasgomez238:
Add NuGet to control group of the experiments. Support was already added here: https://github.com/microsoft/component-detection/pull/1548

Also changes logic to only report telemetry if something was found in the experiment group.

### DIFF
--- a/src/Microsoft.ComponentDetection.Orchestrator/Experiments/Configs/LinuxApplicationLayerExperiment.cs
+++ b/src/Microsoft.ComponentDetection.Orchestrator/Experiments/Configs/LinuxApplicationLayerExperiment.cs
@@ -25,7 +25,8 @@ public class LinuxApplicationLayerExperiment : IExperimentConfiguration
                 or NpmLockfileDetectorBase
                 or PipReportComponentDetector
                 or NuGetComponentDetector
-                or NuGetProjectModelProjectCentricComponentDetector;
+                or NuGetProjectModelProjectCentricComponentDetector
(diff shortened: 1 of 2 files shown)


src/Microsoft.ComponentDetection.Orchestrator/Experiments/Configs/LinuxApplicationLayerExperiment.cs
```diff
@@ -22,12 +23,26 @@ public bool IsInControlGroup(IComponentDetector componentDetector) =>
             is (LinuxContainerDetector and not LinuxApplicationLayerDetector)
                 or NpmComponentDetector
                 or NpmLockfileDetectorBase
-                or PipReportComponentDetector;
+                or PipReportComponentDetector
+                or NuGetComponentDetector
```
sebasgomez238, line 27: Support for nuget components in containers was added here: https://github.com/microsoft/component-detection/pull/1548/changes#diff-24e5311d64467a9ff6699678ae88979b16c70b7932cb35ed751ad85e4f62a686 

Is it reasonable to add all 3 of these nuget detectors to the control group?

grvillic: Yes, given that we are reporting different ecosystems in new Linux detector the experiments give better signals when the control group includes those detectors.

PR #1599, opened 2025-12-17
Introduce C++ SDK component type

grvillic:
## Context
We are experimenting with a new component type for C++ SDK version detection.


src/Microsoft.ComponentDetection.Contracts/TypedComponent/CppSdkComponent.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    [JsonIgnore]
+    public override ComponentType Type => ComponentType.CppSdk;
+
+    [JsonPropertyName("packageUrl")]
+    public override PackageURL PackageUrl
+    {
+        get
+        {
+            var qualifiers = new SortedDictionary<string, string>
```
grvillic, line 46: Purl spec for cpp does not exist, hence we are using generic for now with specific qualifier, see doc: https://github.com/package-url/purl-spec/blob/main/types-doc/generic-definition.md

PR #1603, opened 2026-01-07
Reduce log level for per file level logs in RustSbomDetector

AMaini503:
Reduce log level for per file level logs in RustSbomDetector

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/rust/RustSbomDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/rust/RustSbomDetector.cs
@@ -252,8 +252,6 @@ public class RustSbomDetector : FileComponentDetector
         var normDirectory = this.pathUtilityService.NormalizePath(directory);
         var fileName = Path.GetFileName(location);
 
-        this.Logger.LogDebug("Processing file: {Location}", normLocation);
-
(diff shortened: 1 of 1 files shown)


src/Microsoft.ComponentDetection.Detectors/rust/RustSbomDetector.cs
```diff
@@ -252,7 +252,7 @@ protected override async Task OnFileFoundAsync(
         var normDirectory = this.pathUtilityService.NormalizePath(directory);
         var fileName = Path.GetFileName(location);
 
-        this.Logger.LogInformation("Processing file: {Location}", normLocation);
+        this.Logger.LogDebug("Processing file: {Location}", normLocation);
```
grvillic: I am not sure if we should keep this log, we have a lot of files scanned in an agent, we don't need to be as explicit when a file is picked up, no other detector has it, and unless directory is ignored or something we won't print this. 

On the other hand, if we were skipping a directory or a file (like below) that makes sense to log as debug.

AMaini503: Removed.

grvillic (approved):

PR #1616, opened 2026-01-15
Update LinuxApplicationLayerExperiment control group

JamieMagee:
Updates the `LinuxApplicationLayerExperiment` to correctly measure net-new application packages found exclusively inside containers (e.g., packages installed via RUN npm add lodash).

### Experiment Configuration:
  - Control Group: File-based detectors + LinuxContainerDetector (system packages)
  - Experiment Group: File-based detectors + LinuxApplicationLayerDetector (system + application packages)

This setup ensures:
- Components from manifest files (e.g., package.json) cancel out in both groups
- System packages detected by LinuxContainerDetector cancel out with those from LinuxApplicationLayerDetector
- The diff shows only application packages installed inside containers that weren't in manifest files

### ShouldRecord Logic:
- Returns false when LinuxContainerDetector finds 0 components (no containers scanned)
- Returns true for file-based detectors to avoid prematurely removing the experiment before container detectors run


grvillic:
I don't think we should update the experiment. This detector is introducing packages from multiple ecosystems, we need to know exactly what it found in all of them, that the other detectors couldn't report or vice versa.

PR #1620, opened 2026-02-03
Improve serialization compatibility with new component types

grvillic:
## Context
We have seen downstream customers importing both SBOM and Component Detection nupkgs have issues with deserialization when a new component type is added here that they do not handle yet, causing the entire scan to fail.

## Solution
When "ComponentType" is unknown, return a null component and let downstream clients handle that gracefully so their scan continues succeeding with preexisting component types and they aren't forced to introduce handling of every new type in their codebase.

Since we needed a new serialization converter for system.text.json file, we created a utility class to share component types between both newtonsoft and system.text.json when handling the component type.

### DIFF
--- a/src/Microsoft.ComponentDetection.Contracts/BcdeModels/TypedComponentMapping.cs
+++ b/src/Microsoft.ComponentDetection.Contracts/BcdeModels/TypedComponentMapping.cs
@@ -46,7 +46,7 @@ internal static class TypedComponentMapping
     /// <returns>True if the type discriminator was recognized; otherwise false.</returns>
     public static bool TryGetType(string typeDiscriminator, out Type targetType)
     {
-        if (string.IsNullOrEmpty(typeDiscriminator))
+        if (string.IsNullOrWhiteSpace(typeDiscriminator))
         {
             targetType = null;
             return false;
--- a/test/Microsoft.ComponentDetection.Contracts.Tests/TypedComponentSerializationTests.cs
+++ b/test/Microsoft.ComponentDetection.Contracts.Tests/TypedComponentSerializationTests.cs
@@ -319,4 +319,31 @@ public class TypedComponentSerializationTests
             typeProperties.Should().Be(1, $"The 'type' property should appear exactly once at the root level for {component.Type}");
         }
     }
+
+    [TestMethod]
+    public void TypedComponentMapping_TryGetType_NullDiscriminator_ReturnsFalseAndNull()
+    {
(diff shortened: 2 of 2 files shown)


src/Microsoft.ComponentDetection.Contracts/BcdeModels/TypedComponentConverter.cs
```diff
@@ -46,8 +23,14 @@ public override object ReadJson(
     {
         var jo = JToken.Load(reader);
 
-        var value = (ComponentType)Enum.Parse(typeof(ComponentType), (string)jo["type"]);
-        var targetType = this.componentTypesToTypes[value];
+        var typeString = (string)jo["type"];
+        if (!TypedComponentMapping.TryGetType(typeString, out var targetType))
```
grvillic, line 27: This is the safety check to ensure it can be deserialized safely as null when downstream client doesn't support it yet.

src/Microsoft.ComponentDetection.Contracts/BcdeModels/TypedComponentSystemTextJsonConverter.cs
```diff
@@ -0,0 +1,68 @@
...
+        // Extract the type discriminator
+        // Check both camelCase and PascalCase property names
+        if (!root.TryGetProperty(TypePropertyNameCamelCase, out var typeProperty) &&
+            !root.TryGetProperty(TypePropertyNamePascalCase, out typeProperty))
+        {
+            return null;
+        }
+
+        var typeDiscriminator = typeProperty.GetString();
+        if (!TypedComponentMapping.TryGetType(typeDiscriminator, out var targetType))
```
grvillic, line 40: Safety check is here, the rest gets deserialized using default System.Text libraries.

src/Microsoft.ComponentDetection.Orchestrator/Services/DetectorProcessingService.cs
```diff
@@ -384,7 +387,7 @@ private void LogTabularOutput(ConcurrentDictionary<string, DetectorRunResult> pr
             providerElapsedTime.Sum(x => x.Value.ComponentsFoundCount).ToString(),
             providerElapsedTime.Sum(x => x.Value.ExplicitlyReferencedComponentCount).ToString());
 
-        AnsiConsole.Write(table);
+        this.console.Write(table);
```
grvillic, line 390: While I was running my tests I found this to be disposed and sporadically breaking unit tests, this ensures it consistently succeeds since `AnsiConsole` is a static class

PR #1627, opened 2026-02-03
Mask feed credentials for CLI commands + remove superfluous risky logs

grvillic:
## Context
With custom PyPi setups, if an error is returned by CLI command we may end up logging secrets unsuspectingly.

## Solution
We only need the log at the start the requested CLI command with masked data, the actual exception can be tracked just with the status code. Improve String utilities to be non-greedy and also mask when multiple URLs are provided.

### DIFF
--- a/src/Microsoft.ComponentDetection.Common/Utilities/StringUtilities.cs
+++ b/src/Microsoft.ComponentDetection.Common/Utilities/StringUtilities.cs
@@ -6,8 +6,8 @@ using System.Text.RegularExpressions;
 
 public static class StringUtilities
 {
-    // Matches credentials in URLs like https://user:password@host
-    private static readonly Regex UrlCredentialsRegex = new(@"(?<=https://)(.+?)(?=@)", RegexOptions.Compiled | RegexOptions.IgnoreCase, TimeSpan.FromSeconds(5));
+    // Matches credentials in URLs like http://user:password@host or https://user:password@host
+    private static readonly Regex UrlCredentialsRegex = new(@"(?<=https?://)(.+?)(?=@)", RegexOptions.Compiled | RegexOptions.IgnoreCase, TimeSpan.FromSeconds(5));
 
     // Matches SAS tokens in query strings (e.g., ?sv=...&sig=... or ?se=...&sig=...)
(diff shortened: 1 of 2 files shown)


src/Microsoft.ComponentDetection.Common/Utilities/StringUtilities.cs
```diff
@@ -6,7 +6,7 @@ namespace Microsoft.ComponentDetection.Common;
 
 public static class StringUtilities
 {
-    private static readonly Regex SensitiveInfoRegex = new Regex(@"(?<=https://)(.+)(?=@)", RegexOptions.Compiled | RegexOptions.IgnoreCase, TimeSpan.FromSeconds(5));
+    private static readonly Regex SensitiveInfoRegex = new Regex(@"(?<=https://)(.+?)(?=@)", RegexOptions.Compiled | RegexOptions.IgnoreCase, TimeSpan.FromSeconds(5));
```
grvillic: Currently we were only masking the first URLs, if a string was composed by multiple index urls (rare), we could have missed the second one. Unit test was added to prevent regressions.

src/Microsoft.ComponentDetection.Detectors/pip/PipCommandService.cs
```diff
@@ -192,11 +192,10 @@ private async Task<CommandLineExecutionResult> ExecuteCommandAsync(
 
             if (command.ExitCode == -1 && cancellationToken.IsCancellationRequested)
             {
-                var errorMessage = $"PipReport: Cancelled for file '{formattedPath}' with command '{pipReportCommand.RemoveSensitiveInformation()}'.";
+                var errorMessage = $"PipReport: Cancelled for file '{formattedPath}'.";
                 using var failureRecord = new PipReportFailureTelemetryRecord
                 {
                     ExitCode = command.ExitCode,
-                    StdErr = $"{errorMessage} {command.StdErr}",
```
grvillic, line 199: I realized CLI error can contain other unexpected secrets outside of the provided CLI input that may contain secrets, like a SAS token. We need to play it safe here and only log the exit code.

PR #1628, opened 2026-02-03
Maven combined detector experiment

zhenghao104:
# MavenWithFallbackDetector

## Overview

The `MavenWithFallbackDetector` is an experimental Maven detector that combines **Maven CLI detection** with **static pom.xml parsing fallback**. It provides resilient Maven dependency detection even when the Maven CLI fails (e.g., due to authentication errors with private repositories).

## Key Features

- **Primary detection**: Uses Maven CLI (`mvn dependency:tree`) for full transitive dependency resolution
- **Automatic fallback**: Falls back to static pom.xml parsing when Maven CLI fails
- **Partial failure handling**: Can use Maven CLI results for successful projects while falling back to static parsing for failed ones
- **Authentication error detection**: Identifies and reports authentication failures with guidance
- **Nested pom.xml filtering**: Optimizes multi-module projects by only processing root pom.xml files with Maven CLI
- **Telemetry**: Reports detection method, fallback reasons, and component counts
- **Timeout protection**: 5-minute internal timeout as a safety guardrail for the prepare detection phase

## Detection Flow

```
┌─────────────────────────────────────────────────────────────────────────┐
│                        OnPrepareDetectionAsync                          │
│              (wrapped with 5-minute internal timeout)                   │
└─────────────────────────────────────────────────────────────────────────┘
                                    │
                                    ▼
                    ┌───────────────────────────────┐
                    │ Check CD_MAVEN_DISABLE_CLI    │
                    │ environment variable          │
                    └───────────────────────────────┘
                                    │
                    ┌───────────────┴───────────────┐
                    │                               │
                    ▼                               ▼
            [Disabled=true]                 [false/unset]
                    │                               │
                    ▼                               ▼
        ┌───────────────────┐           ┌───────────────────────┐
        │ Use Static Parser │           │ Check Maven CLI exists │
        │ for ALL pom.xml   │           └───────────────────────┘
        └───────────────────┘                       │
                                    ┌───────────────┴───────────────┐
                                    │                               │
                                    ▼                               ▼
                            [CLI not found]                 [CLI available]
                                    │                               │
                                    ▼                               ▼
                        ┌───────────────────┐       ┌───────────────────────────┐
                        │ Use Static Parser │       │ RemoveNestedPomXmls       │
                        │ for ALL pom.xml   │       │ (filter to root POMs only)│
                        └───────────────────┘       └───────────────────────────┘
                                                                │
                                                                ▼
                                                ┌───────────────────────────────┐
                                                │ Run mvn dependency:tree       │
                                                │ for each root pom.xml         │
                                                │ (sequentially via ActionBlock)│
                                                └───────────────────────────────┘
                                                                │
                                                                ▼
                                                ┌───────────────────────────────┐
                                                │ Read bcde.mvndeps files       │
                                                │ immediately after generation  │
                                                └───────────────────────────────┘
                                                                │
                                                                ▼
                                                ┌───────────────────────────────┐
                                                │ Check results per directory:  │
                                                │ - Success: deps file present  │
                                                │ - Failure: deps file missing  │
                                                └───────────────────────────────┘
                                                                │
                                ┌───────────────────────┬───────┴───────────────┐
                                │                       │                       │
                                ▼                       ▼                       ▼
                        [All succeeded]         [All failed]            [Partial]
                                │                       │                       │
                                ▼                       ▼                       ▼
                    ┌─────────────────┐     ┌─────────────────────┐ ┌─────────────────────┐
                    │ Return          │     │ AnalyzeMvnCliFailure│ │ AnalyzeMvnCliFailure│
                    │ bcde.mvndeps    │     │ Return ALL pom.xml  │ │ Return bcde.mvndeps │
                    │ only            │     │ in failed dirs for  │ │ + pom.xml in failed │
                    │                 │     │ static parsing      │ │ dirs for static     │
                    │ Method: MvnCli  │     │ Method: StaticOnly  │ │ Method: Mixed       │
                    └─────────────────┘     └─────────────────────┘ └─────────────────────┘
```

### 1. Environment Variable Control

```csharp
internal const string DisableMvnCliEnvVar = "CD_MAVEN_DISABLE_CLI";
```

**CD_MAVEN_DISABLE_CLI**:
- Set to `true` to bypass Maven CLI entirely and use only static parsing

Useful when:
- Maven CLI is known to fail in the environment
- Faster scans are preferred (at the cost of transitive dependencies)
- Debugging static parser behavior

### 2. Internal Timeout Protection

The detector has an internal 5-minute timeout (`PrepareDetectionTimeout`) as a safety guardrail:

```csharp
private static readonly TimeSpan PrepareDetectionTimeout = TimeSpan.FromMinutes(5);
```

**Why?** The `OnPrepareDetectionAsync` phase doesn't have the same guardrails as `OnFileFoundAsync`. This timeout prevents hangs in the experimental detector. If the timeout is reached:
1. The operation is cancelled
2. Telemetry is recorded (`TimedOut = true`)
3. Falls back to static pom.xml parsing

### 3. Cancellation Token Propagation

The detector properly propagates cancellation tokens through the entire call chain, combining the framework's cancellation token with the internal timeout:

```
┌───────────────────────────────────────────────────────────────────────────┐
│ Orchestrator (4-min experimental detector timeout CancellationToken)      │
└─────────────────────────────┬─────────────────────────────────────────────┘
                              │
                              ▼
┌───────────────────────────────────────────────────────────────────────────┐
│ MavenWithFallbackDetector.OnPrepareDetectionAsync                         │
│   └── CreateLinkedTokenSource(cancellationToken, 5-min internal timeout)  │
│       (Combines framework token with internal timeout - first wins)       │
└─────────────────────────────┬─────────────────────────────────────────────┘
                              │
                              ▼
┌───────────────────────────────────────────────────────────────────────────┐
│ MavenCommandService.GenerateDependenciesFileAsync()                       │
│   └── CancellationTokenSource.CreateLinkedTokenSource(cancellationToken)  │
│       (Links with optional file-level timeout from env var)               │
└─────────────────────────────┬─────────────────────────────────────────────┘
                              │
                              ▼
┌───────────────────────────────────────────────────────────────────────────┐
│ CommandLineInvocationService.ExecuteCommandAsync(cancellationToken)       │
│   └── cancellationToken.Register(() => process.Kill())                    │
│       (Kills Maven process if any token is cancelled)                     │
└───────────────────────────────────────────────────────────────────────────┘
```

**Key points:**
- The 4-minute detector timeout from the orchestrator is combined with the 5-minute internal timeout
- Whichever triggers first will cancel the operation
- If cancelled, the Maven process is **killed immediately** via `Process.Kill()`
- Additional file-level timeouts can be configured via `MvnCLIFileLevelTimeoutSeconds` environment variable

### 4. Sequential Maven CLI Execution

Maven CLI invocations are processed sequentially using `ActionBlock` with default settings:

```csharp
// Process pom.xml files sequentially to match MvnCliComponentDetector behavior.
// Sequential execution avoids Maven local repository lock contention and
// reduces memory pressure from concurrent Maven JVM processes.
var processPomFile = new ActionBlock<ProcessRequest>(async processRequest => { ... });
```

**Why sequential execution?** 
- Avoids Maven local repository lock contention when multiple `mvn` processes access `~/.m2/repository` simultaneously
- Reduces memory pressure from concurrent Maven JVM processes (each can consume 256MB-1GB+)
- Matches the behavior of `MvnCliComponentDetector` for consistent performance

### 5. RemoveNestedPomXmls

Filters pom.xml files to keep only **root-level** ones for Maven CLI processing.

**Why?** In multi-module Maven projects:
```
/project/
  pom.xml              ← ROOT (parent pom) - KEEP
  /module-a/
    pom.xml            ← NESTED (child) - SKIP
  /module-b/
    pom.xml            ← NESTED (child) - SKIP
```

Running `mvn dependency:tree` on the root pom automatically processes all child modules. Processing each nested pom separately would be redundant and slow.

**Algorithm:**
1. Collect all pom.xml files and their directories
2. Sort directories by path length (shortest first)
3. For each pom.xml, check if any shorter path is a parent directory
4. If a parent directory contains a pom.xml → skip this file (it's nested)
5. If no parent pom.xml found → keep this file (it's a root)

### 6. Maven CLI Execution

Uses `IMavenCommandService.GenerateDependenciesFileAsync()` to run:
```bash
mvn dependency:tree -B -DoutputFile=bcde.mvndeps -DoutputType=text -f<pom.xml>
```

This generates a `bcde.mvndeps` file containing the full dependency tree.

**Immediate file reading:** Unlike `MvnCliComponentDetector` which scans for deps files after all CLI executions, this detector reads the deps file content immediately after each successful CLI execution. This enables per-file success/failure tracking.

### 7. Failure Detection

After each Maven CLI execution, the detector checks the result:
- **CLI succeeded AND deps file exists with content** → Success, add deps file to results
- **CLI succeeded BUT deps file missing/empty** → Treat as failure, log warning
- **CLI failed** → Failure, capture error output for analysis

### 8. Static Parsing Fallback with Directory Optimization

When Maven CLI fails for a directory, the detector scans for all pom.xml files in that directory and its subdirectories:

```csharp
private IEnumerable<ProcessRequest> GetAllPomFilesInDirectories(HashSet<string> directories)
{
    // Normalize directories once for efficient lookup
    var normalizedDirs = directories
        .Select(d => d.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar) + Path.DirectorySeparatorChar)
        .ToList();

    return this.ComponentStreamEnumerableFactory
        .GetComponentStreams(...)
        .Where(componentStream =>
        {
            var normalizedFileDir = fileDir.TrimEnd(...) + Path.DirectorySeparatorChar;
            // Pre-normalized comparison for efficiency
            return normalizedDirs.Any(fd =>
                normalizedFileDir.Equals(fd, StringComparison.OrdinalIgnoreCase) ||
                normalizedFileDir.StartsWith(fd, StringComparison.OrdinalIgnoreCase));
        })
        ...
}
```

**Performance optimization:** Directory paths are normalized once upfront, then reused for all file comparisons, avoiding redundant string operations.

### 9. Static Parsing Limitations

When falling back to static parsing, dependencies are extracted directly from pom.xml:

```xml
<dependencies>
    <dependency>
        <groupId>org.apache.commons</groupId>
        <artifactId>commons-lang3</artifactId>
        <version>3.12.0</version>
    </dependency>
</dependencies>
```

**Limitations of static parsing:**
- Only detects **direct dependencies** (no transitive)
- Cannot resolve version ranges (e.g., `[1.0,2.0)`)
- May not resolve all property references (e.g., `${project.version}`)

### 10. Authentication Error Analysis

When Maven CLI fails, the detector analyzes error patterns:

```csharp
private static readonly string[] AuthErrorPatterns =
[
    "401",
    "403",
    "Unauthorized",
    "Access denied",
];
```

If authentication errors are detected:
1. Sets `fallbackReason = MavenFallbackReason.AuthenticationFailure`
2. Extracts failed endpoint URLs from error messages using regex
3. Logs guidance for resolving authentication issues
4. Records failed endpoints in telemetry (up to 10)

## Detection Methods

| Method | When Used | What's Detected |
|--------|-----------|-----------------|
| `MvnCliOnly` | Maven CLI succeeds for all pom.xml files | Full transitive dependencies |
| `StaticParserOnly` | Maven CLI disabled, unavailable, or fails completely | Direct dependencies only |
| `Mixed` | Maven CLI succeeds for some, fails for others | Transitive (where CLI worked) + Direct (where CLI failed) |

## Fallback Reasons

| Reason | Description |
|--------|-------------|
| `None` | No fallback needed (MvnCli succeeded) |
| `MvnCliDisabledByUser` | User set `CD_MAVEN_DISABLE_CLI=true` |
| `MavenCliNotAvailable` | `mvn` command not found in PATH |
| `AuthenticationFailure` | Maven CLI failed with 401/403 errors |
| `OtherMvnCliFailure` | Maven CLI failed for other reasons (including timeout) |

## Telemetry

The detector records the following telemetry:

| Key | Description |
|-----|-------------|
| `DetectionMethod` | `MvnCliOnly`, `StaticParserOnly`, or `Mixed` |
| `FallbackReason` | Why fallback occurred (if applicable) |
| `MvnCliComponentCount` | Components detected via Maven CLI |
| `StaticParserComponentCount` | Components detected via static parsing |
| `TotalComponentCount` | Total components detected |
| `MavenCliAvailable` | Whether Maven CLI was found |
| `OriginalPomFileCount` | Number of root pom.xml files processed |
| `FailedEndpoints` | URLs of repositories that had auth failures (up to 10) |
| `TimedOut` | `true` if internal 5-minute timeout was reached |
| `PrepareDetectionError` | Exception type if unexpected error occurred |

## Version Property Resolution

Static parsing supports resolving version properties:

```xml
<properties>
    <commons.version>3.12.0</commons.version>
</properties>
<dependencies>
    <dependency>
        <groupId>org.apache.commons</groupId>
        <artifactId>commons-lang3</artifactId>
        <version>${commons.version}</version>  <!-- Resolved to 3.12.0 -->
    </dependency>
</dependencies>
```

The resolver:
1. Checks `<properties>` section in current document
2. Checks other loaded pom.xml documents (for multi-module projects)
3. If unresolved, skips the dependency with a debug log

### As Part of Component Detection

The detector is registered as an experimental detector (`IExperimentalDetector`) and runs automatically when Component Detection scans a directory containing `pom.xml` files.

### Parallel Execution with MvnCliComponentDetector

> **Note:** This detector uses the **same** output filename (`bcde.mvndeps`) and Maven local repository as `MvnCliComponentDetector`. If both detectors run in parallel on the same pom.xml files, they may interfere with each other. The detector is intended to **replace** `MvnCliComponentDetector`, not run alongside it.

### Disabling Maven CLI (Static Parsing Only)

Set the environment variable to use only static parsing:

```bash
# Bash
export CD_MAVEN_DISABLE_CLI=true

# PowerShell
$env:CD_MAVEN_DISABLE_CLI = "true"

# Azure DevOps Pipeline
variables:
  CD_MAVEN_DISABLE_CLI: 'true'
```

## Experiment Configuration

The detector is paired with `MavenWithFallbackExperiment` which compares results against the standard `MvnCliComponentDetector`:

- **Control group**: `MvnCliComponentDetector`
- **Experiment group**: `MavenWithFallbackDetector`

Enable experiments via:
```bash
export CD_DETECTOR_EXPERIMENTS=true
```

---

### Comparison with Existing Detectors

| Aspect | MavenComponentDetector | MvnCliComponentDetector | MavenWithFallbackDetector |
|--------|------------------------|-------------------------|---------------------------|
| **Approach** | Static XML parsing | Maven CLI only | CLI + static fallback |
| **Speed** | ⚡ Fastest | 🐌 Slowest | 🔄 Variable |
| **Accuracy** | Direct deps only | Full transitive | Best effort |
| **Requirements** | None | Maven CLI + network | None (graceful degradation) |
| **Failure handling** | N/A | No fallback | Automatic fallback |
| **Parallelism** | Per-file | Sequential | Sequential |

### Performance Optimizations

1. **Sequential CLI execution**: Uses `ActionBlock` with default settings (sequential) to avoid Maven repository lock contention
2. **Directory normalization**: Pre-normalizes directory paths once for efficient lookup during fallback scanning
3. **Immediate file reading**: Reads deps file content immediately after CLI success, avoiding a second directory scan

---

## Unit Tests

The following unit tests are implemented in `MavenWithFallbackDetectorTests.cs`:

### Maven CLI Scenarios

| Test | Description |
|------|-------------|
| `WhenMavenCliNotAvailable_FallsBackToStaticParsing_Async` | Verifies fallback to static parsing when Maven CLI is not installed |
| `WhenMavenCliNotAvailable_DetectsMultipleDependencies_Async` | Verifies static parsing detects multiple dependencies correctly |
| `WhenMavenCliSucceeds_UsesMvnCliResults_Async` | Verifies Maven CLI results are used when available |
| `WhenMavenCliSucceeds_PreservesTransitiveDependencies_Async` | Verifies transitive dependency relationships are preserved |
| `WhenMavenCliProducesNoOutput_FallsBackToStaticParsing_Async` | Verifies fallback when Maven CLI runs but produces no output |

### Static Parser Edge Cases

| Test | Description |
|------|-------------|
| `StaticParser_IgnoresDependenciesWithoutVersion_Async` | Verifies dependencies without version tags are skipped |
| `StaticParser_IgnoresDependenciesWithVersionRanges_Async` | Verifies version ranges (e.g., `[1.0,2.0)`) are skipped |
| `StaticParser_ResolvesPropertyVersions_Async` | Verifies `${property}` versions are resolved from `<properties>` |
| `StaticParser_IgnoresDependenciesWithUnresolvablePropertyVersions_Async` | Verifies unresolvable properties are skipped |

### Empty/Missing File Scenarios

| Test | Description |
|------|-------------|
| `WhenNoPomXmlFiles_ReturnsSuccessWithNoComponents_Async` | Verifies graceful handling of no pom.xml files |
| `WhenPomXmlHasNoDependencies_ReturnsSuccessWithNoComponents_Async` | Verifies handling of pom.xml with no dependencies |

### Environment Variable Control

| Test | Description |
|------|-------------|
| `WhenDisableMvnCliTrue_UsesStaticParsing_Async` | Verifies `CD_MAVEN_DISABLE_CLI=true` bypasses Maven CLI entirely |
| `WhenDisableMvnCliEnvVarIsFalse_UsesMvnCliNormally_Async` | Verifies `CD_MAVEN_DISABLE_CLI=false` allows Maven CLI to run |
| `WhenDisableMvnCliEnvVarNotSet_UsesMvnCliNormally_Async` | Verifies Maven CLI runs when env var is not set |
| `WhenDisableMvnCliEnvVarSetToInvalidValue_UsesMvnCliNormally_Async` | Verifies Maven CLI runs when env var has invalid value |

### Nested Pom.xml Handling

| Test | Description |
|------|-------------|
| `WhenMvnCliSucceeds_NestedPomXmlsAreFilteredOut_Async` | Verifies nested pom.xml files are filtered when Maven CLI is used |
| `WhenMvnCliFailsCompletely_AllNestedPomXmlsAreRestoredForStaticParsing_Async` | Verifies all nested pom.xml files are restored when Maven CLI fails completely |
| `WhenMvnCliPartiallyFails_NestedPomXmlsRestoredOnlyForFailedDirectories_Async` | Verifies nested pom.xml files are restored only for failed directories |

### Error Analysis and Telemetry

| Test | Description |
|------|-------------|
| `WhenMvnCliFailsWithAuthError_LogsFailedEndpointAndSetsTelemetry_Async` | Verifies auth errors (401/403) are detected and telemetry is set |
| `WhenMvnCliFailsWithNonAuthError_SetsFallbackReasonToOther_Async` | Verifies non-auth errors set telemetry to `OtherMvnCliFailure` |

---

## File Structure

```
src/Microsoft.ComponentDetection.Detectors/maven/
├── MavenWithFallbackDetector.cs      # Main detector implementation
├── MavenWithFallbackDetector.md      # This documentation
├── MvnCliComponentDetector.cs        # Standard Maven CLI detector
├── MavenComponentDetector.cs         # Static pom.xml parser detector
├── MavenCommandService.cs            # Maven CLI execution service
├── IMavenCommandService.cs           # Service interface
└── ...

src/Microsoft.ComponentDetection.Orchestrator/Experiments/Configs/
└── MavenWithFallbackExperiment.cs    # Experiment configuration

test/Microsoft.ComponentDetection.Detectors.Tests/
└── MavenWithFallbackDetectorTests.cs # Unit tests
```

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
@@ -345,10 +345,10 @@ public class MavenWithFallbackDetector : FileComponentDetector, IExperimentalDet
                 }
             }
         },
-        new ExecutionDataflowBlockOptions
-        {
-            CancellationToken = cancellationToken,
-        });
+            new ExecutionDataflowBlockOptions
+            {
+                CancellationToken = cancellationToken,
+            });
 
         await this.RemoveNestedPomXmls(processRequests, cancellationToken).ForEachAsync(
             processRequest =>


src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,706 @@
...
+
+    private readonly IMavenCommandService mavenCommandService;
+    private readonly IEnvironmentVariableService envVarService;
+    private readonly Dictionary<string, XmlDocument> documentsLoaded = [];
+
+    // Track original pom.xml files for potential fallback
+    private readonly ConcurrentBag<ProcessRequest> originalPomFiles = [];
+
+    // Track Maven CLI errors for analysis
+    private readonly ConcurrentBag<string> mavenCliErrors = [];
```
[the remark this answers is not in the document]
zhenghao104: Requires changes to MavenCommandService, working on it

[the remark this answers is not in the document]
zhenghao104: Fixed

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,706 @@
...
+    }
+
+    private void LogAuthErrorGuidance()
+    {
+        var guidance = new StringBuilder();
+        guidance.AppendLine("Maven CLI failed with authentication errors.");
+
+        if (!this.failedEndpoints.IsEmpty)
+        {
+            guidance.AppendLine("2. The following Maven repository endpoints had authentication failures:");
```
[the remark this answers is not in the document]
zhenghao104: Fixed

src/Microsoft.ComponentDetection.Orchestrator/Experiments/Configs/MavenWithFallbackExperiment.cs
```diff
@@ -0,0 +1,24 @@
...
+
+/// <summary>
+/// Experiment to validate MavenWithFallbackDetector against MvnCliComponentDetector.
+/// The MavenWithFallbackDetector combines MvnCli detection with static pom.xml parsing fallback
+/// for cases where Maven CLI fails (e.g., authentication errors with private feeds).
+/// </summary>
+public class MavenWithFallbackExperiment : IExperimentConfiguration
+{
+    /// <inheritdoc />
+    public string Name => "MavenWithFallbackExperiment";
```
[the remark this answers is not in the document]
grvillic: @zhenghao104 - This is a fair suggestion

src/Microsoft.ComponentDetection.Orchestrator/Experiments/Configs/MavenWithFallbackExperiment.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    /// <inheritdoc />
+    public bool IsInControlGroup(IComponentDetector componentDetector) => componentDetector is MvnCliComponentDetector;
+
+    /// <inheritdoc />
+    public bool IsInExperimentGroup(IComponentDetector componentDetector) => componentDetector is MavenWithFallbackDetector;
+
+    /// <inheritdoc />
+    public bool ShouldRecord(IComponentDetector componentDetector, int numComponents) => true;
```
grvillic: only record if either `MavenWithFallbackDetector` or `MvnCliComponentDetector` have at least one component, otherwise you will be just flushing experiments without any meaningful data.

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,753 @@
...
+    /// </summary>
+    internal const string DisableMvnCliEnvVar = "CD_MAVEN_DISABLE_CLI";
+
+    /// <summary>
+    /// Environment variable to enable this detector and disable MvnCliComponentDetector.
+    /// Set to "true" to use this detector instead of MvnCliComponentDetector.
+    /// Usage: Set CD_USE_MAVEN_FALLBACK_DETECTOR=true as a pipeline/environment variable.
+    /// When set, this detector will fully run (including Maven CLI) and MvnCliComponentDetector will be disabled.
+    /// </summary>
+    internal const string UseFallbackDetectorEnvVar = "CD_USE_MAVEN_FALLBACK_DETECTOR";
```
grvillic: This variable shouldn't exist, this fallback detector will only be running as experimental, once we confirm things look good we can copy code from here and place inside mvncli and remove this file.

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,753 @@
...
+    [
+        "401",
+        "403",
+        "Unauthorized",
+        "Not authorized",
+        "authentication",
+        "Could not authenticate",
+        "Access denied",
+        "Forbidden",
+        "status code: 401",
```
grvillic: Aren't this already covered by line 106 and 107?

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,753 @@
...
+    private void LogDebug(string message) =>
+        this.Logger.LogDebug("{DetectorId}: {Message}", this.Id, message);
+
+    private void LogInfo(string message) =>
+        this.Logger.LogInformation("{DetectorId}: {Message}", this.Id, message);
+
+    private void LogWarning(string message) =>
+        this.Logger.LogWarning("{DetectorId}: {Message}", this.Id, message);
+
+    protected override async Task<IObservable<ProcessRequest>> OnPrepareDetectionAsync(
```
grvillic, line 163: This method has significant code >~100 lines, let's break it down into private methods to make it easier to follow/troubleshoot.

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,747 @@
...
+            this.fallbackReason = MavenFallbackReason.AuthenticationFailure;
+
+            // Extract failed endpoints from error messages
+            foreach (var error in this.mavenCliErrors)
+            {
+                var endpoints = this.ExtractFailedEndpoints(error);
+                foreach (var endpoint in endpoints)
+                {
+                    this.failedEndpoints.Add(endpoint);
+                }
```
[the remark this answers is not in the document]
grvillic: @zhenghao104 - This is a fair observation, just ensure it handle null mavenCliErrors object gracefully as if it was empty

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
zhenghao104: Removed, and updated the markdown

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
zhenghao104: Done.

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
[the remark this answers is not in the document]
zhenghao104: mavenCliErrors is always initialized to an empty array so this should work well, will update the code.

src/Microsoft.ComponentDetection.Orchestrator/Experiments/Configs/MavenWithFallbackExperiment.cs
zhenghao104: Added `numComponents > 0`

src/Microsoft.ComponentDetection.Orchestrator/Experiments/Configs/MavenWithFallbackExperiment.cs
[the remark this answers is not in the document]
zhenghao104: Fixed

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,753 @@
...
+                else
+                {
+                    this.Logger.LogDebug(
+                        "Version string for component {Group}/{Artifact} is invalid or unsupported and a component will not be recorded.",
+                        groupId,
+                        artifactId);
+                }
+            }
+
+            this.staticParserComponentCount += componentsFoundInFile;
```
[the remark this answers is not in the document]
zhenghao104: Added

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
zhenghao104: Trimmed it to

   private static readonly string[] AuthErrorPatterns =
    [
        "401",
        "403",
        "Unauthorized",
        "Access denied",
    ];

src/Microsoft.ComponentDetection.Orchestrator/Experiments/Configs/MavenWithFallbackExperiment.cs
```diff
@@ -0,0 +1,24 @@
...
+    public string Name => "MavenWithFallback";
+
+    /// <inheritdoc />
+    public bool IsInControlGroup(IComponentDetector componentDetector) => componentDetector is MvnCliComponentDetector;
+
+    /// <inheritdoc />
+    public bool IsInExperimentGroup(IComponentDetector componentDetector) => componentDetector is MavenWithFallbackDetector;
+
+    /// <inheritdoc />
+    public bool ShouldRecord(IComponentDetector componentDetector, int numComponents) => numComponents > 0;
```
grvillic: See other experiments, just looking at components will cause to skip the whole experiment altogether, you should always return true, unless the new detector does not return any dependencies.

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,818 @@
...
+                    {
+                        Stream = new MemoryStream(Encoding.UTF8.GetBytes(content)),
+                        Location = componentStream.Location,
+                        Pattern = BcdeMvnDepsFileName,
+                    },
+                    SingleFileComponentRecorder = this.ComponentRecorder.CreateSingleFileComponentRecorder(
+                        Path.Combine(depsDir, MavenManifest)),
+                };
+            })
+            .ToList();
```
grvillic: Is this the preexisting behavior of MvnCLI? Seem risky to be enumerating streams within the same file.

src/Microsoft.ComponentDetection.Orchestrator/Experiments/Configs/MavenWithFallbackExperiment.cs
zhenghao104: updated with

```
    public bool ShouldRecord(IComponentDetector componentDetector, int numComponents)
    {
        // Only record telemetry if a Maven detector found components,
        // indicating Maven projects were detected and scanned.
        if (componentDetector is MvnCliComponentDetector or MavenWithFallbackDetector)
        {
            return numComponents > 0;
        }

        return true;
    }
```

Looking at the other experiments and the one I wrote for nuget, this should be correct.

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,818 @@
...
+        }
+
+        // Run Maven CLI detection on all pom.xml files
+        var pomFileDirectories = await this.RunMavenCliDetectionAsync(processRequests, cancellationToken);
+
+        // Collect results and determine which files succeeded/failed
+        var (mvnCliResults, failedPomFiles) = this.CollectMvnCliResults(pomFileDirectories);
+
+        // Determine final detection method and return appropriate results
+        return this.DetermineDetectionOutcome(mvnCliResults, failedPomFiles, pomFileDirectories.Count);
```
grvillic: Confirm this method is protected by the experimental coverage. IIRC it does not have the same guardrails as "OnFileFound", hence wrap all of this with a try catch and chain an additional cancellation token so the execution stops within 4-6 minutes. Assuming most repos complete the full scan within this window. That way we can see execution hit limits in our telemetry but no unexpected hangs for existing users.

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,818 @@
...
+        if (successCount == 0)
+        {
+            this.LogWarning("Maven CLI did not produce any dependency files. Falling back to static pom.xml parsing.");
+            this.AnalyzeMvnCliFailure();
+            this.usedDetectionMethod = MavenDetectionMethod.StaticParserOnly;
+            return this.GetAllPomFilesForStaticParsing();
+        }
+
+        // Partial success - use MvnCli results for successful ones, static parsing for failed ones
+        this.LogWarning($"Maven CLI failed for {failedCount} pom.xml files. Using MvnCli results for successful files and falling back to static parsing for failed files.");
```
grvillic: I haven't manually validated these changes, but have we validated that if a root pom.xml file succeeded with mvn cli, we are excluding child ones? It will be unnecessary work if run CLI or parse directly pom.xml on child pom.xml when the root one covered them. Although the manual parsing is faster, the slow one that significantly impacts perf is MvnCLI.

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,818 @@
...
+        this.usedDetectionMethod = MavenDetectionMethod.Mixed;
+
+        // Get static parsing results for failed pom.xml files
+        var staticParsingResults = this.GetStaticParsingRequestsForFailedFiles(failedPomFiles);
+
+        // Combine MvnCli results with static parsing fallback for failed files
+        return mvnCliResults.Concat(staticParsingResults).ToObservable();
+    }
+
+    protected override async Task OnFileFoundAsync(
```
grvillic: nit: I don't think there is the need to add the `async/await` here, we only need to have a `Task` output.

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,818 @@
...
+            this.fallbackReason = MavenFallbackReason.MavenCliNotAvailable;
+            return false;
+        }
+
+        this.LogDebug("Maven CLI is available. Running MvnCli detection.");
+
+        // Create a temporary Maven local repository for this detector to avoid file locking conflicts
+        // with MvnCliComponentDetector which uses the default ~/.m2/repository
+        this.tempMavenLocalRepository = Path.Combine(Path.GetTempPath(), $"maven-fallback-detector-{Guid.NewGuid():N}");
+        Directory.CreateDirectory(this.tempMavenLocalRepository);
```
FernandoRojo: Discussed offline, i suggested using a semaphor pattern instead of the temp maven repo to avoid a second downlaoding of all dependencies in the parallel case, and just have the two cli detectors share the output file:

```
 public async Task GenerateDependenciesFileAsync(ProcessRequest processRequest, CancellationToken cancellationToken = default)
    {
        var pomFile = processRequest.ComponentStream;
        var semaphore = this.locationLocks.GetOrAdd(pomFile.Location, _ => new SemaphoreSlim(1, 1));

        await semaphore.WaitAsync(cancellationToken);
        try
        {
            if (this.completedLocations.ContainsKey(pomFile.Location))
            {
                this.logger.LogDebug("{DetectorPrefix}: Skipping duplicate \"dependency:tree\" for {PomFileLocation}, already generated", DetectorLogPrefix, pomFile.Location);
                return;
            }

            await this.GenerateDependenciesFileCoreAsync(processRequest, cancellationToken);
            this.completedLocations.TryAdd(pomFile.Location, true);
        }
        finally
        {
            semaphore.Release();
        }
    }

    private async Task GenerateDependenciesFileCoreAsync(ProcessRequest processRequest, CancellationToken cancellationToken)
```

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,818 @@
...
+                            if (last != null && !topLevelDirectories.ContainsKey(last.Name))
+                            {
+                                topLevelDirectories.TryAdd(last.Name, last);
+                            }
+
+                            this.LogDebug($"Discovered {item.Location}.");
+                            s.OnNext(processRequest);
+                            break;
+                        }
+
```
FernandoRojo, line 793: Found a bug in this already existing code: 
This does rely on the ordering of the process requests to make sure the ancestral path comes in first. If a child path is earlier in the list, then it won't get filtered out, I do think the component streams by default are in ancestral order, but I wouldn't rely on that

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
zhenghao104: Yes, https://github.com/microsoft/component-detection/blob/67481944d112efda0974ae30e763f53dd2aff885/src/Microsoft.ComponentDetection.Detectors/maven/MvnCliComponentDetector.cs#L74

But we have two other locations in this file with this behavior to enumerate the files for static parsing when falling back

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
FernandoRojo: A version that removes that bug and also makes this a lot more readable:

```
private IObservable<ProcessRequest> RemoveNestedPomXmls(IObservable<ProcessRequest> componentStreams)
    {
        return componentStreams
            .ToList()
            .SelectMany(allRequests =>
            {
                // Build a set of all directories that contain a pom.xml.
                var pomDirectories = new HashSet<string>(
                    allRequests.Select(r => Path.GetDirectoryName(r.ComponentStream.Location).TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar) + Path.DirectorySeparatorChar),
                    StringComparer.OrdinalIgnoreCase).OrderBy(d => d.Length).ToList(); // Order by ascending length to ensure possible parent paths are checked first.

                return allRequests.Where(request =>
                {
                    var location = Path.GetDirectoryName(request.ComponentStream.Location).TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar) + Path.DirectorySeparatorChar;
                    foreach (var pomDirectory in pomDirectories)
                    {
                        if (pomDirectory.Length >= location.Length)
                        {
                            // Since the list is ordered by length, If the pomDirectory is longer than or equal to the location, there is no possible parent left.
                            break;
                        }

                        if (location.StartsWith(pomDirectory, StringComparison.OrdinalIgnoreCase))
                        {
                            this.LogDebugWithId($"Ignoring {MavenManifest} at {location}, as it has a parent {MavenManifest} that will be processed at {pomDirectory}{MavenManifest}.");
                            return false;
                        }
                    }

                    this.LogDebugWithId($"Discovered {request.ComponentStream.Location}.");
                    return true;
                });
            });
    }
```

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,818 @@
...
+        return pomFileDirectories;
+    }
+
+    /// <summary>
+    /// Collects Maven CLI results and identifies which pom.xml files failed.
+    /// </summary>
+    /// <param name="pomFileDirectories">Dictionary of pom.xml directories to their process requests.</param>
+    /// <returns>A tuple containing the successful MvnCli results and the list of failed pom files.</returns>
+    private (List<ProcessRequest> MvnCliResults, List<ProcessRequest> FailedPomFiles) CollectMvnCliResults(
```
FernandoRojo: Instead of going through all this, why not just populate lists of succededCLI and failedCLI during RunMavenCliDetectionAsync (at line 265)

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
zhenghao104: Looking at the code, the guardrails should apply to `OnPrepareDetectionAsync`  https://github.com/microsoft/component-detection/blob/67481944d112efda0974ae30e763f53dd2aff885/src/Microsoft.ComponentDetection.Orchestrator/Services/DetectorProcessingService.cs#L325

https://github.com/microsoft/component-detection/blob/67481944d112efda0974ae30e763f53dd2aff885/src/Microsoft.ComponentDetection.Contracts/FileComponentDetector.cs#L128

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,818 @@
...
+        // All succeeded - use MvnCli results only
+        if (failedCount == 0 && successCount > 0)
+        {
+            this.usedDetectionMethod = MavenDetectionMethod.MvnCliOnly;
+            this.LogDebug("All pom.xml files processed successfully with Maven CLI.");
+            return mvnCliResults.ToObservable();
+        }
+
+        // All failed - fall back to static parsing for everything
+        if (successCount == 0)
```
FernandoRojo: This logic block for determining which files to put in the observable can be simplified, 

In RunMavenCliDetectionAsync, if the cli command succeeds, then the deps file can be added to the processing set, if it fails, then the pom file should be added instead.

FernandoRojo: All the repeated enumerations of the component streams is uneccessary

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
zhenghao104: Yes. I also added unit tests for this scenario

`WhenMvnCliSucceeds_NestedPomXmlsAreFilteredOut_Async`

Tests that when MvnCli succeeds for root pom.xml, nested module pom.xml files are filtered out
Verifies [GenerateDependenciesFileAsync](vscode-file://vscode-app/c:/Users/zhentan/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is called only once (for root only)
Confirms all components come from MvnCli, not static parsing

`WhenMvnCliFailsCompletely_AllNestedPomXmlsAreRestoredForStaticParsing_Async`

Tests that when MvnCli fails completely, all pom.xml files (including nested) are restored for static parsing

`WhenMvnCliPartiallyFails_NestedPomXmlsRestoredOnlyForFailedDirectories_Async`

Tests the mixed scenario: ProjectA succeeds with MvnCli, ProjectB fails
Verifies nested pom in ProjectA is NOT statically parsed (MvnCli handled it)
Verifies nested pom in ProjectB IS statically parsed (MvnCli failed)
Key assertion: [artifactIds.Should().NotContain("dep-from-nested-a")](vscode-file://vscode-app/c:/Users/zhentan/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) confirms children of successful roots are excluded

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
zhenghao104: Fixed

src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
```diff
@@ -78,10 +91,22 @@ public async Task GenerateDependenciesFileAsync(ProcessRequest processRequest, C
...
+
+            // Read the generated dependencies file and return its content
+            var depsFilePath = Path.Combine(Path.GetDirectoryName(pomFile.Location), outputFileName);
+            string depsFileContent = null;
+            if (File.Exists(depsFilePath))
+            {
+                depsFileContent = await File.ReadAllTextAsync(depsFilePath, cancellationToken);
+            }
+
+            return new MavenCliResult(true, null, depsFileContent);
```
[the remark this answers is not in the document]
zhenghao104: Updated

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
zhenghao104: Fixed

src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
```diff
@@ -78,10 +91,22 @@ public async Task GenerateDependenciesFileAsync(ProcessRequest processRequest, C
...
         else
         {
             this.logger.LogDebug("{DetectorPrefix}: Execution of \"dependency:tree\" on {PomFileLocation} completed successfully", DetectorLogPrefix, pomFile.Location);
+
+            // Read the generated dependencies file and return its content
+            var depsFilePath = Path.Combine(Path.GetDirectoryName(pomFile.Location), outputFileName);
+            string depsFileContent = null;
+            if (File.Exists(depsFilePath))
+            {
+                depsFileContent = await File.ReadAllTextAsync(depsFilePath, cancellationToken);
```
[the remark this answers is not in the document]
zhenghao104: Updated

src/Microsoft.ComponentDetection.Orchestrator/Experiments/Configs/MavenWithFallbackExperiment.cs
zhenghao104: Discussed offline: experiment needs to be in CD internal repo to compare it with legacy maven detector

src/Microsoft.ComponentDetection.Orchestrator/Experiments/Configs/MavenWithFallbackExperiment.cs
zhenghao104: Discussed offline: experiment needs to be in CD internal repo to compare it with legacy maven detector

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
zhenghao104: Added some new logic to account for this. Let me know if you want to discuss offline

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
zhenghao104: Added

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
zhenghao104: Updated

src/Microsoft.ComponentDetection.Orchestrator/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -114,6 +115,7 @@ public static IServiceCollection AddComponentDetection(this IServiceCollection s
         services.AddSingleton<IMavenCommandService, MavenCommandService>();
         services.AddSingleton<IMavenStyleDependencyGraphParserService, MavenStyleDependencyGraphParserService>();
         services.AddSingleton<IComponentDetector, MvnCliComponentDetector>();
+        services.AddSingleton<IComponentDetector, MavenWithFallbackDetector>();
```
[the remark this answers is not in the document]
zhenghao104: With the semaphor implementation, this should be ok. The experiment config will also be in CD internal repository.

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,778 @@
...
+            File.Delete(processRequest.ComponentStream.Location);
+        }
+        catch
+        {
+            // Ignore deletion errors
+        }
+
+        var newCount = this.ComponentRecorder.GetDetectedComponents().Count();
+        Interlocked.Add(ref this.mvnCliComponentCount, newCount - initialCount);
+    }
```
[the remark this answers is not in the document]
zhenghao104: Updated

src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
```diff
@@ -58,7 +68,10 @@ public async Task GenerateDependenciesFileAsync(ProcessRequest processRequest, C
         var pomFile = processRequest.ComponentStream;
         this.logger.LogDebug("{DetectorPrefix}: Running \"dependency:tree\" on {PomFileLocation}", DetectorLogPrefix, pomFile.Location);
 
-        var cliParameters = new[] { "dependency:tree", "-B", $"-DoutputFile={this.BcdeMvnDependencyFileName}", "-DoutputType=text", $"-f{pomFile.Location}" };
+        // Build CLI parameters - add custom local repository if specified
+        string[] cliParameters = string.IsNullOrEmpty(localRepositoryPath)
+            ? ["dependency:tree", "-B", $"-DoutputFile={outputFileName}", "-DoutputType=text", $"-f{pomFile.Location}"]
+            : ["dependency:tree", "-B", $"-DoutputFile={outputFileName}", "-DoutputType=text", $"-f{pomFile.Location}", $"-Dmaven.repo.local={localRepositoryPath}"];
```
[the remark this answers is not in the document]
zhenghao104: We do not need custom local path, reverted change to match existing behavior for MavenCLI

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,778 @@
...
+            var depsFilePath = Path.Combine(pomDir, depsFileName);
+
+            // Generate dependency file using Maven CLI.
+            // Note: If both MvnCliComponentDetector and this detector are enabled,
+            // they may run Maven CLI on the same pom.xml independently.
+            var result = await this.mavenCommandService.GenerateDependenciesFileAsync(
+                processRequest,
+                depsFileName,
+                cancellationToken);
+
```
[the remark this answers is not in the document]
zhenghao104: With the semaphor implementation, this should be ok. The experiment config will also be in CD internal repository.

test/Microsoft.ComponentDetection.Detectors.Tests/MavenWithFallbackDetectorTests.cs
```diff
@@ -0,0 +1,1079 @@
...
+
+        // Setup file utility service to return the deps file content
+        // The detector reads the file from disk after CLI succeeds
+        this.fileUtilityServiceMock.Setup(x => x.Exists(It.Is<string>(s => s.EndsWith(BcdeMvnFileName))))
+            .Returns(true);
+        this.fileUtilityServiceMock.Setup(x => x.ReadAllText(It.Is<string>(s => s.EndsWith(BcdeMvnFileName))))
+            .Returns(content);
+
+        // Only need to create a pom.xml file - the deps file content comes from the mock file service
+        this.DetectorTestUtility.WithFile("pom.xml", content);
```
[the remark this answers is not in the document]
zhenghao104: Good suggestion, added

src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
```diff
@@ -43,7 +57,42 @@ public async Task<bool> MavenCLIExistsAsync()
...
+
+        await semaphore.WaitAsync(CancellationToken.None);
+        try
+        {
+            // Check if another detector already generated the deps file for this location
+            if (this.completedLocations.TryGetValue(pomFile.Location, out var cachedResult))
+            {
+                this.logger.LogDebug("{DetectorPrefix}: Skipping duplicate \"dependency:tree\" for {PomFileLocation}, already generated", DetectorLogPrefix, pomFile.Location);
+                return cachedResult;
+            }
```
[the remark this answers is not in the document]
zhenghao104: Added file existence check to ensures that if a previous consumer deleted bcde.mvndeps, the next caller will re-run Maven CLI instead of getting a stale "success" result.

src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
```diff
@@ -43,7 +57,42 @@ public async Task<bool> MavenCLIExistsAsync()
...
+
+            var result = await this.GenerateDependenciesFileCoreAsync(processRequest, outputFileName, cancellationToken);
+            this.completedLocations.TryAdd(pomFile.Location, result);
+            return result;
```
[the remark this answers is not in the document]
zhenghao104: Updated to only cache successful result

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,779 @@
...
+                failedDirectories.Add(pomDir);
+
+                // Capture error output for later analysis
+                if (!string.IsNullOrWhiteSpace(result.ErrorOutput))
+                {
+                    this.mavenCliErrors.Add(result.ErrorOutput);
+                }
+            }
+        });
+
```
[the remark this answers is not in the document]
zhenghao104: Updated comment to keep sequential logic

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,780 @@
...
+    /// A pom.xml is considered nested if there's another pom.xml in a parent directory.
+    /// </summary>
+    /// <param name="componentStreams">The incoming process requests for pom.xml files.</param>
+    /// <returns>Process requests for only root-level pom.xml files.</returns>
+    private IObservable<ProcessRequest> RemoveNestedPomXmls(IObservable<ProcessRequest> componentStreams)
+    {
+        return componentStreams
+            .ToList()
+            .SelectMany(allRequests =>
+            {
```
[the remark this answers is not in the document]
zhenghao104: @FernandoRojo Would this be an issue?

src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
```diff
@@ -19,6 +20,19 @@ public class MavenCommandService : IMavenCommandService
...
+    /// This allows multiple detectors (e.g., MvnCliComponentDetector and MavenWithFallbackDetector)
+    /// to safely share the same output file without race conditions.
+    /// </summary>
+    private readonly ConcurrentDictionary<string, SemaphoreSlim> locationLocks = new();
+
+    /// <summary>
+    /// Tracks locations where dependency generation has completed successfully.
+    /// Used to skip duplicate executions when multiple detectors process the same pom.xml.
+    /// </summary>
+    private readonly ConcurrentDictionary<string, MavenCliResult> completedLocations = new();
```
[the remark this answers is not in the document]
zhenghao104: Added a clean up method in `OnDetectionFinishedAsync` to clear cache

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,780 @@
...
+        // Try to delete the deps file
+        try
+        {
+            File.Delete(processRequest.ComponentStream.Location);
+        }
+        catch
+        {
+            // Ignore deletion errors
+        }
+
```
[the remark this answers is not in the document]
zhenghao104: File.Delete will not throw on non existent file, so this should be ok

src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
```diff
@@ -19,6 +20,19 @@ public class MavenCommandService : IMavenCommandService
...
+
```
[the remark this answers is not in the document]
zhenghao104: Added some tests

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,780 @@
...
+/// <summary>
+/// Enum representing why fallback occurred.
+/// </summary>
+internal enum MavenFallbackReason
+{
+    /// <summary>No fallback was needed.</summary>
+    None,
+
+    /// <summary>Maven CLI was explicitly disabled via detector argument.</summary>
```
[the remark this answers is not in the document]
zhenghao104: Fixed

test/Microsoft.ComponentDetection.Detectors.Tests/MavenWithFallbackDetectorTests.cs
```diff
@@ -0,0 +1,1089 @@
...
+        // Assert
+        detectorResult.ResultCode.Should().Be(ProcessingResultCode.Success);
+
+        var detectedComponents = componentRecorder.GetDetectedComponents();
+        detectedComponents.Should().HaveCount(3);
+
+        // Verify dependency graph has the transitive relationship
+        var dependencyGraph = componentRecorder.GetDependencyGraphsByLocation().Values.First();
+        dependencyGraph.Should().NotBeNull();
+    }
```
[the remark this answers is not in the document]
zhenghao104: Added

src/Microsoft.ComponentDetection.Detectors/maven/IMavenCommandService.cs
```diff
@@ -11,7 +11,9 @@ public interface IMavenCommandService
 
     Task<bool> MavenCLIExistsAsync();
 
-    Task GenerateDependenciesFileAsync(ProcessRequest processRequest, CancellationToken cancellationToken = default);
+    Task<MavenCliResult> GenerateDependenciesFileAsync(ProcessRequest processRequest, CancellationToken cancellationToken = default);
+
+    Task<MavenCliResult> GenerateDependenciesFileAsync(ProcessRequest processRequest, string outputFileName, CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
zhenghao104: No longer valid

src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
```diff
@@ -43,7 +57,54 @@ public async Task<bool> MavenCLIExistsAsync()
...
+        var depsFilePath = Path.Combine(pomDir, outputFileName);
+
+        // Use semaphore to prevent concurrent Maven CLI executions for the same pom.xml.
+        // This allows MvnCliComponentDetector and MavenWithFallbackDetector to safely share the output file.
+        // Note: We don't pass the cancellation token to WaitAsync so that we can still check the cache
+        // even if cancellation is requested. The cancellation will be honored during the actual CLI execution.
+        var semaphore = this.locationLocks.GetOrAdd(pomFile.Location, _ => new SemaphoreSlim(1, 1));
+
+        await semaphore.WaitAsync(CancellationToken.None);
+        try
```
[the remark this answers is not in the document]
zhenghao104: Made Changes: Check cache before acquiring semaphore and then pass cancellation token to WaitAsync

src/Microsoft.ComponentDetection.Detectors/maven/MvnCliComponentDetector.cs
```diff
@@ -102,6 +102,12 @@ protected override async Task OnFileFoundAsync(ProcessRequest processRequest, ID
         await Task.CompletedTask;
     }
 
+    protected override Task OnDetectionFinishedAsync()
+    {
+        this.mavenCommandService.ClearCache();
```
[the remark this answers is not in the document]
zhenghao104: Deleted ClearCache() entirely

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,783 @@
...
+            this.Telemetry["FailedEndpoints"] = string.Join(";", this.failedEndpoints.Distinct().Take(10));
+        }
+
+        this.LogInfo($"Detection completed. Method: {this.usedDetectionMethod}, " +
+                     $"FallbackReason: {this.fallbackReason}, " +
+                     $"MvnCli components: {this.mvnCliComponentCount}, " +
+                     $"Static parser components: {this.staticParserComponentCount}");
+
+        // Clear the Maven command service cache to prevent unbounded memory growth
+        this.mavenCommandService.ClearCache();
```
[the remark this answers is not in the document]
zhenghao104: Deleted ClearCache() entirely

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,780 @@
...
+    /// Runs Maven CLI detection on all root pom.xml files.
+    /// For each pom.xml, if CLI succeeds, the deps file is added to results.
+    /// If CLI fails, all pom.xml files under that directory are added for static parsing fallback.
+    /// </summary>
+    /// <param name="processRequests">The incoming process requests.</param>
+    /// <param name="cancellationToken">Cancellation token for the operation.</param>
+    /// <returns>An observable of process requests (deps files for CLI success, pom.xml for CLI failure).</returns>
+    private async Task<IObservable<ProcessRequest>> RunMavenCliDetectionAsync(
+        IObservable<ProcessRequest> processRequests,
+        CancellationToken cancellationToken)
```
grvillic, line 270: I don't think we are using the cancellationToken in all places we should, we should force the cancellation if requested within loops and between large chunks of work not only CLI

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,780 @@
...
+
+        return results.ToObservable();
+    }
+
+    /// <summary>
+    /// Gets all pom.xml files in the specified directories and their subdirectories for static parsing.
+    /// </summary>
+    /// <param name="directories">The directories to scan for pom.xml files.</param>
+    /// <returns>ProcessRequests for all pom.xml files in the specified directories.</returns>
+    private IEnumerable<ProcessRequest> GetAllPomFilesInDirectories(HashSet<string> directories)
```
grvillic: Cancellation token should also be provided here, we shouldn't have to wait to enumerate all component streams to stop the execution.

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
zhenghao104: Added

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
zhenghao104: Added

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,813 @@
...
+                Interlocked.Increment(ref cliFailureCount);
+                failedDirectories.Add(pomDir);
+
+                // Capture error output for later analysis
+                if (!string.IsNullOrWhiteSpace(result.ErrorOutput))
+                {
+                    this.mavenCliErrors.Add(result.ErrorOutput);
+                }
+            }
+        });
```
grvillic: The Action Block should also receive the cancellation token itself, see `ExecutionDataflowBlockOptions`

zhenghao104: Added

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -0,0 +1,817 @@
...
+
+        // Try to delete the deps file
+        try
+        {
+            File.Delete(processRequest.ComponentStream.Location);
+        }
+        catch
+        {
+            // Ignore deletion errors
+        }
```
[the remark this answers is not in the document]
zhenghao104: File.Delete will not throw on non existent file, so this should be ok

PR #1633, opened 2026-02-18
Fix Docker image ID test for Docker 29

JamieMagee:
GitHub Actions runners were [updated to Docker 29.1.x on February 9](https://github.com/actions/runner-images/issues/13474). Docker 29 [makes the containerd image store the default](https://docs.docker.com/engine/release-notes/29/#2900) for fresh installs, which changes how image IDs are reported.

With the legacy graphdriver, `docker inspect` returned the **config digest** as the image ID. With the containerd image store, it returns the **manifest digest** instead. The test image hasn't changed -- both digests are valid for it -- but `DockerService_PopulatesBaseImageAndLayerDetailsAsync` was hardcoded to expect only the config digest.

This updates the assertion to accept either digest, so the test passes on both old and new Docker versions.


JamieMagee:
Verification test failures are expected. We are comparing the output from Docker 29, with the old, cached, output from an earlier version of Docker. The verification test output will be updated and cached with Docker 29 after merge to `main`.

PR #1661, opened 2026-02-18
Bump commons-text to 1.12.0 in maven verification test resources

JamieMagee:
Bumps commons-text from 1.9 to 1.12.0 in the maven verification test pom.xml. The old version has a known RCE via string interpolation (CVE-2022-42889).

### DIFF
--- a/.github/workflows/snapshot-verify.yml
+++ b/.github/workflows/snapshot-verify.yml
@@ -53,6 +53,7 @@ jobs:
 
       - name: Download latest release snapshot
         working-directory: ${{ github.workspace }}/release-output
+        shell: bash
         env:
           DOWNLOAD_URL: ${{ steps.download-latest-release-snapshot.outputs.result }}
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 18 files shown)


JamieMagee:
The snapshot verification mismatch is expected, from changing a verification test resource.

The `snapshot-verify` workflow downloads the last published scan result from `main` (which has `commons-text:1.9`) and compares it against a fresh scan of this branch (which has `commons-text:1.12.0`). Since the version changed, the old and new scan results don't match:

- `CheckManifestFiles_ExcludingExperimentalDetectors` fails because `commons-text:1.9` is missing from the new scan and `commons-text:1.12.0` is missing from the old one. Any changed transitives will also mismatch.
- `CheckDetectorsRunTimesAndCounts` will flag the Maven detector's component count changing without a detector version bump.

This is a test resource change, not a detector change, so the failure is expected. Once this merges, `snapshot-publish` on `main` will produce a new baseline with `1.12.0` and future PRs will pass again.

The unit test references to `commons-text:1.9` in `MavenWithFallbackDetectorTests.cs` are independent fixtures, not tied to the verification resources.

PR #1663, opened 2026-02-18
Bump Microsoft.Owin packages to 4.2.2 in nuget verification test resources

JamieMagee:
Bumps Microsoft.Owin and Microsoft.Owin.Host.SystemWeb from 3.1.0 to 4.2.2 in the nuget packages.config verification test. The 3.x versions have a known authentication bypass.

### DIFF
--- a/.github/workflows/snapshot-verify.yml
+++ b/.github/workflows/snapshot-verify.yml
@@ -53,6 +53,7 @@ jobs:
 
       - name: Download latest release snapshot
         working-directory: ${{ github.workspace }}/release-output
+        shell: bash
         env:
           DOWNLOAD_URL: ${{ steps.download-latest-release-snapshot.outputs.result }}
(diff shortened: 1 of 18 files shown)


JamieMagee:
The snapshot verification failures are expected and not a problem with the PR.

The verification tests compare the PR branch's scan output against a baseline artifact from the last `main` build. The baseline has `Microsoft.Owin 3.1.0` and `Microsoft.Owin.Host.SystemWeb 3.1.0`. The PR scan produces `4.2.2` for both. The `CheckManifestFiles_ExcludingExperimentalDetectors` test sees two components missing from each side and fails.

The detector logic hasn't changed, just the test input data. Component count per detector stays the same.

Once this merges, `snapshot-publish.yml` runs on `main` and regenerates the baseline with the `4.2.2` versions. After that, verification tests go back to green for other PRs.

PR #1665, opened 2026-02-18
Bump express, typescript, re2, @types/react in pnpm verification test resources

JamieMagee:
Bumps dependencies in both pnpm v5 and v6 verification test fixtures:

- express 4.17.2 -> 4.22.1
- typescript 4.5.5 -> 4.9.5
- @types/react 17.0.38 -> 17.0.91
- re2 1.17.3 -> 1.23.3

Same dep updates as the npm PR, applied to the pnpm lockfiles.

### DIFF
--- a/.github/workflows/snapshot-verify.yml
+++ b/.github/workflows/snapshot-verify.yml
@@ -53,6 +53,7 @@ jobs:
 
       - name: Download latest release snapshot
         working-directory: ${{ github.workspace }}/release-output
+        shell: bash
         env:
           DOWNLOAD_URL: ${{ steps.download-latest-release-snapshot.outputs.result }}
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -32,6 +32,7 @@
     <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" PrivateAssets="all"/>
     <PackageReference Include="MinVer" PrivateAssets="all"/>
     <PackageReference Include="StyleCop.Analyzers" PrivateAssets="all"/>
+    <PackageReference Include="System.Text.Json" />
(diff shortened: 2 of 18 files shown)


JamieMagee:
The snapshot verification failure is expected.

The snapshot-verify workflow downloads the baseline scan output from `main` (built against the old lockfiles) and compares it to a fresh scan of this PR's updated lockfiles. Since we bumped express, typescript, @types/react, and re2, the transitive dependency trees changed significantly — so the detected component sets no longer match.

Specifically:

- **~35 packages disappeared** from the old dependency trees: `@gar/promisify`, `@npmcli/move-file`, `aggregate-error`, `are-we-there-yet`, `delegates`, `gauge`, `humanize-ms`, `indent-string`, `inflight`, `npmlog`, `once`, `rimraf`, `wrappy`, etc. These were transitive deps of older express/re2/node-gyp versions.

- **~25 new packages appeared**: `@isaacs/fs-minipass`, `@npmcli/agent`, `call-bind-apply-helpers`, `dunder-proto`, `es-errors`, `es-object-atoms`, `fdir`, `get-intrinsic`, `gopd`, `hasown`, `picomatch`, `proc-log`, `side-channel`, `tinyglobby`, etc. These come from the updated dependency trees.

- **Many packages changed versions**: `accepts` 1.3.7→1.3.8, `body-parser` 1.19.1→1.20.4, `cookie` 0.4.1→0.7.2, `depd` 1.1.2→2.0.0, `statuses` 1.5.0→2.0.2, and so on.

The `CheckManifestFiles_ExcludingExperimentalDetectors` test catches all of this, and `CheckDetectorsRunTimesAndCounts` likely flags the different component count too.

This is the same situation as the yarn (#1667), maven (#1661), and nuget (#1663) resource bumps. Once merged, `snapshot-publish` runs on `main` and publishes a new baseline that includes these updated lockfiles.

PR #1669, opened 2026-02-18
Fix System.Text.Json MSB3277 version conflict

JamieMagee:
The Contracts project targets netstandard2.0 and pulls `System.Text.Json` 9.0.x from NuGet. Every other project targets net8.0, which ships with `System.Text.Json` 8.0.x in the framework. When MSBuild sees both versions flowing into the same project, it picks the framework copy and warns:

```
warning MSB3277: Found conflicts between different versions of "System.Text.Json" that could not be resolved.
```

This hits Common, Detectors, Orchestrator, the CLI, and the test projects -- basically everything that references Contracts directly or transitively.

The fix is a one-line addition to `Directory.Build.props`: an explicit `System.Text.Json` PackageReference. This tells MSBuild to prefer the NuGet package (9.0.x, version from `Directory.Packages.props`) over the framework assembly in all net8.0 projects. The Contracts project already had its own reference, so it's unaffected.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -32,6 +32,7 @@
     <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" PrivateAssets="all"/>
     <PackageReference Include="MinVer" PrivateAssets="all"/>
     <PackageReference Include="StyleCop.Analyzers" PrivateAssets="all"/>
+    <PackageReference Include="System.Text.Json" />
   </ItemGroup>
 
 </Project>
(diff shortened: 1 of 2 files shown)


ericstj:
Normally this wouldn't happen but it's a side effect of https://github.com/microsoft/component-detection/blob/ebb5d909753359215f4b91e54bf984755c2f4fec/Directory.Packages.props#L3-L6

Thats preventing the version of the package your references use from reaching RAR.  RAR still sees the indirect reference and warns.  

Another smaller version of this workaround might just be to reference STJ from the contracts project and set PrivateAssets=none on the reference.  Probably you should do that for any framework overlapping package.

PR #1677, opened 2026-02-27
Update npm lockfile 3 ancestor traversal to properly detect dependency roots

FernandoRojo:
NpmLockfile3Detector  had a path-resolution bug when resolving nested dependencies in lockfile v3.
The previous logic tried to reconstruct ancestor lookup paths from dependency-graph component IDs, inproperly traversing combinations of ancesters, instead of walking real lockfile package paths. That can produce invalid candidates and miss valid packages that are hoisted/placed under an ancestor path.

**Root cause:** ancestor traversal was based on component identity and an erroneous ancestor path determination., not the current lockfile path hierarchy.
**Effect:** a dependency can be reported as “not found” (or resolved to an incorrect fallback) even though it exists in packages.
**Given:**

  node_modules/A
  node_modules/A/node_modules/B
  node_modules/A/node_modules/C ← C exists under ancestor A, not under B or root

If B declares dependency C, correct npm-style resolution is:

check 
    node_modules/A/node_modules/B/node_modules/C (missing)
then node_modules/A/node_modules/C (present, should resolve here)
then root node_modules/C (only if still missing)
The buggy approach could skip step 2 due to path reconstruction from component IDs, causing a false miss.

Fix summary

Resolution now walks from [currentPath] upward through real .../node_modules/... ancestors, checking each candidate path before top-level fallback.
This aligns lookup behavior with npm module resolution semantics and prevents false “missing dependency” outcomes in ancestor-hoisted layouts.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/npm/NpmLockfile3Detector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/npm/NpmLockfile3Detector.cs
@@ -1,5 +1,6 @@
 namespace Microsoft.ComponentDetection.Detectors.Npm;
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.Json;
@@ -102,7 +103,7 @@ public class NpmLockfile3Detector : NpmLockfileDetectorBase
             this.RecordComponent(singleFileComponentRecorder, component, lockPackage.Dev ?? false, component);
 
             // Enqueue nested dependencies
-            this.EnqueueNestedDependencies(subQueue, path, lockPackage, packageLookup, singleFileComponentRecorder, component);
+            this.EnqueueNestedDependencies(subQueue, path, lockPackage, packageLookup, component);
 
             // Process sub-dependencies
             while (subQueue.Count > 0)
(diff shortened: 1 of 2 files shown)


src/Microsoft.ComponentDetection.Detectors/npm/NpmLockfile3Detector.cs
grvillic, line 42: Remember to bump detector version

RushabhBhansali (approved):

RushabhBhansali:
Have we investigated why are [verification tests](https://github.com/microsoft/component-detection/actions/runs/22467606798/job/65077173607?pr=1677) failing for Dotnet Components. Dotnet detector is not being touched in this PR. 

```
Expected foundComponent to be True because The component for DotNet-10.0.102 netstandard1.4 unknown - DotNet was not present in the new manifest file. Verify this is expected behavior before proceeding, but found False.
Expected foundComponent to be True because The component for DotNet-10.0.102 net40 unknown - DotNet was not present in the new manifest file. Verify this is expected behavior before proceeding, but found False.
Expected foundComponent to be True because The component for DotNet-10.0.102 net45 unknown - DotNet was not present in the new manifest file. Verify this is expected behavior before proceeding, but found False.
Expected foundComponent to be True because The component for PipReport-certifi 2026.1.4 - pip was not present in the new manifest file. Verify this is expected behavior before proceeding, but found False.
Expected foundComponent to be True because The component for DotNet-10.0.103 netstandard1.4 unknown - DotNet was not present in the old manifest file. Verify this is expected behavior before proceeding, but found False.
Expected foundComponent to be True because The component for DotNet-10.0.103 net40 unknown - DotNet was not present in the old manifest file. Verify this is expected behavior before proceeding, but found False.
Expected foundComponent to be True because The component for DotNet-10.0.103 net45 unknown - DotNet was not present in the old manifest file. Verify this is expected behavior before proceeding, but found False.
Expected foundComponent to be True because The component for NpmLockfile3-yallist 4.0.0 - Npm was not present in the old manifest file. Verify this is expected behavior before proceeding, but found False.
```

PR #1683, opened 2026-03-04
Move new maven detector out from experiment

zhenghao104:
We are replacing `MvnCliComponentDetector` logic and its unit test with `MavenWithFallBackDetector`

A separate PR will remove the legacy Maven static parsing detector in another repository.

### DIFF
--- a/test/Microsoft.ComponentDetection.VerificationTests/ComponentDetectionIntegrationTests.cs
+++ b/test/Microsoft.ComponentDetection.VerificationTests/ComponentDetectionIntegrationTests.cs
@@ -151,7 +151,7 @@ public class ComponentDetectionIntegrationTests
                         rightDependencies.Should().Contain(leftDependency, $"Component dependency {leftDependency} for component {leftComponent} was not in the {rightGraphName} dependency graph.");
                     }
 
-                    leftDependencies.Should().BeEquivalentTo(rightDependencies, $"{rightGraphName} has the following components that were not found in {leftGraphName}, please verify this is expected behavior. {System.Text.Json.JsonSerializer.Serialize(rightDependencies.Except(leftDependencies))}");
+                    leftDependencies.Should().BeEquivalentTo(rightDependencies, $"{rightGraphName} has the following components that were not found in {leftGraphName}, please verify this is expected behavior. {JsonSerializer.Serialize(rightDependencies.Except(leftDependencies))}");
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
FernandoRojo, line 107: Technically don't need this semaphore anymore, but i don't think there's any harm in keeping it there

test/Microsoft.ComponentDetection.VerificationTests/ComponentDetectionIntegrationTests.cs
```diff
@@ -183,7 +186,9 @@ public void CheckDetectorsRunTimesAndCounts()
             var oldMatches = Regex.Matches(this.oldLogFileContents, regexPattern);
             var newMatches = Regex.Matches(this.newLogFileContents, regexPattern);
 
-            newMatches.Should().HaveCountGreaterThanOrEqualTo(oldMatches.Count, "A detector was lost, make sure this was intentional.");
+            // Account for expected removed detectors
+            var expectedMatchCount = oldMatches.Count - expectedRemovedDetectors.Count;
```
FernandoRojo, line 190: There is an issue with adding this, as it would allow any detector to be removed after this and no longer fail this first check phase. 

 Since this is a test that runs as part of the PPE phase and not during the PR, i think it would be best to keep it as is and have that manual confirm step be the expectation. 

We also just need to fix those PPE tests in general as they are flaky and pretty much always fail: https://dev.azure.com/cg-scus1/MyFirstProject/_build?definitionId=697&_a=summary

test/Microsoft.ComponentDetection.VerificationTests/ComponentDetectionIntegrationTests.cs
```diff
@@ -240,7 +255,11 @@ private void ProcessDetectorVersions()
 
             if (newDetector == null)
             {
-                newDetector.Should().NotBeNull($"the detector {cd.DetectorId} was lost, verify this is expected behavior");
+                if (!expectedRemovedDetectors.Contains(cd.DetectorId))
```
FernandoRojo, line 258: This one is fine though

grvillic:
@zhenghao104 - Let's make sure to also update Maven's documentation in https://github.com/microsoft/component-detection/blob/main/docs/detectors/maven.md

PR #1708, opened 2026-03-10
Add OCI image support to Linux scanner

### LINKED ISSUE #100: Support scanning Linux container archives (tarball)
syft already supports this:

```
docker-archive:path/to/yourimage.tar   use a tarball from disk for archives created from "docker save"
oci-archive:path/to/yourimage.tar      use a tarball from disk for OCI archives (from Skopeo or otherwise)
```

It would allow using container build engines other than Docker.

jasonpaulos:
Adds support for scanning image as OCI directories and archive files.

Based on #1709

Fixes #100

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/linux/LinuxContainerDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/linux/LinuxContainerDetector.cs
@@ -275,12 +275,12 @@ public class LinuxContainerDetector(
         }
         catch (Exception e)
         {
-            this.logger.LogWarning(e, "Processing of image {ContainerImage} (kind {ImageType}) failed", imageRef.Reference, imageRef.Kind);
-            RecordImageDetectionFailure(e, imageRef.Reference);
+            this.logger.LogWarning(e, "Processing of image {ContainerImage} (kind {ImageType}) failed", imageRef.OriginalInput, imageRef.Kind);
+            RecordImageDetectionFailure(e, imageRef.OriginalInput);
 
             var singleFileComponentRecorder =
-                componentRecorder.CreateSingleFileComponentRecorder(imageRef.Reference);
-            singleFileComponentRecorder.RegisterPackageParseFailure(imageRef.Reference);
+                componentRecorder.CreateSingleFileComponentRecorder(imageRef.OriginalInput);
+            singleFileComponentRecorder.RegisterPackageParseFailure(imageRef.OriginalInput);
         }
     }
 
--- a/src/Microsoft.ComponentDetection.Detectors/linux/ImageReference.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/linux/ImageReference.cs
@@ -56,20 +56,32 @@ internal class ImageReference
     {
         if (input.StartsWith(OciDirPrefix, StringComparison.OrdinalIgnoreCase))
         {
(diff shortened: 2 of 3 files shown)


src/Microsoft.ComponentDetection.Detectors/linux/LinuxScanner.cs
```diff
@@ -87,10 +87,145 @@ public async Task<IEnumerable<LayerMappedLinuxComponents>> ScanLinuxAsync(
         LinuxScannerScope scope,
         CancellationToken cancellationToken = default
     )
+    {
```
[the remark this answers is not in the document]
jasonpaulos: The parameter was already present and not used before this PR. Keeping it there to minimize unnecessary changes for now

src/Microsoft.ComponentDetection.Detectors/linux/LinuxContainerDetector.cs
```diff
@@ -204,118 +209,347 @@ private static void RecordImageDetectionFailure(Exception exception, string imag
...
+                    DiffId = layer.Digest ?? string.Empty,
+                    LayerIndex = index,
+                })
+                .ToList() ?? [];
+
+            // Extract base image annotations from the Syft source metadata labels
+            var baseImageRef = string.Empty;
+            var baseImageDigest = string.Empty;
+            sourceMetadata?.Labels?.TryGetValue(BaseImageRefAnnotation, out baseImageRef);
+            sourceMetadata?.Labels?.TryGetValue(BaseImageDigestAnnotation, out baseImageDigest);
```
JamieMagee, line 475: This is a bit of a strange pattern. Do you need to initialize `baseImageRef` and `baseImageDigest` to `string.Empty`?

docs/detectors/linux.md
```diff
@@ -11,6 +11,27 @@ Linux detection depends on the following:
...
+#### Digest Only
+
+Images already present in the local Docker daemon can be referenced by just a digest. For example, `sha256:56bab49eef2ef07505f6a1b0d5bd3a601dfc3c76ad4460f24c91d6fa298369ab`.
+
+#### OCI Images
+
+Images present on the filesystem as either an [OCI layout directory](https://specs.opencontainers.org/image-spec/image-layout/) or an OCI image archive (tarball) can be referenced by file path.
+
+- For OCI image layout directories, use the prefix `oci-dir:` followed by the path to the directory, e.g. `oci-dir:/path/to/image`
+- For OCI image archives (tarballs), use the prefix `oci-archive:` followed by the path to the archive file, e.g. `oci-archive:/path/to/image.tar`
```
JamieMagee, line 33: Can OCI image archives be gzipped as well?

JamieMagee (approved):

docs/detectors/linux.md
jasonpaulos: I believe so according to syft, but I have not verified it personally: https://oss.anchore.com/docs/guides/sbom/scan-targets/#archive-scan-targets

src/Microsoft.ComponentDetection.Detectors/linux/LinuxContainerDetector.cs
jasonpaulos: We do need to initialize the strings to something here because the null conditional operators will prevent the TryGetValue from executing if the metadata or labels are null.

PR #1719, opened 2026-03-13
Fix maven detector race conditions

zhenghao104:
# Fix Maven Detector Race Conditions and Improve Reliability

## Overview

This PR addresses critical race conditions and reliability issues in Maven component detection by fixing coordination between `MavenWithFallbackDetector` and `MvnCliComponentDetector`, ensuring consistent and accurate component detection results.

### 🐛 **Issue 1: Race Condition in File Coordination**

**Problem**: Race condition between `MavenWithFallbackDetector` and `MvnCliComponentDetector` where one detector would delete dependency files before the other could process them, causing dramatically inconsistent component counts (e.g., 304 components vs 32 components).

**Root Cause**: Both detectors competed for the same temporary Maven dependency files without proper coordination, leading to file access conflicts.

**Solution**: 
- Implemented register-at-generation pattern in `MavenCommandService.GenerateDependenciesFileAsync()`
- Files are registered with `RegisterFileReader()` immediately upon generation
- Both detectors can safely process the same file before cleanup occurs
- Enhanced logging with detector IDs for better debugging and monitoring
- Proper file lifecycle coordination ensures consistent component detection across both detectors

### 🔧 **Issue 2: Incorrect Static Parser Processing Order**

**Problem**: `MavenWithFallbackDetector` static parsing fallback wasn't working correctly due to using `ConcurrentBag` (unordered collection), causing different processing order compared to `MavenComponentDetector` and inconsistent results.

**Solution**: 
- Changed from `ConcurrentBag<ProcessRequest>` to `ConcurrentQueue<ProcessRequest>`
- Ensures deterministic processing order matching `MavenComponentDetector`
- Provides consistent component detection results during fallback scenarios
- Maintains reliable fallback behavior when Maven CLI fails

### 📊 **Issue 3: Misleading Detection Method Telemetry**

**Problem**: Detection method telemetry was incorrectly labeled as "StaticParserOnly" even when Maven CLI was available and attempted, making monitoring and debugging unclear.

**Root Cause**: Logic incorrectly classified any CLI failures as static-only mode, regardless of whether CLI was actually available and attempted.

**Solution**: 
- Updated classification logic to properly distinguish detection methods:
  - **"MvnCliOnly"**: When all Maven CLI attempts succeed
  - **"Mixed"**: When Maven CLI is available but some attempts fail (includes authentication failures and partial successes)
  - **"StaticParserOnly"**: Only when Maven CLI is explicitly disabled by user or not available on system
- Updated corresponding unit test expectations to reflect correct behavior

## Impact

- ✅ **Eliminates race conditions** causing inconsistent component detection between detectors
- ✅ **Ensures both detectors produce identical results** on the same repositories  
- ✅ **Provides accurate telemetry** for monitoring detection method usage and debugging
- ✅ **Maintains optimal performance** while ensuring correctness and reliability
- ✅ **Improves debugging capabilities** with enhanced logging and detector identification

## Validation

- **Unit Tests**: All 771 Maven unit tests passing
- **Integration Tests**: All 6 verification tests passing against real-world repositories
- **Manual Testing**: Validated on multiple repository types with consistent results between detectors

## Files Changed

- `src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs`
- `src/Microsoft.ComponentDetection.Detectors/maven/IMavenCommandService.cs`  
- `src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs`
- `src/Microsoft.ComponentDetection.Detectors/maven/MvnCliComponentDetector.cs`
- `test/Microsoft.ComponentDetection.Detectors.Tests/MavenWithFallbackDetectorTests.cs`

## Breaking Changes

None. All changes are internal implementation improvements that maintain existing public API contracts.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
@@ -192,18 +192,22 @@ internal class MavenCommandService : IMavenCommandService
     /// <param name="detectorId">The identifier of the detector unregistering the file reader.</param>
     public void UnregisterFileReader(string dependencyFilePath, string detectorId = null)
     {
-        this.fileReaderCounts.AddOrUpdate(dependencyFilePath, 0, (key, count) => Math.Max(0, count - 1));
+        var newCount = this.fileReaderCounts.AddOrUpdate(dependencyFilePath, 0, (key, count) => Math.Max(0, count - 1));
         this.logger.LogDebug(
             "{DetectorId}: Unregistered file reader for {DependencyFilePath}, count: {Count}",
             detectorId ?? "Unknown",
             dependencyFilePath,
-            this.fileReaderCounts.TryGetValue(dependencyFilePath, out var currentCount) ? currentCount : 0);
-        this.TryDeleteDependencyFileIfNotInUse(dependencyFilePath, detectorId);
+            newCount);
+
+        // If no readers remain, attempt cleanup
+        if (newCount == 0)
+        {
+            this.TryDeleteDependencyFileIfNotInUse(dependencyFilePath, detectorId);
+        }
     }
 
     /// <summary>
     /// Attempts to delete a dependency file if no detectors are currently using it.
-    /// Uses atomic counter operations for thread-safe coordination.
     /// </summary>
     /// <param name="dependencyFilePath">The path to the dependency file to delete.</param>
     /// <param name="detectorId">The identifier of the detector requesting the deletion.</param>
@@ -211,25 +215,7 @@ internal class MavenCommandService : IMavenCommandService
     {
         var detector = detectorId ?? "Unknown";
 
-        // Atomically check if any readers are using the file
-        var shouldDelete = false;
-        this.fileReaderCounts.AddOrUpdate(
-            dependencyFilePath,
-            0,
-            (key, count) =>
-            {
-                shouldDelete = count == 0;
-                return count;
-            });
-
-        if (!shouldDelete)
-        {
-            var currentCount = this.fileReaderCounts.TryGetValue(dependencyFilePath, out var count) ? count : 0;
-            this.logger.LogDebug("{DetectorId}: Skipping deletion of {DependencyFilePath} - {Count} readers still active", detector, dependencyFilePath, currentCount);
-            return;
-        }
-
-        // Safe to delete - no readers are using the file
+        // Safe to delete - no readers are using the file (count was already verified to be 0)
         try
         {
             if (File.Exists(dependencyFilePath))


src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
```diff
@@ -171,33 +175,48 @@ public void ParseDependenciesFile(ProcessRequest processRequest)
     /// This prevents premature deletion by other detectors.
     /// </summary>
     /// <param name="dependencyFilePath">The path to the dependency file being read.</param>
-    public void RegisterFileReader(string dependencyFilePath)
+    /// <param name="detectorId">The identifier of the detector registering the file reader.</param>
+    public void RegisterFileReader(string dependencyFilePath, string detectorId = null)
     {
         this.fileReaderCounts.AddOrUpdate(dependencyFilePath, 1, (key, count) => count + 1);
+        this.logger.LogDebug(
+            "{DetectorId}: Registered file reader for {DependencyFilePath}, count: {Count}",
```
FernandoRojo: DetectorId here is just going to be "GenerationStep" every time, i'd just keep it as "Registered file reader for {DependencyFilePath}, count: {Count}" and not bother with the id here since that would add uneccessary mess

src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
```diff
@@ -215,44 +234,44 @@ private void TryDeleteDependencyFileIfNotInUse(string dependencyFilePath)
...
             }
 
-            this.logger.LogDebug("Created coordination file {CoordinationFile} for process {ProcessId}", coordinationFile, processId);
+            this.logger.LogDebug("{DetectorId}: Created coordination file {CoordinationFile} for process {ProcessId}", detector, coordinationFile, processId);
 
             // Give other processes a chance to create their coordination files if needed
             Thread.Sleep(50);
 
             // Verify we still own the coordination (another process might have deleted and recreated it)
             if (!File.Exists(coordinationFile))
```
FernandoRojo: I don't think we need any of this coordination file logic, it can just be 
```
Atomic(
    count--
    check if count == 0
)

if !check:
    do nothing (log that deletion skipped
if check:
    delete file
```

FernandoRojo: so just the first two statements in a simple lock

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -576,6 +575,7 @@ private void ProcessPomFileStatically(ProcessRequest processRequest)
                     {
                         var component = new MavenComponent(groupId, artifactId, versionString);
                         var detectedComponent = new DetectedComponent(component);
+                        this.Logger.LogWarning("FallbackMaven: Registering component: {ComponentName}@{ComponentVersion} from file: {FileName}", component.Id, component.Version, Path.GetFileName(filePath));
```
FernandoRojo: This is a lot of logging, i'd set the level to at debug if we really need it.

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -531,14 +524,20 @@ private void ProcessPomFileStatically(ProcessRequest processRequest)
...
             var document = new XmlDocument();
             document.Load(file.Stream);
 
             lock (this.documentsLoaded)
             {
-                this.documentsLoaded.TryAdd(file.Location, document);
+                var wasAdded = this.documentsLoaded.TryAdd(file.Location, document);
+                if (wasAdded)
+                {
+                    this.Logger.LogWarning("FallbackMaven: Added document to shared dictionary: {FilePath} (Total docs: {Count})", file.Location, this.documentsLoaded.Count);
```
FernandoRojo: debug

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -531,14 +524,20 @@ private void ProcessPomFileStatically(ProcessRequest processRequest)
         var singleFileComponentRecorder = processRequest.SingleFileComponentRecorder;
         var filePath = file.Location;
 
+        this.Logger.LogWarning("FallbackMaven: Processing pom.xml: {FilePath}", filePath);
```
FernandoRojo: debug

src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
```diff
@@ -174,105 +178,73 @@ public void ParseDependenciesFile(ProcessRequest processRequest)
...
+        // Atomically check if any readers are using the file
+        var shouldDelete = false;
+        this.fileReaderCounts.AddOrUpdate(
+            dependencyFilePath,
+            0,
+            (key, count) =>
+            {
+                shouldDelete = count == 0;
+                return count;
+            });
```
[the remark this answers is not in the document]
zhenghao104: Image

Maybe this approach is better?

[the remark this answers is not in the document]
zhenghao104: Added

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
zhenghao104: Removed the extra logging

src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
zhenghao104: Added, check the other comment

src/Microsoft.ComponentDetection.Detectors/maven/MavenCommandService.cs
zhenghao104: Removed, but kept the ID in unregister

PR #1724, opened 2026-03-16
More maven parsing fix for experiment detector

zhenghao104:
# PR: Maven Detection Improvements - Variable Resolution & Centralized Cleanup

## Summary

This PR addresses two critical issues in the Maven detection system:
1. **Variable resolution race conditions** causing incorrect version resolution in multi-module Maven projects
2. **Premature file cleanup** causing race conditions between `MvnCliComponentDetector` and `MavenWithFallbackDetector`

### Issue 1: Variable Resolution Order Dependency

> **Note:** This issue exists in the current detector's implementation—it is not a new bug introduced by this PR. This PR fixes the existing behavior.

**Before:** The variable resolution logic processed POM files in arbitrary order and used a flat `ConcurrentDictionary<string, string>` for variable storage. When the same variable (e.g., `${revision}`) was defined in multiple POM files (parent and child), the resolution result depended on processing order—whichever file was processed last would overwrite the variable value.

**Example scenario:**
```
parent/pom.xml:     <revision>1.0.0</revision>
child/pom.xml:      <revision>2.0.0</revision>  (inherits from parent)
```

If parent was processed after child, the child's dependency would incorrectly resolve to `1.0.0` instead of `2.0.0` (Maven's rule: child properties take precedence over parent).

### Issue 2: Premature bcde.mvndeps File Cleanup

> **Note:** This race condition only occurs during the experimental phase when both `MvnCliComponentDetector` and `MavenWithFallbackDetector` run concurrently. Once `MavenWithFallbackDetector` is promoted from experimental to production and replaces `MvnCliComponentDetector`, only one Maven detector will run and the centralized cleanup will no longer be necessary (though it remains harmless).

**Before:** Each detector (`MvnCliComponentDetector` and `MavenWithFallbackDetector`) independently tracked and cleaned up `bcde.mvndeps` files it created. Since both detectors run concurrently and share the same temporary files, one detector could delete files before the other finished reading them.

### 1. Hierarchy-Aware Variable Resolution

**New approach:** Variables are now stored with file path context:
```csharp
// Old: flat storage (order-dependent)
collectedVariables["revision"] = "1.0.0";

// New: file-scoped storage
collectedVariables["C:\project\parent\pom.xml::revision"] = "1.0.0";
collectedVariables["C:\project\child\pom.xml::revision"] = "2.0.0";
```

**Resolution algorithm:**
1. Build Maven parent-child relationships based on `<parent>` declarations
2. When resolving `${variable}` for a file, walk up the Maven hierarchy (child → parent → grandparent)
3. Return the first match found (implements Maven's "closest ancestor wins" rule)

**Key methods:**
- `FindVariableInMavenHierarchy()` - Walks Maven parent chain to find variable definition
- `ResolveVersionWithHierarchyAwareness()` - Resolves version templates using hierarchy search
- `ResolveDeferredParentRelationships()` - Second pass to resolve parent relationships when POMs are processed out-of-order

### 2. Three-Pass Processing Model

The static parser now uses three passes:

| Pass | Purpose |
|------|---------|
| **First** | Collect variables from all POM files, establish parent relationships where possible |
| **Second** | Resolve deferred parent relationships (for out-of-order processing) |
| **Third** | Resolve pending components using hierarchy-aware variable lookup |

### 3. Centralized Cleanup in DetectorProcessingService

**Before:** Cleanup was distributed across detectors
```csharp
// In MvnCliComponentDetector
private readonly HashSet<string> _createdFiles = new();
// Each detector tracks and deletes its own files
```

**After:** Cleanup moved to `DetectorProcessingService.CleanupMavenFiles()`
```csharp
// After ALL detectors finish:
if (settings.CleanupCreatedFiles ?? true)
{
    this.CleanupMavenFiles(settings.SourceDirectory, detectors);
}
```

**Benefits:**
- No race condition between detectors
- Single responsibility for cleanup
- Respects `CleanupCreatedFiles` setting
- Uses symlink-safe directory traversal (same pattern as `FastDirectoryWalkerFactory`)

### 4. Shared Constants

Created `MavenConstants.cs` to keep detector IDs and filenames in sync:
```csharp
public static class MavenConstants
{
    public const string BcdeMvnDependencyFileName = "bcde.mvndeps";
    public const string MvnCliDetectorId = "MvnCli";
    public const string MavenWithFallbackDetectorId = "MavenWithFallback";
}
```

### 5. Telemetry Improvements

- Added `UnresolvedVariableCount` telemetry (aggregated count instead of per-variable warnings)
- Fixed `PendingComponentCount` to capture count **before** draining the queue
- Changed verbose logs to debug level to reduce noise for customers

### 6. Symlink-Safe File Enumeration

The cleanup logic now uses the same symlink handling pattern as `FastDirectoryWalkerFactory`:
- Detects reparse points (symlinks/junctions)
- Resolves physical paths via `IPathUtilityService`
- Tracks visited directories by real path to prevent infinite loops
- Wraps symlink resolution in try/catch to handle broken symlinks gracefully

## Behavioral Changes

| Aspect | Before | After |
|--------|--------|-------|
| Variable resolution | Order-dependent, flat storage | Hierarchy-aware, file-scoped |
| Multi-module projects | Could resolve wrong version | Correctly follows Maven inheritance |
| File cleanup | Per-detector, race-prone | Centralized, after all detectors complete |
| Circular symlinks | Could cause infinite loop in cleanup | Safely handled with visited tracking |
| Telemetry | Per-variable warnings, incorrect pending count | Aggregated counts, accurate metrics |

## Testing

- All existing unit tests pass (800+ MavenWithFallbackDetector tests, 140+ DetectorProcessingService tests)
- Verified on multiple repositories locally:
- Component counts match between old and new implementations
- No regression in detection accuracy

## Files Changed

| File | Changes |
|------|---------|
| `MavenWithFallbackDetector.cs` | Hierarchy-aware variable resolution, three-pass processing, telemetry fixes |
| `MavenConstants.cs` | New - shared constants for detector IDs and filename |
| `MavenCliCleanupTelemetryRecord.cs` | New - telemetry for cleanup operations |
| `DetectorProcessingService.cs` | Centralized cleanup with symlink-safe enumeration |
| `MavenCommandService.cs` | Use shared constants |
| `MvnCliComponentDetector.cs` | Use shared constants, removed file tracking |
| Test files | Updated to use shared constants, removed timing-based assertions |

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
@@ -141,6 +141,8 @@ public class MavenWithFallbackDetector : FileComponentDetector, IExperimentalDet
     private MavenFallbackReason fallbackReason = MavenFallbackReason.None;
     private int mvnCliComponentCount;
     private int staticParserComponentCount;
+    private int unresolvedVariableCount;
+    private int pendingComponentCountBeforeResolution;
     private bool mavenCliAvailable;
 
     public MavenWithFallbackDetector(
@@ -207,9 +209,6 @@ public class MavenWithFallbackDetector : FileComponentDetector, IExperimentalDet
     private void LogDebugWithId(string message) =>
         this.Logger.LogDebug("{DetectorId}: {Message}", this.Id, message);
 
-    private void LogInfo(string message) =>
-        this.Logger.LogInformation("{DetectorId}: {Message}", this.Id, message);
-
     private void LogWarning(string message) =>
         this.Logger.LogWarning("{DetectorId}: {Message}", this.Id, message);
 
@@ -256,6 +255,8 @@ public class MavenWithFallbackDetector : FileComponentDetector, IExperimentalDet
         this.fallbackReason = MavenFallbackReason.None;
         this.mvnCliComponentCount = 0;
         this.staticParserComponentCount = 0;
+        this.unresolvedVariableCount = 0;
+        this.pendingComponentCountBeforeResolution = 0;
         this.mavenCliAvailable = false;
     }
 
@@ -334,7 +335,7 @@ public class MavenWithFallbackDetector : FileComponentDetector, IExperimentalDet
     {
         if (this.envVarService.IsEnvironmentVariableValueTrue(DisableMvnCliEnvVar))
         {
-            this.LogInfo($"MvnCli detection disabled via {DisableMvnCliEnvVar} environment variable. Using static pom.xml parsing only.");
+            this.LogDebugWithId($"MvnCli detection disabled via {DisableMvnCliEnvVar} environment variable. Using static pom.xml parsing only.");
             this.usedDetectionMethod = MavenDetectionMethod.StaticParserOnly;
             this.fallbackReason = MavenFallbackReason.MvnCliDisabledByUser;
             this.mavenCliAvailable = false;
@@ -354,7 +355,7 @@ public class MavenWithFallbackDetector : FileComponentDetector, IExperimentalDet
 
         if (!this.mavenCliAvailable)
         {
-            this.LogInfo("Maven CLI not found in PATH. Will use static pom.xml parsing only.");
+            this.LogDebugWithId("Maven CLI not found in PATH. Will use static pom.xml parsing only.");
             this.usedDetectionMethod = MavenDetectionMethod.StaticParserOnly;
             this.fallbackReason = MavenFallbackReason.MavenCliNotAvailable;
             return false;
@@ -571,14 +572,15 @@ public class MavenWithFallbackDetector : FileComponentDetector, IExperimentalDet
         this.Telemetry["MavenCliAvailable"] = this.mavenCliAvailable.ToString();
         this.Telemetry["OriginalPomFileCount"] = this.originalPomFiles.Count.ToString();
         this.Telemetry["CollectedVariableCount"] = this.collectedVariables.Count.ToString();
-        this.Telemetry["PendingComponentCount"] = this.pendingComponents.Count.ToString();
+        this.Telemetry["PendingComponentCount"] = this.pendingComponentCountBeforeResolution.ToString();
+        this.Telemetry["UnresolvedVariableCount"] = this.unresolvedVariableCount.ToString();
 
         if (!this.failedEndpoints.IsEmpty)
         {
             this.Telemetry["FailedEndpoints"] = string.Join(";", this.failedEndpoints.Distinct().Take(10));
         }
 
-        this.LogInfo($"Detection completed. Method: {detectionMethodStr}, " +
+        this.LogDebugWithId($"Detection completed. Method: {detectionMethodStr}, " +
                      $"FallbackReason: {fallbackReasonStr}, " +
                      $"MvnCli components: {mvnCliCountStr}, " +
                      $"Static parser components: {staticCountStr}");
@@ -1070,7 +1072,7 @@ public class MavenWithFallbackDetector : FileComponentDetector, IExperimentalDet
 
         if (resolvedCount > 0 || unresolvedCount > 0)
         {
-            this.LogInfo($"Resolved {resolvedCount} deferred parent relationships, {unresolvedCount} remain unresolved");
+            this.LogDebugWithId($"Second pass (parent resolution) completed: {resolvedCount} deferred parent relationships resolved, {unresolvedCount} remain unresolved");
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.ComponentDetection.Orchestrator/Services/DetectorProcessingService.cs
```diff
@@ -422,4 +425,55 @@ private void LogTabularOutput(ConcurrentDictionary<string, DetectorRunResult> pr
...
+            return;
+        }
+
+        try
+        {
+            var searchPattern = "*.mvndeps"; // Simple file pattern for Directory.GetFiles
+
+            this.logger.LogDebug("Starting Maven CLI cleanup in directory: {SourceDirectory}", sourceDirectory.FullName);
+
+            await Task.Run(() =>
```
grvillic: Unless I am missing something, we don't have anything async here. Why is it being wrapped with Task.Run? We will still be blocking a thread while waiting for the output, so it isn't scaling the actual work IIRC.

src/Microsoft.ComponentDetection.Orchestrator/Services/DetectorProcessingService.cs
```diff
@@ -422,4 +425,55 @@ private void LogTabularOutput(ConcurrentDictionary<string, DetectorRunResult> pr
...
+            {
+                // Search recursively for Maven CLI dependency files
+                var filesToClean = Directory.GetFiles(sourceDirectory.FullName, searchPattern, SearchOption.AllDirectories);
```
grvillic: Why not leveraging `FileSystemEnumerable` and `Parallel.ForEach` which can handle streaming and fewer allocations, see [FastDirectoryWalkerFactory.cs](https://github.com/microsoft/component-detection/blob/7b302d5eca371ea792e2028c858ac58c10de0fae/src/Microsoft.ComponentDetection.Common/FastDirectoryWalkerFactory.cs#L19). The longer the list of files to explore the slower this will be. It would be good to have a telemetry record to capture how long did we take doing the cleanup.

src/Microsoft.ComponentDetection.Orchestrator/Services/DetectorProcessingService.cs
```diff
@@ -422,4 +425,55 @@ private void LogTabularOutput(ConcurrentDictionary<string, DetectorRunResult> pr
...
+        var hasMavenDetectors = detectors.Any(d => mavenDetectorIds.Contains(d.Id));
+
+        if (!hasMavenDetectors)
+        {
+            return;
+        }
+
+        try
+        {
+            var searchPattern = "*.mvndeps"; // Simple file pattern for Directory.GetFiles
```
grvillic: Let's make this variable a private file-level constant, also we don't expect other file names, only `bcde.mvndeps` right? Why not only sticking to that pattern and create a public constant variable that can be shared between the detectors that use it, and this place so they are always in sync.

src/Microsoft.ComponentDetection.Orchestrator/Services/DetectorProcessingService.cs
```diff
@@ -422,4 +425,55 @@ private void LogTabularOutput(ConcurrentDictionary<string, DetectorRunResult> pr
...
     }
+
+    /// <summary>
+    /// Cleans up Maven CLI temporary files after all detectors have finished.
+    /// This prevents race conditions between MvnCliComponentDetector and MavenWithFallbackDetector.
+    /// </summary>
+    private async Task CleanupMavenFilesAsync(DirectoryInfo sourceDirectory, IEnumerable<IComponentDetector> detectors)
+    {
+        // Only clean up if Maven detectors are running
+        var mavenDetectorIds = new[] { "MvnCli", "MavenWithFallback" };
```
grvillic: Without reflection I think we can make another constant string shared between the detector IDs and here, so we are never out of sync.

src/Microsoft.ComponentDetection.Orchestrator/Services/DetectorProcessingService.cs
FernandoRojo: We also don't want to end up with the same infinite loop bug from before

src/Microsoft.ComponentDetection.Detectors/maven/MavenWithFallbackDetector.cs
```diff
@@ -213,11 +213,61 @@ private void LogInfo(string message) =>
...
+            // Intentionally empty - just draining the queue
+        }
+
+        while (this.mavenCliErrors.TryDequeue(out _))
+        {
+            // Intentionally empty - just draining the queue
+        }
+
+        while (this.failedEndpoints.TryDequeue(out _))
+        {
```
FernandoRojo, line 250: This does sem redundant just because we know that we should only be evoking each detector once, but i have no hard requirements for removing it.

PR #1752, opened 2026-03-30
Fix VcpkgComponent purl construction for names containing slashes

JamieMagee:
`VcpkgComponent.PackageUrl` builds purl strings by interpolating `this.Name` directly into a string like `$"pkg:vcpkg/{this.Name}@{this.Version}"`. This breaks when vcpkg SPDX entries have names with slashes (e.g. `google/brotli`, `capstone-engine/capstone`), because the purl parser treats those slashes as path separators. If the name produces empty segments between slashes, parsing throws `MalformedPackageUrlException: The purl namespace has an empty segment between '/' separators.`

This switches to the `PackageUrl` component constructor, which accepts the name as a discrete parameter and handles encoding during serialization.

### DIFF
--- a/test/Microsoft.ComponentDetection.Detectors.Tests/VcpkgComponentDetectorTests.cs
+++ b/test/Microsoft.ComponentDetection.Detectors.Tests/VcpkgComponentDetectorTests.cs
@@ -151,6 +151,59 @@ public class VcpkgComponentDetectorTests
     }
 
     [TestMethod]
+    public async Task TestResourceWithSlashInNameProducesValidPackageUrlAsync()
+    {
+        var spdxFile = """
+            {
+                "SPDXID": "SPDXRef-DOCUMENT",
+                "documentNamespace": "https://spdx.org/spdxdocs/brotli-x64-windows",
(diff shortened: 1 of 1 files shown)


src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs
```diff
@@ -53,18 +54,11 @@ public override PackageUrl PackageUrl
...
-            }
-            else if (this.Version != null)
-            {
-                return new PackageUrl($"pkg:vcpkg/{this.Name}@{this.Version}");
-            }
-            else
-            {
-                return new PackageUrl($"pkg:vcpkg/{this.Name}");
-            }
+            var qualifiers = this.PortVersion > 0
```
zhenghao104, line 57: Can we add unit tests to cover the broken case we saw in prod today?

JamieMagee: Absolutely. Added a regression test.

Issue #201, opened 2022-08-16
Investigate our use of `**` with `Dotnet.Glob`


JamieMagee:
I believe we are incorrectly using the `**` pattern. We rolled back in #200, but would like to upgrade still.

Sources:
- https://github.com/microsoft/component-detection/releases/tag/v1.2.1
- https://github.com/microsoft/component-detection/pull/138
- https://github.com/dazinator/DotNet.Glob/issues/65
- https://github.com/dazinator/DotNet.Glob/commit/f1ee0efcc82c375e8124f08053190509eeb3de45

cobya:
Ideally most of these issues will be resolved after the changes from #1091 are merged.

JamieMagee:
Looks like the fix mentioned above never made it into v5. I'll need to cherry-pick it.

PR #1759, opened 2026-04-02
Add component detectors for Docker Compose, Helm, and Kubernetes

jpinz:
This pull request adds new detectors for Docker Compose, Helm, and Kubernetes YAML files, enabling the system to automatically discover and register Docker image references found in these types of configuration files. It also updates the service registration and the `DetectorClass` enum to support these new detectors.

It also adds tests for the new detectors plus the Dockerfile one.

Image references containing unresolved variable placeholders (e.g., `${REGISTRY}/app:${TAG}`, `{{ .Values.image.tag }}`) are silently skipped in all four Docker-related detectors (DockerCompose, Helm, Kubernetes, and Dockerfile) to avoid noisy warnings for otherwise-valid real-world manifests. A shared `DockerReferenceUtility.HasUnresolvedVariables` helper was added to `Microsoft.ComponentDetection.Common` to consolidate this logic, and all detectors—including the existing `DockerfileComponentDetector`—now use it.

The Helm detector uses the combination of a `Chart.yaml`/`Chart.yml` file and a `*values*.(yaml|yml)` file in the same directory to identify Helm charts. Values files found without a co-located Chart file are skipped, avoiding false positives from unrelated `values.yaml` files in non-Helm projects.

A bug in `DetectorRestrictionService` was also fixed: when both `AllowedDetectorIds` and `AllowedDetectorCategories` are specified, DefaultOff detectors introduced via category expansion are now also constrained by the ID allow-list, producing an intersection rather than an unintended union.

#### New Detector Checklist

- [X] Component type defined (if new ecosystem) in `ComponentType.cs` and `DetectorClass.cs`
- [ ] ~~Component class created in `TypedComponent/` folder with proper validation~~
- [X] Detector class created inheriting from `FileComponentDetector` and implementing `IDefaultOffComponentDetector`
- [X] Detector registered in `ServiceCollectionExtensions.cs`
- [X] Constructor properly injects required services
- [X] `SearchPatterns` defined for file discovery
- [X] `OnFileFoundAsync()` implemented with component registration logic
- [x] Unit tests created in `Detectors.Tests/` using `DetectorTestUtilityBuilder`
- [X] Verification test resources added to `VerificationTests/resources/`
- [ ] Detector tested locally with `--DetectorArgs YourDetectorId=EnableIfDefaultOff`


jpinz:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/component-detection/pull/1759#pullrequestreview-4056410104)

src/Microsoft.ComponentDetection.Detectors/helm/HelmComponentDetector.cs
```diff
@@ -0,0 +1,215 @@
...
+        this.Scanner = walkerFactory;
+        this.Logger = logger;
+    }
+
+    public override string Id => "Helm";
+
+    public override IList<string> SearchPatterns { get; } =
+    [
+        "Chart.yaml", "Chart.yml",
+        "chart.yaml", "chart.yml",
```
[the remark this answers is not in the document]
jpinz: We need to check for Chart.yaml and values.yaml files because if someone just has a random file called values.yaml it doesn't necessarily imply helm. But the combination of the values and chart yaml files usually indicates that the values.yaml file that will be parsed is probably helm related.

jpinz:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/component-detection/pull/1759#pullrequestreview-4056892329) Note the response I left about the values/chart yaml files as well

jpinz:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/component-detection/pull/1759#pullrequestreview-4057040599)

src/Microsoft.ComponentDetection.Detectors/helm/HelmComponentDetector.cs
```diff
@@ -0,0 +1,243 @@
...
+        "*values*.yaml", "*values*.yml",
```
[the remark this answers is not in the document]
jpinz: I have tests for this exact scenario, it seems to work just fine

src/Microsoft.ComponentDetection.Detectors/dockercompose/DockerComposeComponentDetector.cs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    public override string Id => "DockerCompose";
+
+    public override IList<string> SearchPatterns { get; } =
+    [
+        "docker-compose.yml", "docker-compose.yaml",
+        "docker-compose.*.yml", "docker-compose.*.yaml",
+        "compose.yml", "compose.yaml",
+        "compose.*.yml", "compose.*.yaml",
```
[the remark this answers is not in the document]
jpinz: I have tests for this exact scenario, it seems to work just fine

docs/detectors/helm.md
```diff
@@ -0,0 +1,50 @@
+# Helm Detection
+
+## Requirements
+
+Helm detection depends on the following to successfully run:
+
+- One or more Helm values files matching the patterns: `*values*.yaml`, `*values*.yml`
+- Chart metadata files (`Chart.yaml`, `Chart.yml`, `chart.yaml`, `chart.yml`) are matched for file discovery but only values files are parsed for image references
```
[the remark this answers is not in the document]
jpinz: I have tests for this exact scenario, it seems to work just fine

docs/detectors/dockercompose.md
```diff
@@ -0,0 +1,48 @@
+# Docker Compose Detection
+
+## Requirements
+
+Docker Compose detection depends on the following to successfully run:
+
+- One or more Docker Compose files matching the patterns: `docker-compose.yml`, `docker-compose.yaml`, `docker-compose.*.yml`, `docker-compose.*.yaml`, `compose.yml`, `compose.yaml`, `compose.*.yml`, `compose.*.yaml`
```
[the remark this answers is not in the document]
jpinz: I have tests for this exact scenario, it seems to work just fine

src/Microsoft.ComponentDetection.Detectors/helm/HelmComponentDetector.cs
```diff
@@ -0,0 +1,243 @@
...
+        // Only process values files — and only when co-located with a Chart.yaml in the same directory.
+        if (!IsValuesFile(fileName))
+        {
+            return;
+        }
+
+        if (!this.helmChartDirectories.ContainsKey(directory))
+        {
+            return;
+        }
```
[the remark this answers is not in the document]
jpinz: @copilot apply changes based on [this feedback](https://github.com/microsoft/component-detection/pull/1759#discussion_r3034247662)

src/Microsoft.ComponentDetection.Detectors/dockercompose/DockerComposeComponentDetector.cs
```diff
@@ -0,0 +1,145 @@
+#nullable disable
```
JamieMagee: Enable nullability for new files please

src/Microsoft.ComponentDetection.Detectors/helm/HelmComponentDetector.cs
```diff
@@ -0,0 +1,254 @@
...
+        this.Scanner = walkerFactory;
+        this.Logger = logger;
+    }
+
+    public override string Id => "Helm";
+
+    public override IList<string> SearchPatterns { get; } =
+    [
+        "Chart.yaml", "Chart.yml",
+        "chart.yaml", "chart.yml",
```
JamieMagee: Are the lowercase `chart.yaml` / `chart.yml` entries here for Linux case-sensitivity? `IsChartFile()` already does `OrdinalIgnoreCase`, so once a file is matched by the walker, it'll be handled correctly. The question is whether the walker's glob is case-sensitive.

src/Microsoft.ComponentDetection.Detectors/kubernetes/KubernetesComponentDetector.cs
```diff
@@ -0,0 +1,248 @@
...
+    {
+        this.ComponentStreamEnumerableFactory = componentStreamEnumerableFactory;
+        this.Scanner = walkerFactory;
+        this.Logger = logger;
+    }
+
+    public override string Id => "Kubernetes";
+
+    public override IList<string> SearchPatterns { get; } = ["*.yaml", "*.yml"];
+
```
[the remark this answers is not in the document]
JamieMagee: +1 on this. I know you need to scan every YAML file, but that means `OnFileFoundAsync` is going to be a _very_ hot path. Make sure it's as optimized as possible.

src/Microsoft.ComponentDetection.Detectors/kubernetes/KubernetesComponentDetector.cs
```diff
@@ -0,0 +1,248 @@
...
+    {
+        "Pod",
+        "Deployment",
+        "StatefulSet",
+        "DaemonSet",
+        "ReplicaSet",
+        "Job",
+        "CronJob",
+        "ReplicationController",
+    };
```
JamieMagee, line 30: Missing `PodTemplate` from the `KubernetesKinds` set. It's not common but it does embed a pod spec with container images.

test/Microsoft.ComponentDetection.Detectors.Tests/HelmComponentDetectorTests.cs
```diff
@@ -0,0 +1,414 @@
...
+        var (scanResult, componentRecorder) = await this.DetectorTestUtility
+            .WithFile("Chart.yaml", chartYaml)
+            .ExecuteDetectorAsync();
+
+        scanResult.ResultCode.Should().Be(ProcessingResultCode.Success);
+        componentRecorder.GetDetectedComponents().Should().BeEmpty();
+    }
+
+    [TestMethod]
+    public async Task TestHelm_ValuesFileObservedBeforeChartYamlAsync()
```
JamieMagee, line 167: This passes because both files default to tmp as their directory. They're co-located by coincidence, not by test design.

I'd add a test where `Chart.yaml` is in `/tmp/chartA/` and `values.yaml` is in `/tmp/chartB/` (via the `fileLocation` parameter) to actually verify the co-location filtering rejects mismatched directories. `TestHelm_ValuesWithoutChartYamlSkippedAsync` tests "no chart at all" but not "chart exists in a different directory."

src/Microsoft.ComponentDetection.Detectors/dockercompose/DockerComposeComponentDetector.cs
```diff
@@ -0,0 +1,145 @@
...
+                    if (!string.IsNullOrWhiteSpace(imageRef))
+                    {
+                        this.TryRegisterImageReference(imageRef, recorder, fileLocation);
+                    }
+                }
+            }
+        }
+    }
+
+    private void TryRegisterImageReference(string imageReference, ISingleFileComponentRecorder recorder, string fileLocation)
```
JamieMagee: `TryRegisterImageReference` is copy-pasted across all three detectors. Consider pulling it into a shared static method on `DockerReferenceUtility`. Something like `TryParseImageReference(string) -> DockerReference?` that handles the unresolved-variable check, parsing, and exception swallowing. The Dockerfile detector does the same work inline too, so it'd consolidate four copies.

JamieMagee:
I think you're going to need to solve https://github.com/microsoft/component-detection/issues/201 before you can merge this.

See https://github.com/microsoft/component-detection/pull/1091. It was originally set to merge in v5.0.0, but accidentally got dropped before release.

src/Microsoft.ComponentDetection.Detectors/helm/HelmComponentDetector.cs
jpinz: Will have to look into the walker glob logic, but removed it for now

src/Microsoft.ComponentDetection.Detectors/dockercompose/DockerComposeComponentDetector.cs
jpinz: Done!

src/Microsoft.ComponentDetection.Detectors/kubernetes/KubernetesComponentDetector.cs
jpinz: Added!

src/Microsoft.ComponentDetection.Detectors/helm/HelmComponentDetector.cs
jpinz: Looks like it is case-insensitive so we should be good

test/Microsoft.ComponentDetection.Detectors.Tests/HelmComponentDetectorTests.cs
jpinz: Technically I believe you could have a `values.yaml` file in a subdirectory of where the `chart.yaml` is, but for the time being we won't support that.

src/Microsoft.ComponentDetection.Detectors/kubernetes/KubernetesComponentDetector.cs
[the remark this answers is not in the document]
jpinz: @JamieMagee I just updated it so that it should be even more optimized now.

src/Microsoft.ComponentDetection.Detectors/dockercompose/DockerComposeComponentDetector.cs
```diff
@@ -0,0 +1,145 @@
...
+
+    public override string Id => "DockerCompose";
+
+    public override IList<string> SearchPatterns { get; } =
+    [
+        "docker-compose.yml", "docker-compose.yaml",
+        "docker-compose.*.yml", "docker-compose.*.yaml",
+        "compose.yml", "compose.yaml",
+        "compose.*.yml", "compose.*.yaml",
+    ];
```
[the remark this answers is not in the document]
jpinz: This has been fixed

test/Microsoft.ComponentDetection.Detectors.Tests/HelmComponentDetectorTests.cs
```diff
@@ -0,0 +1,432 @@
...
+    public async Task TestHelm_LowercaseChartYamlAsync()
+    {
+        var chartYaml = @"
+apiVersion: v2
+name: my-chart
+version: 0.1.0
+";
+
+        var (scanResult, componentRecorder) = await this.DetectorTestUtility
+            .WithFile("chart.yaml", chartYaml)
```
JamieMagee, line 410: Are search patterns case-sensitive by default?

JamieMagee (approved):

JamieMagee (approved):

test/Microsoft.ComponentDetection.Detectors.Tests/HelmComponentDetectorTests.cs
jpinz: Not by default no, but helm's `Chart.yml` file does need to be uppercased. I added that detail in the code comments: https://helm.sh/docs/chart_best_practices/conventions/#usage-of-the-words-helm-and-chart

src/Microsoft.ComponentDetection.Detectors/kubernetes/KubernetesComponentDetector.cs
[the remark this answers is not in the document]
grvillic: @jpinz - This search path is too broad, pick a subset of yaml names, most of them have nothing to do with kubernetes, we should discriminate early before getting to OnFileFound. Pick a subset of yaml patterns that match the common usage.

grvillic (changes requested):
There is no need to add all detectors in a single PR, break them up into separate PRs.

PR #1785, opened 2026-04-17
Add the Docker Compose Detector

jpinz:
This pull request introduces a new detector for Docker Compose files, improves Docker image reference parsing, and enhances documentation. The main changes are the addition of the `DockerComposeComponentDetector` (default-off), refactoring of Dockerfile detection logic to use new utility methods, and updates to documentation to describe the new detector and its behavior.

This is PR 1 of 3 for splitting up #1759


src/Microsoft.ComponentDetection.Common/DockerReference/DockerReferenceUtility.cs
```diff
@@ -38,6 +38,53 @@ public static class DockerReferenceUtility
...
+            return null;
+        }
+
+        try
+        {
+            return ParseFamiliarName(imageReference);
+        }
+        catch
+        {
+            return null;
```
JamieMagee, line 72: Log this?

src/Microsoft.ComponentDetection.Detectors/dockercompose/DockerComposeComponentDetector.cs
```diff
@@ -0,0 +1,124 @@
+#nullable enable
```
JamieMagee: I don't think you need this? nullable is opt-out, not opt-in.

src/Microsoft.ComponentDetection.Detectors/dockerfile/DockerfileComponentDetector.cs
JamieMagee, line 1: The changes here look good, but why include them with the DockerComposeDetector changes?

jpinz: Cause I was refactoring the helper utility methods into the utilities class, and it made sense to migrate the Dockerfile detector to reference those as well.

src/Microsoft.ComponentDetection.Common/DockerReference/DockerReferenceUtility.cs
jpinz: The previous implementation of this logic in just the `DockerfileComponentDetector` would just throw the error instead of swallowing it and returning `null`. Instead of passing an `ILogger` around everywhere, I'm just removing the catch and updating the docs to reflect that it will throw if it can't be parsed, or returns `null` if variables are present.

JamieMagee (approved):

grvillic (approved):

src/Microsoft.ComponentDetection.Common/DockerReference/DockerReferenceUtility.cs
jpinz: I did add logs after all

JamieMagee (approved):

JamieMagee (approved):

PR #1792, opened 2026-04-20
Increase parallelism in smoketest yaml + reduce iterations in PR

grvillic:
## Context
We have seen these smoke test runs take forever to be provisioned. I increased parallelism + reduce number of dotnet scans in PR builds to make them finish faster and made them skip execution for draft PRs.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -10,6 +10,11 @@ on:
   pull_request:
     paths:
       - "src/**"
+      - "Directory.Build.props"
+      - "Directory.Build.targets"
+      - "Directory.Packages.props"
+      - "global.json"
+      - ".github/workflows/smoke-test.yml"
   schedule:
     - cron: "0 0 * * *" # every day at midnight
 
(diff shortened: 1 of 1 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -82,7 +89,8 @@ jobs:
       - name: Run Smoke Test
         working-directory: src/Microsoft.ComponentDetection
         run: |
-          for i in $(seq 1 10); do
+          ITERATIONS=${{ github.event_name == 'schedule' && 10 || 1 }}
+          for i in $(seq 1 $ITERATIONS); do
               dotnet run -c Release -- scan --SourceDirectory ${{ github.workspace }}/smoke-test-repo --Verbosity Verbose || exit 1
           done
```
pauld-msft, line 91: why do we have multiple iterations for the schedule runs?

pauld-msft (approved):

.github/workflows/smoke-test.yml
grvillic: IIRC a while ago there was once an outage in the detector that was caused by multiple runs (OOM), but I think it is an overkill to do that on every single PR. If we have memory issues, we can do a single run on default branch.

pauld-msft (approved):

PR #1822, opened 2026-06-06
feat: defensive identity overrides for MavenComponent and GitComponent

zhenghao104:
## Summary

Two small defensive identity overrides on `TypedComponent` subclasses, plus a `PackageUrl` override on `GitComponent` so github.com repositories get a canonical `pkg:github/...` PURL.

| Component | Change |
|---|---|
| `MavenComponent` | Override `GetExtendedIdProperties()` to return empty (suppress `DownloadUrl`/`SourceUrl` from `Id`). |
| `GitComponent` | Same `GetExtendedIdProperties()` override + new `PackageUrl` override returning `pkg:github/{owner}/{repo}@{commit}` for github.com URLs and `null` for other hosts. |

## Why

Today no detector populates `DownloadUrl`/`SourceUrl` on `MavenComponent` or `GitComponent`, so `Id == BaseId`. Upcoming SBOM-enrichment work in `microsoft/component-detection-internal` will start emitting deterministic download URLs through the Maven/Git converters. Without these overrides, `Id` would change shape to `"BaseId [DownloadUrl:... SourceUrl:...]"`, creating bare-vs-rich identity-merge conflicts between:

- the two `MvnCli` internal paths (mvn-CLI output vs. static pom.xml parser), and
- the multiple Git submitters (CocoaPods / Poetry / Pip / Swift / Ruby / cgmanifest).

`GitComponent.PackageUrl` was the only typed component returning `null`. Adding the github PURL aligns it with every other type and lets downstream tooling (SBOM, AzDO Governance) dedupe github coords by PURL instead of by raw repository URL.

## Behavior changes

- **Maven** — `Id` already equaled `BaseId` for every existing detector; the override locks the contract going forward. No observable change today.
- **Git** — `Id` likewise unchanged today. `PackageUrl` goes from `null` → `"pkg:github/owner/repo@commit"` for **github.com** repositories only. Other hosts (GitLab / Bitbucket / Azure DevOps / GitHub Enterprise) keep returning `null` — consumers should continue to fall back to `RepositoryUrl` for those.
  - Searched both this repo and `microsoft/component-detection-internal` for callers that assert `GitComponent.PackageUrl == null` — none found. `VerificationTests` uses `PackageUrl` in a dedup key but both sides of the intersection run the same binary, so keys still match.

## Edge cases covered

| URL shape | Result |
|---|---|
| `https://github.com/google/guava` | `pkg:github/google/guava@<commit>` |
| `https://github.com/google/guava.git` | `.git` suffix stripped |
| `https://github.com/google/guava/` | trailing slash normalised |
| `https://GitHub.com/google/guava` | host match is case-insensitive |
| `https://github.com:443/google/guava` | port stripped via `Uri.Host` |
| `https://github.com/google/guava?utm=x#readme` | query/fragment ignored via `Uri.AbsolutePath` |
| `https://gitlab.com/foo/bar`, `https://bitbucket.org/foo/bar`, `https://dev.azure.com/...`, `https://github.contoso.com/foo/bar` | `null` |
| `https://github.com/onlyowner`, `https://github.com/owner/repo/tree/main`, `https://github.com/` | `null` (not canonical owner/repo paths) |
| SSH-style `git@github.com:owner/repo.git` | not a parseable absolute `Uri`; rejected upstream |
| `RepositoryUrl == null` or `CommitHash` null/empty (deserialization path) | `null` |

## Tests

`Microsoft.ComponentDetection.Contracts.Tests` — **+10 tests, 84/84 passing**:

- **`PurlGenerationTests`** — explicit `MavenComponentShouldGenerateMavenPurl` plus 7 GitHub PURL cases: canonical, `.git` suffix, trailing slash, case-insensitive host, non-github hosts (gitlab + bitbucket + ADO + GHE in one test), malformed paths (owner-only + too-deep + root-only in one test), missing commit hash.
- **`TypedComponentSerializationTests`** — `MavenComponent_Id_ExcludesDownloadUrlAndSourceUrl_WhenSet` and `GitComponent_Id_ExcludesDownloadUrlAndSourceUrl_WhenSet` prove the overrides exclude both URLs even when set.

Broader regression check (all run locally on this branch):

| Project | Result |
|---|---|
| `Contracts.Tests` | **84 / 84 passed** (incl. 10 new) |
| `Detectors.Tests` | **910 / 910 passed** — `GradleComponentDetectorTests` and others that inspect `Id` strings like `"... - Maven"` still match because only `GetExtendedIdProperties()` changed, not `ComputeBaseId()`. |
| `Orchestrator.Tests` | **151 / 151 passed** |
| `Common.Tests` | 213 / 221 passed (1 pre-existing flake: `DockerService_CanPingDocker` needs a Docker daemon; the other 7 skips are platform-conditioned) |

## Out of scope

- The CD-Internal converter changes (consume the new `GitComponent.PackageUrl`, emit deterministic `DownloadURL` for both Maven and Git) ship in a follow-up PR in `microsoft/component-detection-internal` once a new CD-public NuGet drops.
- A pre-existing bug at `GitComponentConverter.cs:28` in CD-Internal (literal `Replace("https://github.com/", "")` that produces garbage `owner`/`repo` for non-github URLs) — same follow-up PR.

### DIFF
--- a/src/Microsoft.ComponentDetection.Contracts/TypedComponent/GitComponent.cs
+++ b/src/Microsoft.ComponentDetection.Contracts/TypedComponent/GitComponent.cs
@@ -2,10 +2,15 @@
 namespace Microsoft.ComponentDetection.Contracts.TypedComponent;
 
 using System;
+using System.Collections.Generic;
 using System.Text.Json.Serialization;
+using PackageUrl;
 
 public class GitComponent : TypedComponent
 {
+    private const string GithubHost = "github.com";
+    private const string DotGitSuffix = ".git";
+
     public GitComponent(Uri repositoryUrl, string commitHash)
     {
         this.RepositoryUrl = this.ValidateRequiredInput(repositoryUrl, nameof(this.RepositoryUrl), nameof(ComponentType.Git));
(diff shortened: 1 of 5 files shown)


RushabhBhansali:
does verification tests currently test for PackageUrl ? We should probably update the pom file [here](https://github.com/microsoft/component-detection/blob/main/test/Microsoft.ComponentDetection.VerificationTests/resources/maven/lib/pom.xml) to ensure that test data has some packageUrls.

RushabhBhansali (approved):

PR #1825, opened 2026-06-10
Add option to exclude base image components

jasonpaulos:
This PR makes two changes:

### Improve the way components from container images are mapped to layers

Prior to this, the Syft output's "Locations" field was the sole source of a component's file and layer information. This was undesirable because, for packages installed by the system package manager, the only location would be the system package manager database. As a result, any changes to system packages in later layers would result in every system package being mapped to the last layer where the database was modified.

Now, the Syft output's "Metadata.Files" field is used to augment the "Locations" data about files and layers associated with a component, and any known system package manager database files are ignored in favor of other file mappings, if present. This leads to more accurate mappings of components to image layers.

The expected result of this change is that some components which were previously mapped to layers above the base image may now be properly mapped to base image layers.

### Add a new option to exclude components which solely originate from the base image when scanning a container image.

The flag `--ExcludeBaseImageComponents` can be specified to filter out components from the base image.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/linux/LinuxScanner.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/linux/LinuxScanner.cs
@@ -30,6 +30,21 @@ internal class LinuxScanner : ILinuxScanner
 
     private static readonly IList<string> ScopeSquashedParameter = ["--scope", "squashed"];
 
+    /// <summary>
+    /// Well-known package manager database paths whose layer attribution should be ignored
+    /// when determining which layer a system package belongs to. These files are shared across
+    /// all packages managed by the same package manager and get updated whenever any package is
+    /// installed or removed, causing unrelated packages to appear as modified in later layers.
+    /// </summary>
+    private static readonly HashSet<string> PackageManagerDatabasePaths = new(StringComparer.OrdinalIgnoreCase)
+    {
+        "/var/lib/dpkg/status",
+        "/lib/apk/db/installed",
+        "/var/lib/rpm/Packages",
+        "/var/lib/rpm/Packages.db",
+        "/var/lib/rpm/rpmdb.sqlite",
+    };
+
     private static readonly SemaphoreSlim ContainerSemaphore = new SemaphoreSlim(2);
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


JamieMagee (approved):

jpinz (approved):

jpinz (approved):

src/Microsoft.ComponentDetection.Orchestrator/Services/GraphTranslation/DefaultGraphTranslationService.cs
```diff
@@ -41,15 +41,175 @@ public ScanResult GenerateScanResultFromProcessingResult(
...
+            return components;
+        }
+
+        // Build an indexed lookup: containerDetailId → (layerIndex → DockerLayer)
+        var layerLookup = new Dictionary<int, Dictionary<int, DockerLayer>>();
+        foreach (var (id, details) in containerDetailsMap)
+        {
+            if (details.Layers != null)
+            {
+                layerLookup[id] = details.Layers.ToDictionary(l => l.LayerIndex);
```
grvillic: Are we confident these layers are always unique? I have seen many times `ToDictionary` throwing errors when there happen to be two entries with the same value when we don't have much control on the source that emits this data, can we just group by and do a first/last to safely convert to dictionary?

grvillic (approved):

src/Microsoft.ComponentDetection.Orchestrator/Services/GraphTranslation/DefaultGraphTranslationService.cs
jasonpaulos: Great point. I believe layers that have the same exact content will have the same ID here. I'll add proper behavior for this case

jasonpaulos: Actually this line is keyed on layer index, will should always be unique. Still, I added grouping logic just in case.

I also added a unit test for duplicate layer IDs to add coverage to other layer handling code.

grvillic (approved):

src/Microsoft.ComponentDetection.Detectors/linux/LinuxScanner.cs
```diff
@@ -179,11 +194,16 @@ private IEnumerable<LayerMappedLinuxComponents> ProcessSyftOutputWithTelemetry(
             }
         }
 
+        // Build a file path → layerID map from the top-level files listing.
+        // This allows us to determine layer attribution for files owned by a package
+        // even when the artifact's locations only reference the package manager database.
+        var filePathToLayerId = BuildFilePathToLayerMap(syftOutput.Files);
+
```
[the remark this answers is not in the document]
jasonpaulos: The suggestion does not provide a meaningful performance improvement, since it only optimizes for the case where the Syft output has no artifacts.

PR #1837, opened 2026-07-01
Upgrade Contracts to net8

ryanbrandenburg:
Went to use some newer language features in the Contracts projects and they weren't available because it was still on netstandard. Upgraded to net8 as per https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/ and then fixed the resulting build errors.


ryanbrandenburg:
The team thinks some consumers of this project as a package still need netstandard, so I'm going to close this and we'll re-evaluate when the consumers are updated.

PR #1839, opened 2026-07-07
Promote uv.lock detector to stable

ryanbrandenburg:
The telemetry for this detector is looking good so let's promote it from experimental to stable. This will turn it on-by-default and increase the # of components that are detected in any repository that uses uv by default.


ryanbrandenburg:
Closing this for now while we get our deployment story sorted.

ShaiNahum:
I would love to see it getting deployed! My coworker just flagged my usage of it on a PR with:
> Do we really want to opt in to an experimental feature? Those always seem to come back to bite you.

PR #1851, opened 2026-07-23
Use DTO for uv.lock parsing

ryanbrandenburg:
This simplifies the parsing of uv.lock files by just using the built-in model parser. It comes at the expense of some resiliency to malformed files, but given that these are generated files and the reliability of data from a manually edited file is questionable that seems either preferable or like an acceptable tradeoff to me.

### DIFF
--- a/src/Microsoft.ComponentDetection.Detectors/uv/UvPackage.cs
+++ b/src/Microsoft.ComponentDetection.Detectors/uv/UvPackage.cs
@@ -3,31 +3,30 @@ namespace Microsoft.ComponentDetection.Detectors.Uv;
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Runtime.Serialization;
+using System.Text.Json.Serialization;
 using Microsoft.ComponentDetection.Contracts.TypedComponent;
 
-[DataContract]
 internal class UvPackage
 {
-    [DataMember(Name = "name")]
+    [JsonPropertyName("name")]
     public string Name { get; set; } = string.Empty;
 
-    [DataMember(Name = "version")]
+    [JsonPropertyName("version")]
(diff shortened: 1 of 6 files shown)


src/Microsoft.ComponentDetection.Detectors/uv/UvPackage.cs
```diff
@@ -2,25 +2,37 @@ namespace Microsoft.ComponentDetection.Detectors.Uv;
...
-    public required string Version { get; init; }
+    [DataMember(Name = "version")]
+    public string Version { get; set; } = string.Empty;
 
+    [DataMember(Name = "dependencies")]
     public List<UvDependency> Dependencies { get; set; } = [];
 
-    // Metadata dependencies (requires-dist)
-    public List<UvDependency> MetadataRequiresDist { get; set; } = [];
+    [DataMember(Name = "metadata")]
```
grvillic: Why `DataMember`? I have seen some "JsonPropertyName" used in the repository, is there any particular reason we still had to use it in this package? Probably nothing will break but for my own understanding of the current state of things with our serialization libraries since I know some has migrated but not all of it yet.

ryanbrandenburg: I was following the pattern used by [other TOML detectors](https://github.com/microsoft/component-detection/blob/main/src/Microsoft.ComponentDetection.Detectors/poetry/Contracts/PoetryPackage.cs). 

Although interestingly Tomlyn has explicit support for `JsonPropertyName` and I can't find any evidence in their repo that they pay attention to [DataMember] so you're likely right, or more generally the default name transformations are sufficient without explicit serialization attributes needed.

PR #1852, opened 2026-07-23
Extract DownloadUrl from uv.lock file

ryanbrandenburg:
This gets us extracting a DownloadUrl from packages in uv.lock files using either sdist (source distribution) or wheel (a pre-built distribution format).

Because the `wheels` field is an array that may contain many links (per-build-environement compilations) I thought it was preferable to offer the singular and platform-agnostic sdist as the primary download location.

The [DTO PR](https://github.com/microsoft/component-detection/pull/1851) and this will conflict with each other.


ryanbrandenburg:
CoPilot makes a good point, if the DownloadUrl is used to make identity decisions than selecting it from a wheels list of even a sdist may be inappropriate since neither of those are proport to be "canonical" links.

PR #1858, opened 2026-08-19
Handle circular dependencies

ryanbrandenburg:
As-is it's possible to get stuck in a look when A->B->A (Or A->B->C->A, etc) which eventually causes a stack-overflow. This PR adds a set to keep track of visited packages so we can be sure we never end up in a loop. Before the fix the unit test I added would crash the test-host. I also checked the fix against the real-world report which alerted me to this issue.


ryanbrandenburg:
Closing this in favor of https://github.com/microsoft/component-detection/pull/1857#pullrequestreview-4976774733

PR #1866, opened 2026-09-15
Make Swift Detector experimental and improve output

ryanbrandenburg:
This detector exists but isn't getting a lot of use since it's off by default. Let's make it an experiment so it gets a bit of exercise.


AMaini503 (approved):

src/Microsoft.ComponentDetection.Contracts/TypedComponent/SwiftComponent.cs
```diff
@@ -11,24 +11,38 @@ namespace Microsoft.ComponentDetection.Contracts.TypedComponent;
...
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="SwiftComponent"/> class.
+    /// </summary>
+    /// <param name="name">The name of the component.</param>
+    /// <param name="version">The version of the component.</param>
+    /// <param name="packageUrl">The package URL of the component.</param>
+    /// <param name="kind">The Swift package kind.</param>
+    /// <param name="hash">The Git commit hash of the component.</param>
+    public SwiftComponent(string name, string version, string packageUrl, string kind, string hash)
```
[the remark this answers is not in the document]
ryanbrandenburg: Resolved by adding a parameterless constructor.

src/Microsoft.ComponentDetection.Detectors/swiftpm/SwiftResolvedComponentDetector.cs
```diff
@@ -81,18 +81,10 @@ private void ProcessPackageResolvedFile(ISingleFileComponentRecorder singleFileC
                         name: package.Identity,
                         version: version,
                         packageUrl: package.Location,
+                        kind: package.Kind,
                         hash: package.State.Revision);
                     var newDetectedSwiftComponent = new DetectedComponent(component: detectedSwiftComponent);
                     singleFileComponentRecorder.RegisterUsage(newDetectedSwiftComponent);
-
```
grvillic, line 87: why is this being removed? If a swift package manager is using a git url one can get additional vulnerability coverage, see https://osv.dev/list?ecosystem=GIT. Not all swift components come from git tho, so adding the relevant condition is necessary to avoid noise.

grvillic (changes requested):

src/Microsoft.ComponentDetection.Contracts/TypedComponent/SwiftComponent.cs
```diff
@@ -37,6 +42,12 @@ public SwiftComponent(string name, string version, string packageUrl, string has
     [JsonPropertyName("version")]
     public string Version { get; }
 
+    [JsonPropertyName("kind")]
```
grvillic, line 43: do we truly need to commit hash and Kind? Mind adding context on why these was added, possibly Kind makes sense, but commit hash might explode the data a bit, or we can end up with duplicate packages if one has commit and the other one does not.

src/Microsoft.ComponentDetection.Contracts/TypedComponent/SwiftComponent.cs
grvillic: this component ID, is what makes the component unique across a scan. If we are adding new properties we need to review if they indeed make a component unique and include here, or update existing entry to include the additional fields, etc. We should be consistent tho, and have a reason why to capture that metadata.

src/Microsoft.ComponentDetection.Detectors/swiftpm/SwiftResolvedComponentDetector.cs
ryanbrandenburg: First some context

A standard "pins" entry in a Package.resolved file looks like the following:
```json
{
  "identity" : "aexml",
  "kind" : "remoteSourceControl",
  "location" : "https://github.com/tadija/AEXML.git",
  "state" : {
    "revision" : "db806756c989760b35108146381535aec231092b",
    "version" : "4.7.0"
  }
},
```
There exist other `kind`s such as "registry", "local", and `sourceControl` but they are explicitly not supported by the current detector and seem less common.

And here's some example entries from before my change, both from the same ScanManifest output file:
The SwiftComponent
```json
{
  "locationsFoundAt": [
    "/Cuckoo/Generator/Generator.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved"
  ],
  "component": {
    "type": "Swift",
    "name": "aexml",
    "version": "4.7.0",
    "packageUrl": {
      "Scheme": "pkg",
      "Type": "swift",
      "Namespace": "github.com/tadija",
      "Name": "aexml",
      "Version": "4.7.0",
      "Qualifiers": {
        "repository_url": "https://github.com/tadija/AEXML.git"
      },
      "Subpath": null
    },
    "id": "aexml 4.7.0 - Swift"
  },
  "detectorId": "Swift",
  "isDevelopmentDependency": null,
  "dependencyScope": null,
  "topLevelReferrers": [],
  "ancestralReferrers": [],
  "containerDetailIds": [],
  "containerLayerIds": {},
  "targetFrameworks": []
},
```
The GitComponent
```json
{
  "locationsFoundAt": [
    "/Cuckoo/Generator/Generator.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved"
  ],
  "component": {
    "type": "Git",
    "repositoryUrl": "https://github.com/tadija/AEXML.git",
    "commitHash": "db806756c989760b35108146381535aec231092b",
    "tag": "4.7.0",
    "packageUrl": {
      "Scheme": "pkg",
      "Type": "github",
      "Namespace": "tadija",
      "Name": "aexml",
      "Version": "db806756c989760b35108146381535aec231092b",
      "Qualifiers": null,
      "Subpath": null
    },
    "id": "https://github.com/tadija/AEXML.git : db806756c989760b35108146381535aec231092b - Git"
  },
  "detectorId": "Swift",
  "isDevelopmentDependency": null,
  "dependencyScope": null,
  "topLevelReferrers": [],
  "ancestralReferrers": [],
  "containerDetailIds": [],
  "containerLayerIds": {},
  "targetFrameworks": []
},
```

The commit hash was the only bit of information present in the GitComponent but not in the SwiftComponent, so I thought we could prevent some confusion/duplication by only reporting a component once. If GitComponent's get some kind of special treatment then I think there's a case for making all `"kind" : "remoteSourceControl",` entries (the only kind currently supported but that's a different question) GitComponents and not emitting the SwiftComponent at all (until/unless we support "kind": "registry" or others).

src/Microsoft.ComponentDetection.Contracts/TypedComponent/SwiftComponent.cs
ryanbrandenburg: I don't believe it is possible to have a Package.resolved remoteSourceControl entry without a "revision" (hash). [Here's some relevant code](https://github.com/swiftlang/swift-package-manager/blob/fc49787af79722e02abf63b12824108d58c56eb4/Sources/Workspace/Workspace%2BResolvedPackages.swift#L128-L145) from SwiftPM. Unfortunately they don't have a formal specification for Package.resolved files but it would make sense that for your Git-based package manager the revision would always be required (and controlling) because the tag/version can be changed at a later date (and it would even be quite common for scenerios like "latest"). I'd go so far as to say that a Package.resolved remoteSourceControl entry without `state.revision` is malformed.

I think there's a stronger case for removing "kind" since right now we explicitly only support "remoteSourceControl" values out of "kind", but I would still vote to include it to make future support for `"kind": "registry"` more straightforward, and so that we can add statistics for the same to see if adding that support is worthwhile.

I'm less worried about exploding the data here since as long as we only keep 1 of the SwiftComponent and GitComponent we're starting out having halved our output from this detector.

src/Microsoft.ComponentDetection.Contracts/TypedComponent/SwiftComponent.cs
ryanbrandenburg: Ah, good catch. If we land on reporting "remoteSourceControl" as a SwiftComponent I think we should update this to include the CommitHash, and potentially remove the version until we have support for kind=registry. If we report remoteSourceControl as GitComponent then I don't think this file will get used at all until we add registry support.

src/Microsoft.ComponentDetection.Contracts/TypedComponent/SwiftComponent.cs
```diff
@@ -11,31 +11,43 @@ namespace Microsoft.ComponentDetection.Contracts.TypedComponent;
...
+    public string Version { get; set; }
+
+    [JsonPropertyName("kind")]
+    public string Kind { get; set; }
+
+    [JsonPropertyName("commitHash")]
+    public string CommitHash { get; set; }
+
+    [JsonPropertyName("repositoryUrl")]
+    public Uri RepositoryUrl { get; set; }
```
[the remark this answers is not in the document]
ryanbrandenburg: The typed-component tests are attempting to we don't have any serialization issues like it not finding a constructor or not being able to populate a parameter because it's get only. Backward compatible POCO classes are not a goal.

src/Microsoft.ComponentDetection.Contracts/TypedComponent/SwiftComponent.cs
```diff
@@ -11,31 +11,43 @@ namespace Microsoft.ComponentDetection.Contracts.TypedComponent;
...
     /// </summary>
     /// <param name="name">The name of the component.</param>
     /// <param name="version">The version of the component.</param>
-    /// <param name="packageUrl">The package URL of the component.</param>
-    /// <param name="hash">The hash of the component.</param>
-    public SwiftComponent(string name, string version, string packageUrl, string hash)
+    /// <param name="repositoryUrl">The source repository URL of the component.</param>
+    /// <param name="hash">The Git commit hash of the component.</param>
+    /// <param name="kind">The Swift package kind.</param>
+    public SwiftComponent(string name, string version, string repositoryUrl, string hash, string kind)
```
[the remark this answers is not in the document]
ryanbrandenburg: Binary compatibility is not a goal.

src/Microsoft.ComponentDetection.Detectors/swiftpm/SwiftResolvedComponentDetector.cs
ryanbrandenburg: Actually on reflection this morning I'm not sure that GitComponent is the right move. If we take [the Go Detector](https://github.com/microsoft/component-detection/blob/469778f6091a34f9888069cf66497bdee888b079/src/Microsoft.ComponentDetection.Detectors/go/Parsers/GoSumParser.cs#L56) as an example of another PM that uses Git to drive things it always reports it as a GoComponent and then leaves it to the consumer to decide what to do with the fact that a GoComponent can ALSO be consumed as a Git repo. On the surface that makes more sense to me than having consumers have to check the detectorId in order to understand that this is a Swift component.

That doesn't seem to be universally applied though given [examples like CocoaPods](https://github.com/microsoft/component-detection/blob/469778f6091a34f9888069cf66497bdee888b079/src/Microsoft.ComponentDetection.Detectors/cocoapods/PodComponentDetector.cs#L81-L95) so it seems like there's either something deeper driving the distinction or we're not consistent today.

src/Microsoft.ComponentDetection.Contracts/TypedComponent/SwiftComponent.cs
ryanbrandenburg: To this end I changed the ExtendedBaseId to be $"{repositoryUrl} {commitHash} - {component.Type}". RepositoryUrl and commitHash are the unique identifiers for a given package's identity/provenance. Name and version are basically syntactic sugar over the top of those definitive values, which AFAIK means they don't belong on the ID at all. Happy to chat about it if that doesn't meet the design though.

grvillic (approved):

AMaini503 (approved):

